PR #145, opened 2015-02-20
Simplify to use HEAD instead of GET requests

colemarkham:


eugenp:
Hey Cole - it's an interesting idea to use HEAD - thanks. I merged the work, as I do think it's cleaner than using GET and not doing anything with the actual response body, but I did notice one thing that I find odd. The second test now consistently takes twice as long to run. I haven't looked into why that is yet. 
Cheers, 
Eugen.

colemarkham:
That is interesting. I didn't look at timing. I assumed the HEAD would be
faster. I don't suppose there's any caching with the GET is there? What
happens if you only run the second test and not the first, or change the
order?

PR #268, opened 2015-10-18
Update the codebase as per the documented standards

mnkartik:


mnkartik:
Added new test method in ValidationTest

PR #283, opened 2015-11-24
Setting Up Swagger 2 with a Spring REST API

alexVengrovsk:
It is a code for Setting Up Swagger 2 with a Spring REST API tutorial


alexVengrovsk:
Old brocken code

PR #299, opened 2015-12-09
Adding Cassandra Templates and CQL Queries tutorial code samples

sameira:


davidmorley:
Hi @sameira -- the code looks great. Really happy you used an embedded Cassandra instance for the tests. It makes it super-simple for the readers to play around!

One thing I found was that one of the tests failed periodically (`whenSavingBook_thenAvailableOnRetrieval_usingPreparedStatements`). Basically, I think that the `ingest` method is asynchronous, and sometimes the read that follows it is faster than the ingest, so there's nothing to retrieve, and when you run your assertion, the `retrievedBook` is null.

Thoughts?

sameira:
Hi David,

Thanks for the feedback. 
I have noticed a transient Cassandra startup timeout issue but not any assertion issues.
Can you please share the error logs if you have any ?

sameira:
Hi David,

Hi David.

I made three changes,

1. Started Cassandra embedded with timeout of 25ms
2. Added thread sleep of 5ms for stability to the test.
                        (Anyway, The default consistency level setting is ONE which makes sure the data is being written at once before read.)
3. Updated Spring Data Cassandra version 

Can you please have a look now ?

PR #310, opened 2015-12-16
spring-javamail

jamesjithin:
JavaMail implementation with Spring.


davidmorley:
@jamesjithin Let's remove the `.classpath` and `.project` files. Additionally, we may want to consider modifying the example somewhat. There are a couple concerns:

* First, we've hardwired a "fake" username and password into this code
* The test doesn't actually test anything; it just executes the code

My concern with the fake email addresses/password and is that we'll need to add instructions (e.g. in a `README.md`) on how/where to modify the code so that it is usable. Additionally, you might want to consider using mocking in the test to verify that the library would contact GMail to send an email.

I think maybe the best way to approach all of this would be to add some non-test code, such as a `GmailSenderService` that uses the created beans to send messages. Then you can write a test that tests the `GmailSenderService` and verifies that the `JavaMailSender` would actually be invoked.

You could also look at externalizing the properties for username and password so that it can be configured in a properties file.

Thoughts?

jamesjithin:
@davidmorley, have made the changes. Please review.

davidmorley:
@jamesjithin The test doesn't appear to work (even after following the instructions). Have you tried running it?

jamesjithin:
@davidmorley, I had run the test and it worked with a properly configured credentials.

davidmorley:
@eugenp Can you take a look at this and make sure it's not just something I'm doing wrong?

@jamesjithin I also moved the `MailBeans.xml` and `application.properties` into `/src/test/resources`, which is standard for Maven-based applications.

eugenp:
Quick comments: 
- let's use the standard Eclipse formatter; if you're working in InteliJ - unfortunately I haven't created a corresponding one, but, right at the end, a simple way to solve it is to open up the project in Eclipse and do a project-wide formatting pass
- tests can declare thrown exceptions, so the try-catch is unnecessary
- if this is a Spring enabled test - then let's use the Spring test framework instead of bootstrapping a context manually like that
- the tests sets the "TO" email but it also needs to set the "FROM"

- since this is a single test, let's do the simplest thing possible - which would be sending a single email (instead of 2); we can of course have other tests that send more, but the very first one should be simple

- finally, the test isn't (for me at least) able to find/load up the MailBeans.xml - have a look at a more mature way of loading that resource in (there are many other Spring enabled tests in many of the other modules of the project - you can use any of them as a reference / starting point)

Hope that helps.

PR #321, opened 2015-12-30
Moved from XML config to Java config

gmaipady:


davidmorley:
@gmaipady The update looks great! Thanks.

PR #328, opened 2016-01-10
Initial commit, added code for article, "Injecting Mockito Mocks into Spring Beans"

amedviediev:
Added code for article on Injecting Mockito Mocks into Spring Beans

### DIFF
--- a/mockito-mocks-spring-beans/.gitignore
+++ b/mockito-mocks-spring-beans/.gitignore
@@ -1,0 +1,13 @@
+*.class
+
+#folders#
+/target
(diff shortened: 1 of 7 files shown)


davidmorley:
Hey, @amedviediev. I have just one suggestion regarding the code:  I think the effect of the mock injection would be clearer if the user names returned by the real and mock services were different. For example, we could have the mock service return something like "Mock user name" and the real service return "Real user name".

I'll go ahead and merge the pull request and make the change. I just wanted you to know what I was changing.

PR #332, opened 2016-01-14
Added RAML files for first two articles

KevinGilmore:
Hi Eugen, I have committed files for the first two RAML articles into the raml directory. Files for the first article are in the introduction folder, and for the second in the resource-types-and-traits folder.
-Kevin


eugenp:
Looks good.

PR #344, opened 2016-01-30
Source code for the article "Java - zip and unzip a collection of files"

joetaras:
Added source code for the article "Java - zip and unzip a collection of files"


core-java-8/src/main/java/com/baeldung/unzip/UnzipFile.java
```diff
@@ -0,0 +1,39 @@
...
+				while ((len = zis.read(buffer)) > 0) {
+					fos.write(buffer, 0, len);
+				}
+				fos.close();
+				zipEntry = zis.getNextEntry();
+			}
+			zis.closeEntry();
+			zis.close();
+		} catch (final IOException ex) {
+			ex.printStackTrace();
```
eugenp: I already covered the try-catch and how we can remove it entirely, as well as the formatting. Once these are done, let me know and I'll merge this in. 
Cheers, 
Eugen.

PR #378, opened 2016-03-04
Add form tag example

ivanp81:


davidmorley:
Hey Ivan. We'll actually keep the XML config in the code -- it was just in the article that we wanted both. I'll add the changes you made to the multipart handling, though.

PR #452, opened 2016-06-18
Spring REST service with protocol buffers.

nguyennamthai:


spring-protobuf/src/main/java/com/baeldung/protobuf/BaeldungTraining.java
```diff
@@ -0,0 +1,2719 @@
+// Generated by the protocol buffer compiler.  DO NOT EDIT!
```
pivovarit: This is autogenerated an unnecessary. Remove it

PR #453, opened 2016-06-18
Add mock-comparisons module

varomorf:
For article Mockito vs EasyMock vs JMockit with id 12717.


mock-comparisons/pom.xml
```diff
@@ -0,0 +1,105 @@
...
+		</resources>
+
+		<plugins>
+
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<version>${maven-compiler-plugin.version}</version>
+				<configuration>
+					<source>1.7</source>
```
pivovarit: Why not 1.8?

varomorf:
@pivovarit I must admit to have commited the sin of copy-paste and not looking deep enough into the code :P

PR #457, opened 2016-06-22
Code and related files for the dependency injection demos.

reymalahay:
@eugenp i recently made more changes to the code, as per editorial discussion with @atheedom


dependency-injection/build.gradle
```diff
@@ -0,0 +1,43 @@
+apply plugin: 'java'
```
pivovarit: @eugenp is this ok to mix Gradle with Maven?

eugenp: Not generally, unless we have a good reason to do so, I'd stick with Maven.

reymalahay: @eugenp, should i delete the build.gradle file?

eugenp: I haven't looked at the code in detail, but yes - basically, most projects here use Maven, so let's make sure this does the same.

reymalahay: @eugenp, i'll do that now. closing this pull request.

eugenp: You can update a PR, you don't always need to close it and do a new one.

reymalahay: @eugenp, thank you for the advice.

PR #459, opened 2016-06-23
webjars demo project added

bhaskey:
webjars directory added to tutorial-master. project is added as
webjarsdemo inside that


tomjoebob:
@bhaskey You have some extraneous files in here that need to be removed before I can merge this. You should only have the .java files and pom.xml. The other jars, .properties files and executables shouldn't be part of the pull request.

bhaskey:
@tomjoebob removed properties files.

PR #476, opened 2016-07-05
Added code samples for the article Intro to QueryDSL

forketyfork:
Code samples for the article http://inprogress.baeldung.com/intro-to-querydsl/

I didn't put my code in the existing `querydsl` module, because that module uses Spring and a rather old and incompatible version of QueryDSL. My project `querydsl-intro` is much more simple, does not have a lot of dependencies and does not use Spring, so I put it into a separate module. But I can try to merge it with `querydsl`, if needed.

### DIFF
--- a/querydsl/pom.xml
+++ b/querydsl/pom.xml
@@ -14,30 +14,25 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <java.version>1.6</java.version>
-        <junit.version>4.10</junit.version>
-        <spring.version>3.1.0.RELEASE</spring.version>
-        <hibernate.version>4.3.11.Final</hibernate.version>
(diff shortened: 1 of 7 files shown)


pivovarit:
Looks good to go. I will do CR when I get home :)
Cheers!

eugenp:
I think that's fine - the only quick note I have is - it's worth seeing if the spring module can be upgraded to use a newer version of QueryDSL. That way you could simply have the code in the old module (but in a separate root package to separate it from everything else). 
If not - a new module is fine. 
Cheers, 
Eugen.

forketyfork:
Updated the PR. I've merged my project with querydsl module, this was not so hard, but I had to update all of the dependencies (new Hibernate requires new Spring, new Spring requires new slf4j etc), change the logging from log4j to logback, and update the old code due to some API differences.

PR #493, opened 2016-07-12
adding spring controller changes

prashant1067:


prashant1067:
adding spring controller

spring_controller/src/main/java/com/test/controller/TestController.java
```diff
@@ -0,0 +1,26 @@
+
+/**
+ * @author Prashant Dutta
+ *
+ */
+package com.test.controller;
```
pivovarit: Package name should be com.baeldung

pivovarit:
Please check if this code can't be placed in an already existing module.

pivovarit:
Also, please make short runnable tests for this. There is some Spring MVC testing tools, so this should be fairly easy

prashant1067:
Changed the package name and introduced from com.test.* to com.baledung.*

prashant1067:
Can't find a suitable module to move this code . Since this code is solely linked to controllers I believe that there should be a seperate module for this. Please do  advise on this.

pivovarit:
If spring-all is not suitable, you can start a new one 'spring-controller'

prashant1067:
I have done that.

pivovarit:
And when can we expect tests for this? mockMVC will be helpful

prashant1067:
Added test cases using mock mvc.

PR #504, opened 2016-07-14
Add missing modules

pivovarit:
... and clean some random stuff up.


pivovarit:
@eugenp this one is for you

PR #513, opened 2016-07-19
Added examples for the @RestClientTest article, fixed mail port

forketyfork:
Added examples for the article about @RestClientTest.
Also changed the port in mail tests from 25 to 8025, as opening sockets on ports below 1024 on many systems requires admin privileges, which affects the build (the tests did not run on my machine without sudo).


pivovarit:
Excellent work. Merging :+1:

PR #542, opened 2016-07-28
Spring MVC Email application added.

oreva:
Pull request to add the SpringMVCEmail application.


spring-mvc-email/pom_old.xml
```diff
@@ -0,0 +1,142 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pkhanal: do we still need this file? Can we delete it?

oreva: Yes, sure. We can delete it.

spring-mvc-email/src/main/java/com/baeldung/spring/app/config/AppConfig.java
```diff
@@ -0,0 +1,96 @@
...
+    private Properties javaMailProperties() {
+        Properties props = new Properties();
+        props.setProperty("mail.transport.protocol", "smtps");
+        props.setProperty("mail.smtp.port", "25");
+
+        props.put("mail.smtp.auth", "true");
+        props.put("mail.smtp.starttls.enable", "true");
+        props.put("mail.smtp.starttls.required", "true");
+        return props;
+    }*/
```
pkhanal: Can we remove the commented code block?

spring-mvc-email/src/main/java/com/baeldung/spring/mail/EmailServiceImpl.java
```diff
@@ -0,0 +1,114 @@
...
+        SimpleMailMessage mailMessage = new SimpleMailMessage(template);
+
+        mailMessage.setFrom(mailObject.getFrom());
+        mailMessage.setTo(mailObject.getTo());
+        mailMessage.setSubject(mailObject.getSubject());
+        mailMessage.setText(String.format(template.getText(), mailObject.getText()));
+
+        return mailMessage;
+    }*/
+}
```
pkhanal: Same as before. If we don't need the commented code block anymore, please remove it.

pkhanal:
Here is what I would suggest for reader to run the sample and do a quick test.
Create a simple html page with three buttons - 
1. A text field that takes destination address.
2. A Send button to send a simple email.
3. A Send button to send email with attachment. You can use a static file for attachment.
4. For template based email scenario, you can create a simple form with couple of text fields that serve as data to the template.

That way, reader can run the application and properly follow the scenarios that article is trying to explain.
Remove all the unnecessary code from the project.

oreva:
All the latest improvements were implemented.

pivovarit:
I assume this one is dead. Let me know if it is not

PR #610, opened 2016-08-15
LoggerInterceptor plus tests

maibin:


spring-security-rest-full/src/main/resources/webSecurityConfig.xml
```diff
@@ -32,5 +32,9 @@
     </authentication-manager>
 
     <global-method-security pre-post-annotations="enabled"/>
+    
+<!--     <mvc:interceptors>
```
pivovarit: This could be easily removed

PR #635, opened 2016-08-25
Removed @Lazy annotation from @Controller.

christian-raedel:
Small fix related to *BAEL-168*.


slavisa-baeldung:
merged already

PR #639, opened 2016-08-26
Source Code for Wicket Intro Article

diegopm:
The source code for the Wicket Intro article.


wicket-intro/CafeAddress/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
slavisa-baeldung: try to copy log setup from any other module, we use logback and slf4j as a facade

wicket-intro/CafeAddress/src/main/java/com/baeldung/wicket/examples/cafeaddress/CafeAddress.java
```diff
@@ -0,0 +1,72 @@
+package com.baeldung.wicket.examples.cafeaddress;
```
slavisa-baeldung: This class is properly formatted, the same style you should use on the other classes and view templates. Usually, importing standard eclipse formatter to the IDE your using works out. 

Formatting we should apply on templates as well

wicket-intro/CafeAddress/src/main/resources/log4j2.xml
diegopm: It was generated by the maven command I used to build the scaffold. I'm not using logging in this example, so I will remove it.

PR #673, opened 2016-09-07
Spring integration with Tiles

kirangonella:
Hi Eugen,

Spring integration with Tiles code

Thank you,
Kiran.


kirangonella:
Hi Eugen,

Do help to merge the above code so that I can share the code link in the code.

PR #704, opened 2016-09-27
Best Practices while using Maps in Java

seemamani:
Maps in Java are used to represent associations between key-value pairs. These implementations of the java.util.Map interface contain specialized features that make it suitable for specific situations. This article explains the right ways to use maps to get the best out of them in different circumstances.


seemamani:
An article on the best practices to be followed while using Maps in Java

PR #723, opened 2016-10-05
custom user repository and test

lor6:


spring-security-custom-permission/src/test/java/org/baeldung/web/CustomUserDetailsServiceTest.java
```diff
@@ -0,0 +1,75 @@
...
+            user.setUsername(USERNAME);
+            user.setPassword(passwordEncoder.encode(PASSWORD));
+
+            myUserRepository.save(user);
+
+            UsernamePasswordAuthenticationToken auth = new UsernamePasswordAuthenticationToken(USERNAME2, PASSWORD);
+            authenticationProvider.authenticate(auth);
+        } finally {
+            myUserRepository.removeUserByUsername(USERNAME);
+        }
```
pkhanal: Seems like we can get rid of try-finally block from individual test and move the statement(s) in the finally block to teardown method.
```
@After
public void tearDown() {
    myUserRepository.removeUserByUsername(USERNAME);
}
```

pivovarit:
@lor6 can this be closed already?

PR #737, opened 2016-10-11
Removed unnecessary code and SOP statements

kirangonella:


spring-jms/src/main/java/com/baeldung/spring/jms/SampleListener.java
```diff
@@ -10,16 +10,16 @@
 
 public class SampleListener implements MessageListener {
 
+    private JmsTemplate jmsTemplate;
     public JmsTemplate getJmsTemplate() {
-        return getJmsTemplate();
+        this.jmsTemplate = jmsTemplate;
```
pivovarit: You changed the getter into a setter without changing a name. Is this intentional?

PR #750, opened 2016-10-15
BAEL 317: Setting up EJB

pritambanerjee999:
EJB Client and EJB Remote


pritambanerjee999:
Slavisa can you please try with this?

PR #754, opened 2016-10-17
BAEL-276: Review comment changes

sunilgulabani:
Please review and merge.

### DIFF
--- a/core-java/src/main/java/com/baeldung/java/nio/selector/EchoClient.java
+++ b/core-java/src/main/java/com/baeldung/java/nio/selector/EchoClient.java
@@ -17,6 +17,11 @@ public class EchoClient {
         return instance;
     }
 
+    public static void stop() throws IOException {
+        client.close();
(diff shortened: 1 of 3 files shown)


core-java-8/src/test/java/com/baeldung/encoderdecoder/EncoderDecoder.java
```diff
@@ -8,31 +8,87 @@
...
+        StringBuilder requestParamsString = new StringBuilder();
+        requestParams.forEach((key, value) -> {
+            try {
+                requestParamsString
+                        .append(key)
+                        .append("=")
+                        .append(URLEncoder.encode(value,
+                                StandardCharsets.UTF_8.toString()))
+                        .append("&");
+            } catch (UnsupportedEncodingException e) {
```
pivovarit: Let's not try-catch exceptions in tests. We can simply add this to a method signature just like it was before

PR #769, opened 2016-10-24
Test cases for examples in article "A guide to Java 8 forEach"

arekssu:
Here are the test cases of the code used for the article "A guide to Java 8 forEach"


arekssu:
Here are the examples used in article "Guide to Java 8 forEach"

PR #773, opened 2016-10-26
All Eclipse modules

maibin:


annotations/annotation-user/pom.xml
```diff
@@ -27,6 +27,13 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+        
+        <dependency>
+			<groupId>org.projectlombok</groupId>
+			<artifactId>lombok</artifactId>
+			<version>1.16.10</version>
+			<scope>provided</scope>
+		</dependency>
```
maibin: The Java Builder Pattern does not work without additional plugins in Eclipse, Lombok provides @Builder annotation, that works fine.

spring-all/src/main/webapp/WEB-INF/web.xml
```diff
@@ -9,11 +9,11 @@
         <servlet-class>
             org.springframework.web.servlet.DispatcherServlet
         </servlet-class>
-        <load-on-startup>1</load-on-startup>
         <init-param>
             <param-name>contextConfigLocation</param-name>
             <param-value>/WEB-INF/test-mvc.xml</param-value>
         </init-param>
+        <load-on-startup>1</load-on-startup>
```
maibin: Yes, Eclipse is order specific, load on startup needs to be the last one.

spring-rest/src/main/java/org/baeldung/config/WebConfig.java
```diff
@@ -42,6 +43,7 @@ public void configureMessageConverters(final List<HttpMessageConverter<?>> messa
         // messageConverters.add(new MappingJackson2XmlHttpMessageConverter(builder.createXmlMapper(true).build()));
         messageConverters.add(new ProtobufHttpMessageConverter());
         messageConverters.add(new KryoHttpMessageConverter());
+        messageConverters.add(new ByteArrayHttpMessageConverter());
```
maibin: It's connected with previous article, Top 7 mistakes, as in one example I am returning image itself, and this converts byte to browser friendly version)

spring-rest/src/main/webapp/WEB-INF/api-servlet.xml
```diff
@@ -1,48 +1,55 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<beans xmlns="http://www.springframework.org/schema/beans" xmlns:mvc="http://www.springframework.org/schema/mvc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xmlns:context="http://www.springframework.org/schema/context"
-    xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd 
+<beans xmlns="http://www.springframework.org/schema/beans"
+	xmlns:mvc="http://www.springframework.org/schema/mvc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:context="http://www.springframework.org/schema/context"
+	xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd 
     http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context.xsd 
-    http://www.springframework.org/schema/mvc http://www.springframework.org/schema/mvc/spring-mvc-4.0.xsd"
```
maibin: Again, this is connected with previous article, Top 7 mistakes, as we needed also formatter in XML config. I added this:
<bean
				class="org.springframework.http.converter.ByteArrayHttpMessageConverter">
				<property name="supportedMediaTypes">
					<list>
						<value>image/jpeg</value>
					</list>
				</property>
			</bean>

spring-freemarker/pom.xml
```diff
@@ -30,11 +29,11 @@
...
 		</dependency>
-		
+
 		<dependency>
-    		<groupId>org.springframework</groupId>
-    		<artifactId>spring-context-support</artifactId>
-    		<version>${spring.version}</version>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-context-support</artifactId>
+			<version>${spring.version}</version>
```
maibin: OK

annotations/annotation-user/pom.xml
eugenp: Sure, but the idea here is not to modify the underlying logic - just to fix the issues that pop up. 
For example, this code is used here: http://www.baeldung.com/java-annotation-processing-builder
Actually, this is an important point across the board - we have to be careful modifying the code because it is likely used by an article on the site. 
Hope that helps. Cheers, 
Eugen.

PR #797, opened 2016-11-01
fixed network interface tetst

egimaben:


core-java/src/test/java/com/baeldung/java/nio2/FileTest.java
```diff
@@ -152,11 +158,16 @@ public void givenPath_whenFailsToDeleteNonEmptyDir_thenCorrect() throws IOExcept
 
     }
 
-    @Test(expected = NoSuchFileException.class)
```
pivovarit: Can you remove this change?

egimaben: damn, I don't recall making this change. Anyways just updated

PR #815, opened 2016-11-08
fixed imports for sslcontext for apache 4.4

nbosecker:


KevinGilmore:
Six of the classes are deprecated since 4.3 of the httpclient module, and we're now using 4.5.

PR #817, opened 2016-11-09
Apache CXF Aegis data bindings

nguyennamthai:


pivovarit:
I assume that this one is not relevant anymore, let me know if not.

PR #830, opened 2016-11-14
Spring Integration Intro

k0l0ssus:


slavisa-baeldung:
hey hey Tayo, I've copied the main config to spring-integration module on the main branch. Let us try to use relative path so that the code is immediately runnable. Also, I'm bootsrapping integration from the test. Still file is not copied, can you check it out, best

pivovarit:
@k0l0ssus hey, is this one still valid?

k0l0ssus:
Not any more @pivovarit  - these files are no longer necessary given what Slavisa has already put in the project

PR #832, opened 2016-11-14
BAEL-385: core-java Properties class

DianeDuan:
add a few demos for the article: The Java Properties Class


DianeDuan:
replace main methods with unit tests

PR #839, opened 2016-11-15
asyncronous channel apis

egimaben:


eugenp:
Hey @egimaben @pivovarit - we should definitely try a squash and merge here.

pivovarit:
@eugenp I am always doing this :)

PR #840, opened 2016-11-16
Update FileCopyConfig.java

k0l0ssus:
Changed the extension search to look for a wildcard extension
Set expectReply(false)[service activator will throw exception after copying file without it)
Used class reference instead of FQCN


k0l0ssus:
The file won't be copied because the extension is a direct reference (.jpg) instead of a wildcard (*.jpg). Also changed the context loading to use the Class instead of the FQCN

PR #844, opened 2016-11-17
Improvement for Spring Security Basic Authentication.

oreva:
Java-based configuration implemented.


spring-security-basic-auth/src/main/java/org/baeldung/spring/SecSecurityConfigJava.java
```diff
@@ -0,0 +1,40 @@
...
+    public void configureGlobal(AuthenticationManagerBuilder auth) throws Exception {
+        auth
+          .inMemoryAuthentication()
+          .withUser("user1").password("user1Pass").authorities("ROLE_USER");
+    }
+
+    @Override
+    protected void configure(HttpSecurity http) throws Exception {
+        http.authorizeRequests()
+                .antMatchers("/securityNone").permitAll()
```
pkhanal: Why do we need this? 
Are we using any path **/securityNone**?

spring-security-basic-auth/src/main/java/org/baeldung/security/filter/configuration/CustomWebSecurityConfigurerAdapter.java
```diff
@@ -1,17 +0,0 @@
...
-
-@Configuration
-public class CustomWebSecurityConfigurerAdapter extends WebSecurityConfigurerAdapter {
-
-    @Override
-    protected void configure(HttpSecurity http) throws Exception {
-        http.addFilterAfter(new CustomFilter(), BasicAuthenticationFilter.class);
-    }
-
-}
```
pkhanal: The class is being referred in the following article. 
http://www.baeldung.com/spring-security-custom-filter

It might be confusing to reader if we delete the existing class. May be we should just keep the same class and add code here instead.

spring-security-basic-auth/src/main/java/org/baeldung/spring/SecSecurityConfigXML.java
```diff
@@ -4,12 +4,12 @@
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.ImportResource;
```
pkhanal: Organize imports. We can probably remove unused imports.

spring-security-basic-auth/src/main/java/org/baeldung/security/filter/configuration/CustomWebSecurityConfigurerAdapter.java
oreva: Done in the last commit.

spring-security-basic-auth/src/main/java/org/baeldung/spring/SecSecurityConfigXML.java
oreva: Done.

spring-security-basic-auth/src/main/java/org/baeldung/spring/SecSecurityConfigJava.java
oreva: This path was present in the XML config, so I just did the same in the Java config. I can see from the application that this path we can enter manually to see that it is not under security while homepage is under security.

pkhanal: ok..thanks for clarifying

PR #861, opened 2016-11-27
Added BAEL-425 Intro to Servets module named javax-servlets

sbmaggarwal:


javax-servlets/web/index.jsp
```diff
@@ -0,0 +1,30 @@
...
+        <option>Venus</option>
+        <option>Mars</option>
+    </select>
+    <br/><br/>
+
+    <input type="submit" value="Submit"/>
+</form>
+
+</body>
+</html>
```
mklew: Why do you need a jsp file since this article is about Servlets?

PR #866, opened 2016-11-29
BAEL-453: How to use the Spring FactoryBean - use custom bean name

DianeDuan:


spring-core/src/test/java/com/baeldung/factorybean/FactoryBeanJavaConfigTest.java
```diff
@@ -16,7 +18,9 @@
 
     @Resource
```
pivovarit: @Autowired should be here (we are not using a particular name)

spring-core/src/test/java/com/baeldung/factorybean/FactoryBeanJavaConfigTest.java
```diff
@@ -16,7 +18,9 @@
...
     private Tool tool;
-    @Resource(name = "&toolFactory")
+
+    @Autowired
+    @Qualifier("&tool")
```
pivovarit: Those two lines should be replaced by @Resource("&tool") because we are referencing a one particular resource by its name

PR #871, opened 2016-12-01
BAEL-7: Join and Split Collections

tochoromero:


core-java/src/test/java/org/baeldung/java/collections/JoinSplitCollectionsUnitTest.java
```diff
@@ -0,0 +1,178 @@
...
+import org.junit.Test;
+
+public class JoinSplitCollectionsUnitTest {
+
+	@Test
+	public void whenJoiningTwoArrays_thenJoined() {
+		String[] animals1 = new String[] { "Dog", "Cat" };
+		String[] animals2 = new String[] { "Bird", "Cow" };
+		String[] result = Stream.concat(Arrays.stream(animals1), Arrays.stream(animals2))
+				.toArray(size -> new String[size]);
```
pivovarit: size -> new String[size] can be replaced with a methodref: String[]::new

core-java/src/test/java/org/baeldung/java/collections/JoinSplitCollectionsUnitTest.java
```diff
@@ -0,0 +1,178 @@
...
+	
+		assertEquals(result, "Apple, Orange, Grape");
+	}
+
+	@Test
+	public void whenSplitCollectionHalf_thenConverted() {
+		Collection<Integer> numbers = Arrays.asList(1, 2, 3, 4, 5, 6, 7, 8, 9, 10);
+		Collection<Integer> result1 = new ArrayList<>();
+		Collection<Integer> result2 = new ArrayList<>();
+		AtomicInteger count = new AtomicInteger();
```
pivovarit: Why do we need an AtomicInteger here?

core-java/src/test/java/org/baeldung/java/collections/JoinSplitCollectionsUnitTest.java
```diff
@@ -0,0 +1,178 @@
...
+		assertEquals(result.get(2), "Paul Smith");
+		assertEquals(result.get(3), "Susan Anderson");
+	}
+
+	@Test
+	public void whenConvertCollectionToStringMultipleSeparators_thenConverted() {
+		String fruits = "Apple. , Orange, Grape. Lemon";
+
+		Collection<String> result = Arrays.stream(fruits.split("[,|.]"))
+				.map(String::trim).filter(next -> !next.isEmpty())
```
pivovarit: Let's give .filter() a separate line

core-java/src/test/java/org/baeldung/java/collections/JoinSplitCollectionsUnitTest.java
tochoromero: It can be replaced indeed, will make the change

core-java/src/test/java/org/baeldung/java/collections/JoinSplitCollectionsUnitTest.java
tochoromero: Since the Streams are not concurrent it doesn't really need to be an AtomicInteger.
It becomes indispensable once the stream is parallel. So I guess it can be changed here to int, but it is a good practice and future-proof to use an AtomicInteger.

core-java/src/test/java/org/baeldung/java/collections/JoinSplitCollectionsUnitTest.java
tochoromero: Whoops, I still need to do the formatting by hand

core-java/src/test/java/org/baeldung/java/collections/JoinSplitCollectionsUnitTest.java
tochoromero: Tried to just change it to an int, but it has some problems. You have to make it final in order to use it inside the stream, but then because it is final you cannot just increment it. So I guess AtomicInteger is the best option.

PR #891, opened 2016-12-13
LiveTest included

sbmaggarwal:
root package renamed to com.baeldung.servlets, Test renamed to LiveTest so that it is ignored by CI.


javax-servlets/src/test/java/com/baeldung.servlets/LiveTest.java
```diff
@@ -13,7 +13,7 @@
 
 import static org.junit.Assert.assertEquals;
 
-public class FormServletTest {
+public class LiveTest {
```
pivovarit: I meant that it should stick to convention *LiveTest so let's name it FormServletLiveTest

sbmaggarwal:
Changed test name to FormServletLiveTest

PR #912, opened 2016-12-21
Spark Java Article BAEL-498

joshiparthin:
For Spark Java Article.


joshiparthin:
@KevinGilmore I think this is because I created spark-java module by using the the new module creation tool in eclipse. I will try to update this push.

joshiparthin:
@KevinGilmore  Please see the changes in the latest commit: https://github.com/eugenp/tutorials/pull/912/commits/608c9784a95e534735780508ec2c918463d6a77a

spark-java/src/main/java/com/baeldung/sparkjava/StatusResponse.java
```diff
@@ -0,0 +1,21 @@
+package com.baeldung.sparkjava;
+
+public enum StatusResponse {
+    SUCCESS ("Success"),
+    ERROR ("Error");
+    
+    private String status;
```
KevinGilmore: This should be a final field.

spark-java/src/main/java/com/baeldung/sparkjava/StatusResponse.java
```diff
@@ -0,0 +1,21 @@
...
+   
+    StatusResponse(String status) {
+        this.status = status;
+    }
+    
+    public String getStatus() {
+        return status;
+    }
+
+    public void setStatus(String status) {
```
KevinGilmore: An enum should be immutable and never contain a setter.

spark-java/src/main/java/com/baeldung/sparkjava/UserService.java
```diff
@@ -0,0 +1,13 @@
+package com.baeldung.sparkjava;
+
+import java.util.Collection;
+import java.util.HashMap;
+
+public interface UserService {
+    public void addUser (User user) ;
+    public Collection<User> getUsers () ;
+    public User getUser (String id) ;
+    public User editUser (String id, HashMap userArg) throws Exception;
```
KevinGilmore: 1. Change HashMap to Map (program to the interface, not to the implementation).
Or consider that it is more common in a REST service for the PUT operation to accept an entire JSON object instead of only the fields to be changed like you do here using the Map, and instead you can just implement it as public void User editUser(User user) ...

2. Don't declare that something throws 'Exception'. Instead, create your own exception class (e.g. UserException) and throw that.

spark-java/src/main/java/com/baeldung/sparkjava/UserServiceMapImpl.java
```diff
@@ -0,0 +1,45 @@
...
+    
+    public Collection<User> getUsers () {
+        return  userMap.values();
+    }
+    
+    public User getUser (String id) {
+        return userMap.get(id);
+    }
+    
+    public User editUser (String id, HashMap userArg) throws Exception{
```
KevinGilmore: This implementation becomes trivial if you change the method to accept a User object instead of an id and Map.

spark-java/src/main/java/com/baeldung/sparkjava/UserServiceMapImpl.java
```diff
@@ -0,0 +1,45 @@
...
+        User toEdit = userMap.get(id);
+        if (toEdit == null )
+            return null;
+        toEdit.setEmail((userArg.get("email")!=null) ? (String) userArg.get("email") : toEdit.getEmail() );
```
KevinGilmore: This comment applies to the next three lines as well if you stick with a Map implementation:
Notice that when the key is not found in the map, you are unnecessarily setting the property to the value that it already has. Just use 'if' statements and set the properties only when the key exists in the map.

spark-java/src/main/java/com/baeldung/sparkjava/StatusResponse.java
joshiparthin: Done.

spark-java/src/main/java/com/baeldung/sparkjava/UserService.java
joshiparthin: Changes to Map and Created a new Exception Class. Changed the flow of the function a bit. 

About passing the JSON itself, Actually it passes the JSON only, If you see line 34 of SparkRestExample.java (put("/users/:id", (request, response) -> {..) it converts JSON to map and hence map only has fields which are sent to be edited.

I was looking for better ways to do this since inception and have also shared my concerned earlier for the same. 

Let me know if we can still make this better after this above edits i've done in the editUser function.

spark-java/src/main/java/com/baeldung/sparkjava/UserServiceMapImpl.java
joshiparthin: I wanted to do the same and hence wanted to go for reflection solution, if you remember I asked you earlier. My problem was what happens to those fields which are not "set" and passed in PUT request. 

Hence the Map solutions. Please suggest if current approach is not doable.

spark-java/src/main/java/com/baeldung/sparkjava/UserServiceMapImpl.java
joshiparthin: Done. I thought it would not look "cool" to have so many if's in the function. But then I realized as you rightly guided, we cannot compromise simplicity of the code in making it look more fancy. Of late I realized, rather than fancy, it looked more complicated. Thanks for guiding.

spark-java/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.baeldung</groupId>
+  <artifactId>spark-java</artifactId>
+  <version>0.1.0-SNAPSHOT</version>
+  <name>spark-java</name>
+  <url>http://maven.apache.org</url>
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
KevinGilmore: The project does not compile when I try to build locally with maven because the project uses Lambdas and maven is trying to compile to Java 1.5.
You have to set the version to 1.8 in the pom. Check some other project to see how they accomplish this.

spark-java/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+  </properties>
+  <dependencies>
+    <dependency>
```
KevinGilmore: Be consistent with the indentation in the pom -- either always indent with tabs, or always with spaces, in increments of four spaces.

spark-java/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
+			<version>3.8.1</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>com.sparkjava</groupId>
+			<artifactId>spark-core</artifactId>
+			<version>2.5.3</version>
```
KevinGilmore: The latest spark-core is 2.5.4

spark-java/src/main/java/com/baeldung/sparkjava/SparkRestExample.java
```diff
@@ -0,0 +1,65 @@
+package com.baeldung.sparkjava;
+
+import static spark.Spark.*;
+
+import java.util.HashMap;
```
KevinGilmore: Remove the unused imports

spark-java/src/main/java/com/baeldung/sparkjava/User.java
```diff
@@ -0,0 +1,61 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package com.baeldung.sparkjava;
+
+import java.util.Map;
```
KevinGilmore: Remove the unused imports

spark-java/src/main/java/com/baeldung/sparkjava/UserService.java
```diff
@@ -0,0 +1,14 @@
+package com.baeldung.sparkjava;
+
+import java.util.Collection;
```
KevinGilmore: Remove the unused imports

spark-java/src/main/java/com/baeldung/sparkjava/UserService.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public interface UserService {
+    public void addUser (User user) ;
+    public Collection<User> getUsers () ;
+    public User getUser (String id) ;
+    public User editUser (User user) throws Exception;
```
KevinGilmore: Declare that it throws UserException

spark-java/src/main/java/com/baeldung/sparkjava/UserServiceMapImpl.java
```diff
@@ -0,0 +1,64 @@
+package com.baeldung.sparkjava;
+
+import java.util.Arrays;
```
KevinGilmore: Remove the unused imports

spark-java/src/main/java/com/baeldung/sparkjava/UserServiceMapImpl.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class UserServiceMapImpl  implements UserService{
+    private HashMap<String, User> userMap;
+    
+    public UserServiceMapImpl() {
```
KevinGilmore: Use the @Overrides annotation on all the public methods that you are implementing from the interface

spark-java/pom.xml
joshiparthin: @KevinGilmore  Done

spark-java/src/main/java/com/baeldung/sparkjava/SparkRestExample.java
joshiparthin: Done

spark-java/src/main/java/com/baeldung/sparkjava/User.java
joshiparthin: Done

spark-java/src/main/java/com/baeldung/sparkjava/UserService.java
joshiparthin: Done

spark-java/src/main/java/com/baeldung/sparkjava/UserService.java
joshiparthin: Done

spark-java/src/main/java/com/baeldung/sparkjava/UserServiceMapImpl.java
joshiparthin: Done

spark-java/src/main/java/com/baeldung/sparkjava/UserServiceMapImpl.java
joshiparthin: Done

PR #973, opened 2017-01-08
BAEL-260	 Create PDF in Java

pezetem:


pivovarit:
@pezetem this one is not needed anymore, right? let me know

pezetem:
@pivovarit sure can be deleted

PR #990, opened 2017-01-12
String to char and vice versa 

Noroi5:
Code samples to show the conversion from String to char array and vice versa.


Noroi5:
Including Junit testcase as mentioned in the review.

Noroi5:
Review changes included.

PR #996, opened 2017-01-14
BAEL-390: JSON Processing in Java EE 7

felipe-gdr:
Accompanying code for the JSON Processing in Java EE 7


pivovarit:
I assume this one is no longer valid. let me know if it is not

PR #999, opened 2017-01-15
file upload max size controller, handler

lor6:


spring-mvc-forms/pom.xml
```diff
@@ -42,6 +42,11 @@
         </dependency>
 
 
+        <dependency>
+        	<groupId>commons-fileupload</groupId>
+        	<artifactId>commons-fileupload</artifactId>
+        	<version>1.3.2</version>
```
KevinGilmore: Put the version in a property like in the other dependencies.

PR #1007, opened 2017-01-17
Bael 382 kotlin

tomekl007:


pedja4:
Had to configure java sources in pom.xml

PR #1033, opened 2017-01-22
fix slope one bug

Youmoo:
Can test against this sample data:

image


pivovarit:
@Youmoo hey, I assume this one is no longer valid. If it is, feel free to reopen it(or just let me know if you cant' do this)

PR #1073, opened 2017-01-29
Solr w Apache SolrJ

nbosecker:


nbosecker:
I've updated the article and the code as suggested. I agree about the try/catch blocks, I think by default I always use them when Eclipse complains due to habit! So I renamed the test file and removed the try/catch blocks. 

As an aside, I always have problems with Git so I'm not sure if I updated the commit correctly. I thought I deleted the old test but it doesn't look right. Let me know.

spring-data-solr/src/test/java/com/baeldung/solrjava/SolrJavaIntegrationTest.java
```diff
@@ -0,0 +1,71 @@
+package org.baeldung.solrjava;
```
KevinGilmore: The package name should be com.baeldung.solrjava

spring-data-solr/pom.xml
```diff
@@ -51,6 +51,11 @@
             <version>${spring.version}</version>
             <scope>test</scope>
         </dependency>
+  	<dependency>
+  	    <groupId>org.apache.solr</groupId>
+  	    <artifactId>solr-solrj</artifactId>
+ 	    <version>6.4.0</version>
```
KevinGilmore: The latest version is 6.4.1.
Extract the version into a property.

nbosecker:
I've updated the pom and corrected the incorrect package path.

spring-data-solr/src/test/java/com/baeldung/solrjava/SolrJavaIntegrationTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+
+        solr = new HttpSolrClient("http://localhost:8983/solr/bigboxstore");
+        solr.setParser(new XMLResponseParser());
+    }
+
+    @Test
+    public void givenAdd_thenVerifyAdded() throws SolrServerException, IOException {
```
KevinGilmore: This test should probably be called "whenAdd_thenVerifyAdded".
Remember, in the "givenX_whenY_thenZ" convention, the "givenX" is optional, but the "whenY" should always be there.

spring-data-solr/src/test/java/com/baeldung/solrjava/SolrJavaIntegrationTest.java
```diff
@@ -0,0 +1,70 @@
...
+        assertEquals(docList.getNumFound(), 1);
+
+        for (SolrDocument doc : docList) {
+            assertEquals((String) doc.getFieldValue("id"), "123456");
+            assertEquals((Double) doc.getFieldValue("price"), (Double) 599.99);
+        }
+    }
+
+    @Test
+    public void givenDelete_thenVerifyDeleted() throws SolrServerException, IOException {
```
KevinGilmore: Same as above -- whenDelete_then...

PR #1085, opened 2017-02-01
Fix to Unit test for Hibernate  one to many 

mobiblunt:
Fixed unit test for the hibernate one to many tutorial.


spring-hibernate4/src/test/java/com/baeldung/hibernate/oneToMany/main/HibernateOneToManyAnnotationMainTest.java
```diff
@@ -11,30 +20,56 @@
...
-    @After
-    public void tearDown() {
+    
+
+    @Test
+    public void testAddItemsToCart() {
+        Cart cart = new Cart();
+        Items item1 = new Items("I10", 10, 1, cart);
+        Set<Items> itemsSet = new HashSet<Items>();
+        assertEquals(0, itemsSet.size());
```
maibin: This is standard Java feature. It should not be a part of the test, the same for checking if the item1 object is null

spring-hibernate4/src/test/java/com/baeldung/hibernate/oneToMany/main/HibernateOneToManyAnnotationMainTest.java
```diff
@@ -11,30 +20,56 @@
...
-        fail("The test failed.");
+    public void testSaveCart(){
+       Cart cart = new Cart();
+       Items item1 = new Items("I10", 10, 1, cart);
+       Set<Items> itemsSet = new HashSet<Items>();
+       itemsSet.add(item1);
+       cart.setItems(itemsSet);
+       session.persist(cart);
+       session.getTransaction().commit();
+       session.close();
```
maibin: There is no single Assertion in this test.
You should be able to get the cart objects before the save test, check the size, then save and it should be incremented by 1.
Let me know if that makes sense.

pom.xml
```diff
@@ -121,7 +121,6 @@
         <module>spring-exceptions</module>
         <module>spring-freemarker</module>
         <module>spring-hibernate3</module>
-        <module>spring-hibernate4</module>
```
maibin: Why did you exclude this from the tutorials pom?
Please revert it.

PR #1089, opened 2017-02-02
Test Suffix In Class Name

dotun:


pivovarit:
It looks like those changes were already integrated. Closing

PR #1118, opened 2017-02-05
Serializing/Deserializing Optional with Jackson

lsoarestd:
http://jira.baeldung.com/browse/BAEL-665


pivovarit:
I see those changes were already incorporated

PR #1129, opened 2017-02-07
Adding Hibernate 5 Project

Sandeep4odesk:


spring-hibernate5/src/main/resources/persistence-mysql.properties
```diff
@@ -0,0 +1,13 @@
+# jdbc.X
+jdbc.driverClassName=com.mysql.jdbc.Driver
```
KevinGilmore: When running the FooServiceBasicIntegrationTest, I get the following message in the console:
Loading class `com.mysql.jdbc.Driver'. This is deprecated. The new driver class is `com.mysql.cj.jdbc.Driver'

spring-hibernate5/src/main/resources/persistence-mysql.properties
```diff
@@ -0,0 +1,13 @@
...
+jdbc.url=jdbc:mysql://localhost:3306/spring_hibernate4_01?createDatabaseIfNotExist=true
```
KevinGilmore: Should this change to spring_hibernate5...?

spring-hibernate5/src/main/java/com/baeldung/spring/PersistenceConfig.java
```diff
@@ -0,0 +1,177 @@
...
+@EnableJpaRepositories(basePackages = { "com.baeldung.persistence" }, transactionManagerRef = "jpaTransactionManager")
+@EnableJpaAuditing
+@PropertySource({ "classpath:persistence-mysql.properties" })
+@ComponentScan({ "com.baeldung.persistence" })
+public class PersistenceConfig {
+
+    @Autowired
+    private Environment env;
+
+    public PersistenceConfig() {
```
KevinGilmore: Since this class does not extend another class, we do not need this constructor.

spring-hibernate5/src/main/java/com/baeldung/spring/PersistenceXmlConfig.java
```diff
@@ -0,0 +1,18 @@
...
+import org.springframework.context.annotation.ImportResource;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
+
+@Configuration
+@EnableTransactionManagement
+@ComponentScan({ "com.baeldung.persistence.dao", "com.baeldung.persistence.service" })
+@ImportResource({ "classpath:hibernate5Config.xml" })
+public class PersistenceXmlConfig {
+
+    public PersistenceXmlConfig() {
```
KevinGilmore: Same as before - don't need to add a no-arg constructor and invoke super() if the class does not extend another class (other than Object).

Sandeep4odesk:
Changes committed as per comments. Please check and revert.

spring-hibernate5/src/test/java/com/baeldung/persistence/service/FooStoredProceduresIntegrationTest.java
```diff
@@ -0,0 +1,115 @@
+package com.baeldung.persistence.service;
+
+import java.util.List;
+
+import static org.apache.commons.lang3.RandomStringUtils.randomAlphabetic;
+import static org.junit.Assert.assertEquals;
+
+import org.hibernate.Query;
```
KevinGilmore: the Query import is not used and appears to be deprecated

spring-hibernate5/pom.xml
```diff
@@ -0,0 +1,239 @@
...
+    </build>
+
+    <properties>
+        <!-- Spring -->
+        <org.springframework.version>4.3.5.RELEASE</org.springframework.version>
+	    <org.springframework.security.version>4.2.1.RELEASE</org.springframework.security.version>
+	    <org.springframework.data.version>1.10.6.RELEASE</org.springframework.data.version>
+        
+        <!-- persistence -->
+        <hibernate.version>5.2.6.Final</hibernate.version>
```
KevinGilmore: Version 5.2.8.Final is the latest

spring-hibernate5/pom.xml
```diff
@@ -0,0 +1,239 @@
...
+        <hibernate-envers.version>${hibernate.version}</hibernate-envers.version>
+        <mysql-connector-java.version>6.0.5</mysql-connector-java.version>
+        <tomcat-dbcp.version>8.5.9</tomcat-dbcp.version>
```
KevinGilmore: Latest is 8.5.11

Sandeep4odesk:
I implemented the changes you asked for, but could you look for whenPersistDetached_thenThrowsException inside SaveMethodsTest.java. I am not getting to the fix for this test case. Issue seems to be with transaction commit, but not able to understand where to fix this.

KevinGilmore:
Sandeep, I think it would be sufficient to change the expected exception to PersistenceException.

Sandeep4odesk:
Please try now, I have tried by changing the type of Exception. Beside that tearDown() is throwing TransactionException, which is troubling me. I have tried by surrounding it in try-catch block. It is printing the stackTrace, please check and suggest for this issue.

PR #1137, opened 2017-02-08
Spring data ldap

MahamedReda:


pivovarit:
@MahamedReda this one is not valid anymore, right? let me know if it needs to be reopened

PR #1178, opened 2017-02-15
apache bval project

lor6:


apache-bval/src/test/java/com/baeldung/validation/ValidationTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Test
+    public void givenInvalidPassword_whenValidatePassword_thenConstraintViolation() {
+        User user = new User("ana@yahoo.com", "password", "Ana", 20);
+        Set<ConstraintViolation<User>> violations = validator.validateProperty(user, "password");
+        assertEquals("message incorrect", "Invalid password", violations.iterator()
+            .next()
+            .getMessage());
+
+        user.setPassword("password#");
```
KevinGilmore: I think these three lines (82-84) were left over from the original, so they can be removed.

PR #1186, opened 2017-02-17
BAEL-9

Thoughtscript:


spring-boot-servlet/src/main/java/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,3 @@
+Manifest-Version: 1.0
+Main-Class: com.jspbaeldung.ApplicationMain
```
KevinGilmore: not the correct class?

spring-boot-servlet/src/main/java/com/baeldung/props/Constants.java
```diff
@@ -0,0 +1,20 @@
+package com.baeldung.props;
+
+import org.springframework.beans.factory.annotation.Autowired;
+
+import java.util.Properties;
+
+public final class Constants {
```
KevinGilmore: Is this class still being used? I didn't see any references to it or any of the public properties.

spring-boot-servlet/src/main/java/com/baeldung/props/PropertyLoader.java
```diff
@@ -0,0 +1,27 @@
+package com.baeldung.props;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+public class PropertyLoader {
```
KevinGilmore: Is this class still used, other than by the Constants class?
If you decide that you no longer need the Constants class, then do you still need this one?

spring-boot-servlet/src/main/java/com/baeldung/props/PropertyLoader.java
```diff
@@ -0,0 +1,27 @@
...
+        Properties prop = new Properties();
+        InputStream input = null;
+        try {
+          input = getClass().getResourceAsStream(file);
+          prop.load(input);
+        } catch (IOException ex) {
+          if (input != null) {
+            input.close();
+          }
+        } finally {
```
KevinGilmore: In Eclipse, I see a compiler warning (finally block does not complete normally).
May be a moot point if you no longer need this class.

spring-boot-servlet/pom.xml
```diff
@@ -0,0 +1,55 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>com.baeldung</groupId>
+    <artifactId>ServletReg</artifactId>
```
Thoughtscript: Should have caught that - thanks.

spring-boot-servlet/src/main/java/com/baeldung/props/Constants.java
Thoughtscript: This exists simply to illustrate how to set this up - otherwise no.

PR #1191, opened 2017-02-17
dcassiani@gmail.com

dcassiani:
Simple code with Constructor and Setter-based injections.

It has some Logback configuration as well, but otherwise, it is pretty clean.


dcassiani:
Switching for Annotations

PR #1196, opened 2017-02-19
Moved code to apache-solrj from spring-data-solr, updated code to 6.4.0 solrj, fixed dependencies in both modules

nbosecker:


apache-solrj/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+            <version>${junit.version}</version>
+            <scope>test</scope>
+    </dependency>
+  </dependencies>
+  
+  <build>
+    <sourceDirectory>src</sourceDirectory>
+    <plugins>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
```
KevinGilmore: missing the groupId tag for the maven-compiler-plugin: its value should be org.apache.maven.plugins (see one of the other projects for example)

apache-solrj/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+        <version>3.1</version>
+        <configuration>
+          <source/>
```
KevinGilmore: source and target should be set to 1.8

nbosecker:
Fixed the pom.xml, was able to mvn build locally. Let me know if still broken.

PR #1222, opened 2017-02-23
merging modules spring-mvc-forms into spring-mvc-simple [BAEL-596]

joshiparthin:


joshiparthin:
@pivovarit  sounds great.

PR #1225, opened 2017-02-23
models, repos, test

lor6:


spring-data-rest/src/main/test/com/baeldung/relationships/SpringDataRelationshipsTest.java
```diff
@@ -0,0 +1,103 @@
+package com.baeldung.relationships;
```
KevinGilmore: This should be in src/test/java rather than src/main/test -- otherwise the Maven build won't run it

spring-data-rest/src/test/java/org/baeldung/relationships/SpringDataRelationshipsTest.java
```diff
@@ -0,0 +1,103 @@
+package com.baeldung.relationships;
```
KevinGilmore: based on the package name, the file should be in **com**/baeldung/relationships instead of **org**/baeldung/relationships

PR #1251, opened 2017-02-26
Added more indexing/delete/query examples to code

nbosecker:
Also updated the document in WP


apache-solrj/src/test/java/com/baeldung/solrjava/SolrJavaIntegrationTest.java
```diff
@@ -36,15 +36,65 @@ public void whenAdd_thenVerifyAdded() throws SolrServerException, IOException {
         assertEquals(docList.getNumFound(), 1);
```
KevinGilmore: swap order of arguments in all assertEquals() calls

nbosecker: Done! good catch

PR #1337, opened 2017-03-08
#BAEL-505

aietcn:
Quick Example of Concurrent Test Execution in Spring 5


spring-5/pom.xml
```diff
@@ -9,7 +9,7 @@
     <packaging>jar</packaging>
 
     <name>spring-5</name>
```
pivovarit: We need to also add that module to the main pom.xml

aietcn:
main pom change appended @pivovarit

PR #1357, opened 2017-03-10
Refactor LDAPClient

pivovarit:


spring-ldap/src/main/java/com/baeldung/ldap/client/LdapClient.java
```diff
@@ -34,16 +27,20 @@ public void authenticate(final String username, final String password) {
...
-        List<String> users = ldapTemplate.search("ou=users", "cn=" + username, new AttributesMapper<String>() {
-            public String mapFromAttributes(Attributes attrs) throws NamingException {
-                return (String) attrs.get("cn").get();
-            }
-        });
-        return users;
+        return ldapTemplate.search(
+          "ou=users",
+          "cn=" + username,
+          (AttributesMapper<String>) attrs -> (String) attrs
```
pivovarit: This lambda instead of the anonymous inner class is the most important part

PR #1364, opened 2017-03-11
Finite automata example

amihaiemil:
Example code for [this](http://jira.baeldung.com/browse/BAEL-652) article


amihaiemil:
@pivovarit Here is the PR for the article :D

PR #1400, opened 2017-03-14
Bael 518 protobuffer

tomekl007:


pivovarit:
@tomekl007 should not this one be merged?

PR #1414, opened 2017-03-15
BAEL-707 Refactored the code as per comments

baljeet20:


baljeet20:
Build failure due to out of memory exception

PR #1468, opened 2017-03-21
#BAEL-636

aietcn:


aietcn:
@pivovarit I noticed that in the travis-ci integration, tests in the module spring-5 was not run. Thus the part added in this PR was not tested actually.

PR #1493, opened 2017-03-25
Spring State Machine

dkornishev:


spring-state-machine/src/test/java/com/baeldung/spring/statemachine/StateMachineTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Test
+    public void whenSimpleStringMachineActionState_thenActionExecuted() {
+        stateMachine.sendEvent("E3");
+        assertEquals("S3", stateMachine.getState().getId());
+
+        boolean acceptedE4 = stateMachine.sendEvent("E4");
+
+        assertTrue(acceptedE4);
+        Awaitility.await().atMost(5, TimeUnit.SECONDS).until(() -> assertEquals("S4", stateMachine.getState().getId()));
```
pivovarit: Those two assertions should be replaced with the await() call. No need for wrapping the assertion

PR #1511, opened 2017-03-27
Jackson Map Serialize/Deserialize

nbosecker:


jackson/src/test/java/com/baeldung/jackson/deserialization/JacksonMapDeserializeTest.java
```diff
@@ -0,0 +1,67 @@
...
+			throws JsonParseException, JsonMappingException, IOException {
+
+		final String jsonInput = "{\"key\": \"value\"}";
+		final ObjectMapper mapper = new ObjectMapper();
+		TypeReference<HashMap<String, String>> typeRef = new TypeReference<HashMap<String, String>>() {
+		};
+
+		final Map<String, String> map = mapper.readValue(jsonInput, typeRef);
+		final String expectedOutput = "{key=value}";
+		Assert.assertEquals(expectedOutput, map.toString());
```
KevinGilmore: A more appropriate test for these deserializations might be to assert that, given the key, you get the expected value from the map.
For example:
Assert.assertEquals("value", map.get("key"));

nbosecker:
Thanks Kevin, I've made the change you suggested. I realized in making that change that i had neglected equals()/hashCode() in the MyPair class, and have added those. I've also updated the reference in the article to include equals()/hashCode() commented out for space.

PR #1524, opened 2017-03-28
BAEL-732 : Converting Strings to Enums

chandravadans:
Code for Jira [732](http://jira.baeldung.com/browse/BAEL-732)


chandravadans:
Will make another PR with updated tests

PR #1533, opened 2017-03-29
Bael 766 flink

tomekl007:


libraries/pom.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
KevinGilmore: This line is only a change in spacing/indentation and should be restored to its original form.

libraries/src/main/java/com/baeldung/flink/LineSplitter.java
```diff
@@ -0,0 +1,19 @@
...
+import org.apache.flink.api.java.tuple.Tuple2;
+import org.apache.flink.util.Collector;
+
+public class LineSplitter implements FlatMapFunction<String, Tuple2<String, Integer>> {
+
+  @Override
+  public void flatMap(String value, Collector<Tuple2<String, Integer>> out) {
+    String[] tokens = value.toLowerCase().split("\\W+");
+    
+    for (String token : tokens) {
```
KevinGilmore: Why not use more of a Java 8 style to do this part.

libraries/src/main/java/com/baeldung/flink/WordCount.java
```diff
@@ -0,0 +1,21 @@
+package com.baeldung.flink;
+
+import org.apache.flink.api.java.DataSet;
+import org.apache.flink.api.java.ExecutionEnvironment;
+import org.apache.flink.api.java.aggregation.Aggregations;
+import org.apache.flink.api.java.tuple.Tuple2;
+
+import java.util.Arrays;
```
KevinGilmore: unused import

libraries/src/main/java/com/baeldung/flink/LineSplitter.java
```diff
@@ -0,0 +1,19 @@
+package com.baeldung.flink;
+
+import org.apache.flink.api.common.functions.FlatMapFunction;
+import org.apache.flink.api.java.tuple.Tuple2;
+import org.apache.flink.util.Collector;
+
+public class LineSplitter implements FlatMapFunction<String, Tuple2<String, Integer>> {
```
KevinGilmore: add @SuppressWarnings("serial") to class or create the default serial version ID

libraries/src/main/java/com/baeldung/flink/WordCount.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.List;
+
+public class WordCount {
+
+    public static DataSet<Tuple2<String, Integer>> startWordCount(ExecutionEnvironment env, List<String> lines) throws Exception {
+        DataSet<String> text = env.fromCollection(lines);
+
+        return text.flatMap(new LineSplitter())
+                .groupBy(0)
+                .aggregate(Aggregations.SUM, 1);
```
KevinGilmore: you could just use .sum(1) instead of the .aggregate(...)

libraries/src/test/java/com/baeldung/flink/WordCountTest.java
```diff
@@ -0,0 +1,120 @@
...
+        assertThat(collect.contains(new Tuple2<>("is", 2))).isTrue();
+        assertThat(collect.contains(new Tuple2<>("this", 2))).isTrue();
+        assertThat(collect.contains(new Tuple2<>("second", 1))).isTrue();
+        assertThat(collect.contains(new Tuple2<>("first", 1))).isTrue();
+    }
+
+    @Test
+    public void givenListOfAmounts_whenUseMapReduce_thenSumAmountsThatAreOnlyAboveThreshold() throws Exception {
+        //given
+        DataSource<Integer> amounts = env.fromElements(1, 29, 40, 50);
```
KevinGilmore: Why not use DataSet instead of DataSource?

libraries/src/main/java/com/baeldung/flink/WordCount.java
tomekl007: yes, i know but aggregate is more flexible because you can pass other types of operations, so i prefer to show that method, ok?

libraries/src/main/java/com/baeldung/flink/WordCount.java
tomekl007: done

libraries/src/test/java/com/baeldung/flink/WordCountTest.java
tomekl007: when you create a source, then you get DataSource, when you apply transformations on it, then you get DataSet, so this is correct

libraries/src/main/java/com/baeldung/flink/LineSplitter.java
tomekl007: done

libraries/src/main/java/com/baeldung/flink/LineSplitter.java
tomekl007: changed, to java 8 stream

libraries/pom.xml
tomekl007: formatting after the change is proper, previously was a bad indentation

KevinGilmore: I see, it's not a big deal when there is only one, but if there were a lot of changes like this, they would become "noise" when trying to review the PR :)

libraries/src/main/java/com/baeldung/flink/WordCount.java
KevinGilmore: Yes it's ok, just making an observation, not necessarily requesting a change.

libraries/pom.xml
KevinGilmore: FYI, the PR shows that there is a merge conflict with this file. Can you take a look and resolve it in your branch?

libraries/src/test/java/com/baeldung/flink/WordCountTest.java
```diff
@@ -0,0 +1,196 @@
...
+import org.junit.Test;
+
+import java.time.ZonedDateTime;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+public class WordCountTest {
```
pivovarit: Let's make it WordCountIntegrationTest

PR #1556, opened 2017-04-01
Charge a credit card with Stripe

lsoares:
Spring Boot project that includes Stripe Java library and does a charge to a credit card.

BAEL-742


stripe/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-devtools</artifactId>
```
KevinGilmore: Let's omit dev-tools from the module

stripe/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.16.16</version>
+        </dependency>
+        <dependency>
+            <groupId>com.stripe</groupId>
+            <artifactId>stripe-java</artifactId>
+            <version>4.0.0</version>
```
KevinGilmore: The latest now is 4.2.0

stripe/src/main/resources/templates/checkout.html
```diff
@@ -0,0 +1,34 @@
...
+            <input type='hidden' th:value='${amount}' name='amount' />
+            <label>Price:<span th:text='${amount/100}' /></label>
+            <!-- NOTE: data-key/data-amount/data-currency will be rendered by Thymeleaf -->
+            <script
+                src='https://checkout.stripe.com/checkout.js' 
+                class='stripe-button'
+                th:attr='data-key=${stripePublicKey},
+                         data-amount=${amount},
+                         data-currency=${currency}'
+                data-name='Baldung'
```
KevinGilmore: Baeldung (spelling)

stripe/src/main/resources/templates/checkout.html
```diff
@@ -0,0 +1,34 @@
...
+                margin: 12px;
+            }
+        </style>
+    </head>
+    <body>
+        <form action='/charge' method='POST' id='checkout-form'>
+            <input type='hidden' th:value='${amount}' name='amount' />
+            <label>Price:<span th:text='${amount/100}' /></label>
+            <!-- NOTE: data-key/data-amount/data-currency will be rendered by Thymeleaf -->
+            <script
```
KevinGilmore: Although it is supported elsewhere now, It's more typical to have a <script> tag in the <head> section. 
Either way, you do need a closing </script> tag instead of the shortcut <script ...  />

lsoares: in which section should it be at? I could not understand.

KevinGilmore: I meant to say that I've only seen <script> tags in the <head> tag (sorry, I know I typed that in the comment but must have erased it before I submitted it) -- then again, I have been out of the UI side of web dev for a few years ;) so I will defer to you as to its placement.
But I am reasonably sure that you do still need the closing </script> tag -- the shorthand form will not be interpreted correctly -- I've experienced this firsthand, and everything I've read on stackoverflow and other sites claims that this is still the case today.

lsoares: do you mean inside `head `tag?

lsoares: it's not common to have `script`s inside `form`s but look [at their example](https://stripe.com/docs/checkout/tutorial#embedding) :)

KevinGilmore: So yes, the placement within the <form> tag is fine.

lsoarestd: Yes the enclosing script I did it right away. Thanks​. Will push when I implemented your suggestion of amount.

lsoarestd:
@KevinGilmore can you check now? thx

stripe/src/main/java/com/baeldung/ChargeController.java
```diff
@@ -0,0 +1,33 @@
+package com.baeldung;
```
KevinGilmore: Let's move all classes to the package com.baeldung.stripe

stripe/src/main/java/com/baeldung/StripeService.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Service;
+
+@Service
+public class StripeService {
+
+    @Value("${STRIPE_SECRET_KEY}")
+    private String secretKey;
+
+    public void charge(ChargeRequest chargeRequest)
```
KevinGilmore: Let's return the Charge object returned by Charge.create(...)

stripe/src/main/resources/templates/result.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html xmlns='http://www.w3.org/1999/xhtml' xmlns:th='http://www.thymeleaf.org'>
+    <head>
+        <title>Result</title>
+    </head>
+    <body>
+        <div th:if='${error}' th:text='${error}' style='color: red;'></div>
+        <div th:unless='${error}' style='color: green;'>Success!</div>
```
KevinGilmore: Here's an idea: If you return the Charge object from the StripeService to the ChargeController, you could use some of the Charge response data to populate additional elements in the model for the result page. I don't know what's interesting to show, but you could pick something (maybe the charge id?) of interest.
What do you think?

lsoares: makes sense. I'll take a look at their response data.

PR #1564, opened 2017-04-02
vinod.kashyap12@gmail.com

vinodkashyap:
New author article


spring-core/.tern-project
```diff
@@ -0,0 +1,17 @@
+{
```
pivovarit: this file looks like it was commited by accident

PR #1572, opened 2017-04-02
BAEL-716 TestNG vs JUnit improvement

dhruba619:


core-java/src/test/java/com/baeldung/junit4vstestng/SortedTests.java
```diff
@@ -0,0 +1,25 @@
...
+import org.junit.Test;
+import org.junit.runners.MethodSorters;
+
+
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
+public class SortedTests {
+    
+    @Test
+    public void a_givenString_whenChangedtoInt_thenTrue(){
+        assertTrue(
```
slavisa-baeldung: the code is good, but you can keep it a single line. The manual line break is important only inside the article. You can fix it in all tests and I'll merge and suggest to publish the changes, best

dhruba619: Sure, let remove the line breaks.

PR #1593, opened 2017-04-05
Fix whenSerializingDateToISO8601_thenSerializedToText test

bozaro:
Suddenly,  `1970-01-01T02:30:00.000+0000` is not a correct ISO 8601 date: it mix basic timezone format and extended date and time format.

Correct value looks like:

 * `1970-01-01T02:30:00.000+00:00` (Extended format)
 * `19700101T023000.000+0000` (Basic format)


bozaro:
https://github.com/FasterXML/jackson-databind/issues/1020#issuecomment-292246750

PR #1595, opened 2017-04-05
Merge task for BAEL-596 web vs init into spring-mvc-java

joshiparthin:


spring-mvc-java/src/main/java/com/baeldung/spring/web/config/AppInitializer.java
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.web.context.ContextLoaderListener;
+import org.springframework.web.context.support.AnnotationConfigWebApplicationContext;
+import org.springframework.web.servlet.DispatcherServlet;
+
+import javax.servlet.ServletContext;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRegistration;
+
+// copied from - spring-mvc-web-vs-initializer 
+public class AppInitializer implements WebApplicationInitializer {
```
joshiparthin: @pivovarit The name of class needs to be matched with the one in the article: http://www.baeldung.com/spring-xml-vs-java-config . When i checked the source code linked to this article, in module spring-mvc-web-vs-initializer we have a different name. Also there is no different classes for 100% java and hybrid example as in java. I have created one class for hybrid example in spring-mvc-java and this class as it in spring-mvc-java. Either we need to rename the class for matching the article or vis-a-versa

spring-mvc-java/src/main/java/com/baeldung/spring/web/config/MyHybridAppInitializer.java
```diff
@@ -0,0 +1,26 @@
...
+import javax.servlet.ServletContext;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRegistration;
+
+import org.springframework.web.WebApplicationInitializer;
+import org.springframework.web.context.support.XmlWebApplicationContext;
+import org.springframework.web.servlet.DispatcherServlet;
+
+// copied from - spring-mvc-web-vs-initializer 
+public class MyHybridAppInitializer implements WebApplicationInitializer {
```
joshiparthin: @pivovarit  as I have mentioned in earlier comment, I have created this class here in spring-mvc-java to match the case of the article: http://www.baeldung.com/spring-xml-vs-java-config .

PR #1620, opened 2017-04-09
Code Modified as per review comments

bs-santosh:
Code Modified as per review comments by Grzegorz Piwowarek


bs-santosh:
Wrong checkin

PR #1626, opened 2017-04-10
Spring Cloud Zookeeper Updated

TehreemNisa:
- Rechecked and reformatted all code using eclipse formatting.
- Made only string response for the producer application. For 4.5 the configuration file is required so the consumer application can look for zookeeper. Please check the section 5 it shows how to see the working solution should I still drop it.
- Added the maven central links like you mentioned.
- I rechecked the names for HelloWorld and Greeting Application, didn't notice any discrepancies.Please let me know if any came in your notice.


TehreemNisa:
Resolved conflicts

PR #1685, opened 2017-04-19
Bael 736

lor6:


spring-boot/src/main/java/com/baeldung/dynamicvalidation/ContactInfoValidator.java
```diff
@@ -12,27 +13,34 @@
...
+    private String pattern;
+
     @Override
     public void initialize(final ContactInfo contactInfo) {
+        if (StringUtils.isEmptyOrWhitespace(expressionType)) {
+            LOG.error("Contact info type missing!");
+        } else {
+            pattern = expressionRepository.findOne(expressionType)
+                .map(ContactInfoExpression::getPattern)
+                .get();
```
pivovarit: And what if there is no result for a given expression type?

PR #1704, opened 2017-04-22
BAEL-758

aietcn:
1. removed existing duplicate dependencies
2. added dependencies and code for #BAEL-758


libraries/pom.xml
```diff
@@ -39,12 +39,31 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>${maven-surefire-plugin.version}</version>
                 <configuration>
+                    <testFailureIgnore>true</testFailureIgnore>
+                    <includes>
+                        <include>**/*Test.java</include>
```
pivovarit: Why do we need to add this explicitly?

libraries/pom.xml
```diff
@@ -39,12 +39,31 @@
...
+                        <include>**/GithubUserProfilePayload.java</include>
```
pivovarit: Is this a test class? If yes, the name should end with "Test"

libraries/src/main/java/com/baeldung/serenity/github/RetrieveUtil.java
```diff
@@ -0,0 +1,29 @@
...
+import org.apache.http.impl.client.HttpClientBuilder;
+import org.apache.http.util.EntityUtils;
+
+import java.io.IOException;
+
+public class RetrieveUtil {
+
+    public static <T> T retrieveResourceFromResponse(final HttpResponse response, final Class<T> clazz) throws IOException {
+        final String jsonFromResponse = EntityUtils.toString(response.getEntity());
+        final ObjectMapper mapper = new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
pivovarit: Let's make the ObjectMapper a private static final field. There is no need for creating it each time we call this methid

libraries/pom.xml
aietcn: `GithubUserProfilePayload.java` is indeed a test case, but the Test file name should be in accord with the story name: `github_user_profile_payload.story`. I think it's forced by Serenity.
I did a workaround: renamed the story file name to `github_user_profile_payload_live_test.story`, so that the test can detect the relevant story file.

libraries/pom.xml
aietcn: In Serenity examples, I intentionally wrote a must-fail test to demonstrate how it can be reflected in the Serenity report. Meanwhile, we don't want the automated tests to stop because of this 'must-fail' test. That's why `testFaillureIgnore: true` was added.

I'll mark the test as `@ignored` as I have explained how to reflect the failures in the article.

aietcn:
Changes done! @pivovarit

PR #1759, opened 2017-04-30
adds serialization code

eprescribe:


pivovarit:
@eprescribe I assume this one is not valid anymore, let me know if this is not the case

PR #1760, opened 2017-04-30
BAEL-786 Refactored code after feedback

dhruba619:


dhruba619:
Will open fresh with only relevant changes

PR #1784, opened 2017-05-04
Add unit test for File path article - BAEL-827

vivekkr12:
Add a unit test class to demonstrate the results of the three different methods to get filesystem path.

### DIFF
--- a/core-java/src/test/java/org/baeldung/java/io/JavaFilePathUnitTest.java
+++ b/core-java/src/test/java/org/baeldung/java/io/JavaFilePathUnitTest.java
@@ -1,0 +1,97 @@
+package org.baeldung.java.io;
+
+import static org.junit.Assert.*;
+
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.io.File;
+import java.io.IOException;
+
+public class JavaFilePathUnitTest {
+
+    private static String userDir;
+
+    @BeforeClass
+    public static void createFilesAndFolders() throws IOException {
+        userDir = System.getProperty("user.dir");
+
+        new File(userDir + "/baeldung/foo").mkdirs();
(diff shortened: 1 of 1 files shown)


core-java/src/test/java/org/baeldung/java/io/JavaFilePathUnitTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Test
+    public void whenCanonicalPathWithDotShorthandResolved_thenSuccess() throws IOException {
+        File file = new File("baeldung/bar/baz/./baz-one.txt");
+        String expectedPath = isWindows() ? userDir + "\\baeldung\\bar\\baz\\baz-one.txt" : userDir + "/baeldung/bar/baz/baz-one.txt";
+        String actualPath = file.getCanonicalPath();
+        assertEquals(expectedPath, actualPath);
+    }
+
+    @Test(expected = IOException.class)
```
KevinGilmore: This test fails on Unix/Linux. If you use the JUnit Rules, you can check to see whether it is a Windows system and, if so, tell it to expect IOException. See https://github.com/junit-team/junit4/wiki/Rules for more info.

core-java/src/test/java/org/baeldung/java/io/JavaFilePathUnitTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.io.File;
+import java.io.IOException;
+
+public class JavaFilePathUnitTest {
+
+    private static String userDir;
+
+    @BeforeClass
+    public static void createFilesAndFolders() throws IOException {
+        userDir = System.getProperty("user.dir");
```
KevinGilmore: Instead of the user's home directory, make use of the TemporaryFolder construct provided in JUnit Rules.

PR #1807, opened 2017-05-08
Spring YAML Configuration

TehreemNisa:


TehreemNisa:
Added properties file

PR #1839, opened 2017-05-12
BAEL-318: updated activation page

sybraimah:
Updated the activation.jsp page to reflect the tutorial


pivovarit:
@sybraimah which article is this related to? It does not look like BAEL-318

PR #1863, opened 2017-05-16
BAEL-871: What is serialVersionUID?

fromalexwithluv:


pivovarit:
I assume this one is not valid anymore. Let me know if this is not the case

PR #1881, opened 2017-05-19
Hibernate5 multitenancy changing to hibernate.properties  config file.

joshiparthin:


hibernate5/src/main/java/com/baeldung/hibernate/HibernateMultiTenantUtil.java
```diff
@@ -18,78 +22,90 @@
...
-        } else if (tenantStrategy.toUpperCase().equals("SCHEMA")) {
-            return originalUrl + String.format(SCHEMA_TOKEN, tenant);
-        } else {
-            throw new UnsupportedTenancyException("Not yet supported");
-        }
-    }
-
-    public static final String SCHEMA_TOKEN = ";INIT=CREATE SCHEMA IF NOT EXISTS %1$s\\;SET SCHEMA %1$s";
-    public static final String DEFAULT_DB_NAME = "mydb1";
+	private static SessionFactory sessionFactory;
```
KevinGilmore: I can't tell what changed, because you changed the indents from spaces to tabs. Can you put them back to spaces using the standard formatter?

hibernate5/src/main/java/com/baeldung/hibernate/MultiTenantHibernateTest.java
```diff
@@ -0,0 +1,74 @@
+package com.baeldung.hibernate;
```
KevinGilmore: We had removed this file in an earlier commit. It must have slipped back in when you switched machines.

hibernate5/src/test/java/com/baeldung/hibernate/MultiTenantHibernateIntegrationTest.java
```diff
@@ -7,6 +7,8 @@
```
KevinGilmore: Couple of things here:
1) it does not compile -- looks like the junit dependency was removed from the pom at some point, as was the build configuration.
2) in a test, you typically do not catch Exceptions; instead, declare the thrown exceptions on the test method

PR #1888, opened 2017-05-21
Spring bean injection

Iucid:
Different Types of Bean Injection in Spring (http://inprogress.baeldung.com/wp-admin/post.php?post=42545&action=edit)


pivovarit:
@Iucid we are not merging this one, this was only for the evaluation purposes

PR #1948, opened 2017-05-29
spring-5-mvc project

asif-anwar:
spring-5-mvc


pivovarit:
@asif-anwar is this change associated with any particular task?

asif-anwar:
@pivovarit 

It's for following task,
http://jira.baeldung.com/browse/COURSE-54

PR #1950, opened 2017-05-29
Added DAO for Supplier

joshiparthin:


hibernate5/src/test/java/com/baeldung/hibernate/MultiTenantDaoHibernateIntegrationTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+import static org.junit.Assert.assertNull;;
+
+public class MultiTenantDaoHibernateIntegrationTest {
+    @Test
+    public void givenDBMode_whenFetchingSuppliersByName_thenChecking() throws UnsupportedTenancyException, IOException {
+        SessionFactory sessionFactory = HibernateMultiTenantUtil.getSessionFactory();
+        
+        SupplierDao myDb1Dao = new SupplierDao(sessionFactory, "mydb1");
+        Supplier db1SupplierName = myDb1Dao.findByName("John");
```
KevinGilmore: Test fails. Both this findByName and the second return null, because you did not insert the data into the database as you did in the other test class. Recommend creating a variable for the name "John", using the variable in both calls to findByName, and doing an assertNotNull on the first one.

PR #1957, opened 2017-05-31
ADDED:unit test cases with DAO,service,controller and PASSED.

yangboz:
ADDED:unit test cases with DAO,service,controller and PASSED.


pivovarit:
We are not merging the evaluation code

PR #1958, opened 2017-05-31
Demo app: Types of Dependency Injection in Spring (elyor.m@gmail.com)

elyor85:


pivovarit:
We are not merging the evaluation code

PR #1963, opened 2017-05-31
Difference between URL and URI #864.

hariprasad108:
Difference between URL and URI, pom file.


pivovarit:
@hariprasad108 you can just keep commiting to one branch. The PR will pick those changes up automatically

PR #1987, opened 2017-06-03
Difference between url and uri bael 864 hariprasad

hariprasad108:


hariprasad108:
Difference URI URL REST project has been moved from spring-starter module to spring-rest.

PR #2033, opened 2017-06-10
dhruba619-master

slavisa-baeldung:
BAEL-776 - porting the vertx-spring code to vertx module


pivovarit:
@slavisa-baeldung is this one still valid?

PR #2045, opened 2017-06-11
BAEL-791 Liquibase rollback

amilabanuka:
added the sample liquibase rollback scripts


amilabanuka:
closing since the build is failing due to the db connection

PR #2069, opened 2017-06-13
Introduction to Ascii doctor

ante003:


ante003:
@pivovarit , can you check why I can't merge. It doesn't show any error but only message that it cannot merge automaticly.

PR #2099, opened 2017-06-16
Self-Healing services with Hystrix

ahamedm:
Kindly verify the commit 091ea10


pivovarit:
@ahamedm I assume this one is the old one- let me know if this is not the case

PR #2108, opened 2017-06-18
Difference between url and uri bael 864 hariprasad net

hariprasad108:
Java-Networking deleted to check result of build.


hariprasad108:
Bael-864 Java Core Networking has been finished.

PR #2114, opened 2017-06-20
BAEL-509

Thoughtscript:


Thoughtscript:
@KevinGilmore: might be my confusion but I thought that was the last (stablelish) official release from Apache? The War Runner Tomcat 8 is here: https://mvnrepository.com/artifact/org.apache.tomcat.maven/tomcat8-war-runner  and is also at the same release spec. I've gone ahead and used the Tomcat 7 version which has a stable release at 2.2 since Tomcat 8 supports Tomcat 7 wars.

PR #2150, opened 2017-06-25
BAEL-861 Introduction to Awaitlity

retroq:
http://jira.baeldung.com/browse/BAEL-861


libraries/src/test/java/com/baeldung/awaitility/AsyncServiceTest.java
```diff
@@ -0,0 +1,84 @@
...
+public class AsyncServiceTest {
+    private AsyncService asyncService;
+
+    @Before
+    public void setUp() {
+        asyncService = new AsyncService();
+    }
+
+    @Test
+    public void givenAsyncService_whenInitialize_thenInitOccurs1() {
```
KevinGilmore: Coming up with good names for the tests is a little tricky for this, because if you look at it one way, we are testing the async service, but since the article is about testing, you could also say we're testing the Awaitility library itself. Either way, the "givenAsyncService" doesn't seem quite right.

I'm 50-50 on changing them. Let's leave them as-is for now, and if asked to rename them during a higher-level review, we'll address it then.

PR #2153, opened 2017-06-25
BAEL-947 - moving from @GetMapping to @RequestMapping on account of https://github.com/spring-cloud/spring-cloud-netflix/issues/760

slavisa-baeldung:
BAEL-947 - moving from @GetMapping to @RequestMapping on account of https://github.com/spring-cloud/spring-cloud-netflix/issues/760


bandanag1:
Still getting error for @GetMapping in spring-cloud. Error: FactoryBean threw exception on object creation; nested exception is java.lang.IllegalStateException: Method get not annotated with HTTP method type (ex. GET, POST)
Any idea any one? Can we still use it with feign client in spring boot 2?

eugenp:
Hey @bandanag1 quick question - is that on our codebase here?

PR #2170, opened 2017-06-27
PR for Introduction to Apache Commons Collections MapUtils

rockoder:


apache-commons-collections/src/test/java/org/baeldung/java/MapUtilsTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+import org.apache.commons.collections4.MapUtils;
+import org.apache.commons.collections4.PredicateUtils;
+import org.apache.commons.collections4.TransformerUtils;
+import org.junit.Test;
+
+public class MapUtilsTest {
+
+    @Test
+    public void whenCreateMapFrom2DArray_THEN_MUST_CREATE_Map() {
```
slavisa-baeldung: Tipically, we're using the style that in this case would match whenCreateMapFrom2DArray_theMapIsCreated etc. So even the than part is better when camelCased

apache-commons-collections/src/test/java/org/baeldung/java/MapUtilsTest.java
```diff
@@ -0,0 +1,87 @@
...
+        Map colorMap = MapUtils.putAll(new HashMap(), new String[][] {
+            {"RED", "#FF0000"},
+            {"GREEN", "#00FF00"},
+            {"BLUE", "#0000FF"}
+        });
+
+        assertNotNull(colorMap);
```
slavisa-baeldung: we should do more complete assertions, likely using hamcrest map matchers, as it is nice and readable

PR #2176, opened 2017-06-29
How to zip two collections in Java? 

daveRanjan:


core-java/src/main/java/com/baeldung/ZipCollectionExample.java
```diff
@@ -0,0 +1,41 @@
...
+import com.google.common.collect.Streams;
+import org.jooq.lambda.Seq;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.IntStream;
+
+public class ZipCollectionExample {
+    static List<String> names = new ArrayList<>(Arrays.asList("John", "Jane", "Jack", "Dennis"));
```
pivovarit: Is there any reason why we can't simply use Arrays.asList without passing that to ArrayList?

core-java/src/main/java/com/baeldung/ZipCollectionExample.java
```diff
@@ -0,0 +1,41 @@
...
+
+    static List<Integer> ages = new ArrayList<>(Arrays.asList(24, 25, 27));
```
pivovarit: Same here

core-java/src/main/java/com/baeldung/ZipCollectionExample.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public static void main(String[] args) {
+        //      Using Streams API from Guava 21
+        Streams
```
pivovarit: Separate samples should be organized in the form of short test cases. When it's done like this, readers can run all of them separately easily

PR #2233, opened 2017-07-08
junit 5 extensions

lor6:


junit5/pom.xml
```diff
@@ -10,25 +10,27 @@
     <name>junit5</name>
     <description>Intro to JUnit 5</description>
 
-    <parent>
```
pivovarit: Why did you remove the parent?

PR #2264, opened 2017-07-15
#BAEL-1007

aietcn:
improvement to #BAEL-397


aietcn:
@pivovarit the jackson-dataformat-smile dependency is needed by servo-atlas

PR #2286, opened 2017-07-19
BAEL-771

Thoughtscript:
Neuroph Example


Thoughtscript:
"The command "travis_wait 60 mvn -q test -fae" failed and exited with 1 during ."

PR #2293, opened 2017-07-19
Examples for BAEL-422

NareshBabuPB:
Collection API in Vavr

### DIFF
--- a/vavr/src/test/java/com/baeldung/vavr/collections/CollectionAPIUnitTest.java
+++ b/vavr/src/test/java/com/baeldung/vavr/collections/CollectionAPIUnitTest.java
@@ -1,0 +1,261 @@
+package com.baeldung.vavr.collections;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Comparator;
+
+import org.junit.Test;
+
+import io.vavr.Tuple;
+import io.vavr.Tuple2;
+import io.vavr.collection.Array;
+import io.vavr.collection.CharSeq;
+import io.vavr.collection.HashSet;
+import io.vavr.collection.Iterator;
+import io.vavr.collection.List;
+import io.vavr.collection.Map;
+import io.vavr.collection.Queue;
+import io.vavr.collection.Set;
+import io.vavr.collection.SortedMap;
+import io.vavr.collection.SortedSet;
+import io.vavr.collection.Stream;
+import io.vavr.collection.TreeMap;
+import io.vavr.collection.TreeSet;
+import io.vavr.collection.Vector;
+
+public class CollectionAPIUnitTest {
+
+    @Test
+    public void givenEmptyList_whenStacked_thenCorrect() {
+        List<Integer> intList = List.empty();
+
+        List<Integer> anotherList = intList.push(4)
+            .push(0);
+        Iterator<Integer> iterator = anotherList.iterator();
+
+        assertEquals(new Integer(0), iterator.next());
+        assertEquals(new Integer(4), iterator.next());
+    }
+
+    @Test
+    public void givenList_whenPrependTail_thenCorrect() {
+        List<Integer> intList = List.of(1, 2, 3);
+
+        List<Integer> newList = intList.tail()
+            .prepend(0);
+
+        assertEquals(new Integer(1), intList.get(0));
+        assertEquals(new Integer(2), intList.get(1));
+        assertEquals(new Integer(3), intList.get(2));
+
+        assertNotSame(intList.get(0), newList.get(0));
+        assertEquals(new Integer(0), newList.get(0));
+        assertSame(intList.get(1), newList.get(1));
+        assertSame(intList.get(2), newList.get(2));
+    }
+
+    @Test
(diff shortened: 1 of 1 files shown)


vavr/src/test/java/com/baeldung/vavr/collections/CollectionAPIUnitTest.java
```diff
@@ -0,0 +1,261 @@
...
+    @Test
+    public void givenQueue_whenEnqueued_thenCorrect() {
+        Queue<Integer> queue = Queue.of(1, 2, 3);
+        Queue<Integer> secondQueue = queue.enqueue(4)
+            .enqueue(5);
+
+        assertEquals(3, queue.size());
+        assertEquals(5, secondQueue.size());
+
+        secondQueue.dequeue()
```
pivovarit: Not sure about this structure. Isn't there an easier way of checking if all queue elements are in there? Queue is essentially an iterable so it should be doable easily

NareshBabuPB: Yes, Queue is iterable. But, I wanted to show that dequeue() function returns a Tuple2 which is different from a mutable Queue. And that there is a map() function to work on the contents of the Tuple2.

pivovarit: This is definitely a must-have to show those things but not this way by nesting assertions in _map_ calls and not by nesting _dequeue_ calls inside _map_ calls.

If you want to show that a Tuple gets returned containing _head_ and _tail_, you can start easily with:
`Tuple2<Integer, Queue<Integer>> result = secondQueue.dequeue();`
 and then perform necessary assertions. 

Next thing, if you want to show a _map_ call, definitely do not go with names like _(k,v)_ which suggest that this key-value relationship. Keep in mind that _map_ always needs to return another _Tuple_ 

`result.map((head, tail) -> Tuple.of(head.toString(), tail))`

There is also a method that you can use for transforming the whole tuple into something that is not wrapped in the Tuple:
`result.apply((head, tail) -> head.toString());`

and so on... you can store results of those operations as variables and perform assertions on them.

vavr/src/test/java/com/baeldung/vavr/collections/CollectionAPIUnitTest.java
```diff
@@ -0,0 +1,261 @@
...
+        javaList.add(4);
+    }
+
+    @Test
+    public void givenList_whenSquared_thenCorrect() {
+        List<Integer> vavrList = List.of(1, 2, 3);
+        Number sum = vavrList.map(i -> i * i)
+            .sum();
+
+        assertEquals(new Long(14), sum);
```
pivovarit: The boxing of Long is unnecessary

vavr/src/test/java/com/baeldung/vavr/collections/CollectionAPIUnitTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+        Iterator<String> iterator = set.iterator();
+        assertEquals("Blue", iterator.next());
+        assertEquals("Green", iterator.next());
+        assertEquals("Red", iterator.next());
+    }
+
+    @Test
+    public void givenSortedSet_whenReversed_thenCorrect() {
+        Comparator<String> reverseCompare = (a, b) -> b.compareTo(a);
```
pivovarit: the `(a, b) -> b.compareTo(a)` can be replaced with `Comparator.reverseOrder();`

vavr/src/test/java/com/baeldung/vavr/collections/CollectionAPIUnitTest.java
```diff
@@ -0,0 +1,261 @@
...
+        assertEquals(new Integer(3), iterator.next()
+            ._1());
+    }
+
+    @Test
+    public void givenJavaList_whenConverted_thenCorrect() {
+        java.util.List<Integer> javaList = java.util.Arrays.asList(1, 2, 3, 4);
+        List<Integer> vavrList = List.ofAll(javaList);
+
+        assertTrue(vavrList instanceof io.vavr.collection.List);
```
pivovarit: Those _instanceof_ assertions do not really bring anything new here. You can easily remove them and leave this test without assertions. After all, those are just supposed to be a short `main` methods and not full-blown tests

vavr/src/test/java/com/baeldung/vavr/collections/CollectionAPIUnitTest.java
```diff
@@ -0,0 +1,261 @@
...
+        Iterator<String> iterator = set.iterator();
+        assertEquals("Red", iterator.next());
+        assertEquals("Green", iterator.next());
+        assertEquals("Blue", iterator.next());
+    }
+
+    @Test
+    public void givenMap_whenIterated_thenCorrect() {
+        Map<Integer, List<Integer>> map = List.rangeClosed(0, 10)
+            .groupBy(i -> i % 2);
```
pivovarit: Please use the provided formatter. Continuation indents should be 2 spaces

vavr/src/test/java/com/baeldung/vavr/collections/CollectionAPIUnitTest.java
NareshBabuPB: Is it Ok to have this example in a private method in the same test class? Just wondering if a separate main class is required just for a few lines of code.

For rest of the comments, will work on it as suggested.

PR #2367, opened 2017-08-03
BAEL-1080 Introduction to Future and Pattern Matching in Vavr

hunghoang:
I've closed the last pull request and open another one by removing all Pattern Matching Unit test and add  some unit test of Future with Pattern Matching


spring-all/src/main/java/org/baeldung/bean/injection/Author.java
```diff
@@ -0,0 +1,5 @@
+package org.baeldung.bean.injection;
```
pivovarit: Please remove files that are not related to the article

vavr/src/test/java/com/baeldung/vavr/future/FutureUnitTest.java
```diff
@@ -0,0 +1,212 @@
...
+            .intValue());
+    }
+
+    @Test
+    public void givenAFuture_WhenCallCancel_ShouldStopTheActionOfFuture() {
+        long waitTime = 1000;
+        int[] store = new int[1];
+        Future<Integer> future = Future.of(() -> {
+            Thread.sleep(waitTime);
+            store[0] = 1;
```
pivovarit: It should be possible to obtain the result through Future's API and not by doing this hack

hunghoang:
create another one

PR #2384, opened 2017-08-07
Why so many files?

nbosecker:
I'm screwed up on Git, like always, but the top 2 files (the pom.xml and the unit test for the code) are the only 2 relevant changes here. I'm not sure how to fix this problem with the PR.


pivovarit:
@nbosecker you need to keep updating your own fork with changes from the upstream, if you do not do this your changes keep conflicting with what is already there and become a merging hell

pivovarit:
I assume that one is no longer valid

PR #2388, opened 2017-08-07
BAEL-1044 - Introduction to NoException

egoebelbecker:
Classes for NoException tutorial. Test classes did not seem warranted since the code is trivial. 

Since NoException uses Logger for exceptions, examples use stdout to differentiate between NoException's output and code path when an exception is not thrown.


egoebelbecker:
Canceling so I can submit a newer request.

PR #2411, opened 2017-08-10
Bael 817

dimitarsazdovski:


libraries/src/test/java/com/baeldung/geotools/BuiltFeaturesTest.java
```diff
@@ -0,0 +1,27 @@
+package com.baeldung.geotools;
+
+import static org.junit.Assert.assertNotNull;
+
+import org.geotools.feature.DefaultFeatureCollection;
+import org.geotools.feature.simple.SimpleFeatureBuilder;
+import org.junit.Test;
+import org.opengis.feature.simple.SimpleFeatureType;
+
+public class BuiltFeaturesTest {
```
KevinGilmore: Let's rename this class to GeoToolsUnitTest (the "UnitTest" part is a naming convention).

PR #2428, opened 2017-08-12
BAEL-1085

Thoughtscript:


spring-rest/src/main/java/net/baeldung/ApplicationMain.java
```diff
@@ -0,0 +1,20 @@
+package net.baeldung;
```
pivovarit: com.baeldung :)

spring-rest/src/main/java/net/baeldung/ApplicationMain.java
```diff
@@ -0,0 +1,20 @@
...
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.builder.SpringApplicationBuilder;
+import org.springframework.boot.web.support.SpringBootServletInitializer;
+
+@SpringBootApplication
+public class ApplicationMain extends SpringBootServletInitializer {
```
pivovarit: Do we need SpringBootServletInitializer? I believe that should work only with the @SpringBootApplication

spring-rest/src/main/java/net/baeldung/controllers/ExamplePostController.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.web.bind.annotation.RequestBody;
+import net.baeldung.transfer.LoginForm;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+@Controller
+@RequestMapping("/post")
+public class ExamplePostController {
+
+    Logger log = LoggerFactory.getLogger(ExamplePostController.class);
```
pivovarit: private static final log

spring-rest/src/main/java/net/baeldung/services/ExampleService.java
```diff
@@ -0,0 +1,12 @@
+package net.baeldung.services;
+
+import net.baeldung.transfer.LoginForm;
+import org.springframework.stereotype.Service;
+
+@Service
+public class ExampleService {
+
+        public boolean fakeAuthenticate(LoginForm lf) {
+                return (lf.getUsername().equals(lf.getUsername()) && lf.getPassword() == lf.getPassword());
```
pivovarit: Wouldn't it be easier to simply return true?

spring-rest/src/main/java/net/baeldung/transfer/LoginForm.java
```diff
@@ -0,0 +1,24 @@
+package net.baeldung.transfer;
+
+public class LoginForm {
+    private String username;
+    private String password;
+
+    public LoginForm() {};
```
pivovarit: That's redundant

spring-security-mvc-socket/src/main/java/com/baeldung/springsecuredsockets/config/SecurityConfig.java
```diff
@@ -89,7 +89,7 @@ protected void configure(HttpSecurity http) throws Exception {
                 .antMatchers("/", "/index", "/authenticate")
                 .permitAll()
                 .antMatchers("/secured/**/**",
-                        "/secured/success", "/secured/socket", "/secured/success")
+                        "/secured/success", "/secured/socket")
```
pivovarit: Is that related?

spring-rest/src/test/java/net/baeldung/controllers/ExamplePostControllerServiceUnitTest.java
```diff
@@ -0,0 +1,60 @@
...
+                    .content(jsonBody)
+                    .contentType("application/json"))
+                    .andDo(print())
+                    .andExpect(status().isOk());
+        } catch (Exception e) {
+            System.out.println("Exception: " + e);
+        }
+    }
+
+    @After
```
pivovarit: In @Before you are assigning something to mockMvc - so this @After does not make any impact and can be safely removed

spring-rest/src/test/java/net/baeldung/controllers/ExamplePostControllerIntegrationTest.java
```diff
@@ -0,0 +1,54 @@
...
+                    .andDo(print())
+                    .andExpect(status().isOk())
+                    .andExpect(content().contentType(MediaType.APPLICATION_JSON_UTF8))
+                    .andExpect(jsonPath("$.text").value("Thanks For Posting!!!"));
+        } catch (Exception e) {
+            System.out.println("Exception: " + e);
+        }
+    }
+
+    @After
```
pivovarit: Same here, you can remove that

spring-rest/src/main/java/net/baeldung/ApplicationMain.java
Thoughtscript: Normally, the convention is com, org, then net (used rarely).  I thought it easier to simply and cleanly add my code into a separate grouping. However, I think I'm supposed to move this over to Spring Angular instead. So... my point is moot.

spring-rest/src/main/java/net/baeldung/controllers/ExamplePostController.java
Thoughtscript: Will make that change.

spring-rest/src/main/java/net/baeldung/services/ExampleService.java
Thoughtscript: Yeah - it could be simplified - just wanted to flesh that out - here it does qualify as a kind of authentication check since it compares values.

spring-security-mvc-socket/src/main/java/com/baeldung/springsecuredsockets/config/SecurityConfig.java
Thoughtscript: No but this is a minor change requested elsewhere.

spring-rest/src/test/java/net/baeldung/controllers/ExamplePostControllerServiceUnitTest.java
Thoughtscript: True but leaving the full before, test, and after trio allows someone learning to see how a good test should be set up.

Thoughtscript:
This code is being moved over to Spring-Rest-Angular per document review! Most of the points above are moot due to that change but some will be incorporated for sure. Thanks!

Thoughtscript:
Had removed a duplicate ant matcher from BAEL-509 - this was the incorrect process - reverting now, article has been corrected, but will issue a separate PR for this

PR #2461, opened 2017-08-18
BAEL-812: moving examples to rule-engines folder

felipeazv:
BAEL-812: List of Rules Engines for Java
- removed evaluation article files (on spring-core/src/main/java/com/baeldung/autowired)
- move examples folders (easy-rules, rulebook and openl-tablets) from /tutorials to /tutorials/rule-engines


rule-engines/easy-rules/pom.xml
```diff
@@ -0,0 +1,15 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<groupId>com.baeldung.easyrules</groupId>
+	<artifactId>easy-rules</artifactId>
+	<version>1.0.0-SNAPSHOT</version>
+
```
pivovarit: Let's add the parent here, you can look it up in other modules

felipeazv: Sure!

PR #2477, opened 2017-08-21
BAEL-1028 - Spring Boot GraphQL Example

sazzer:


sazzer:
I'm not 100% sure I've put this in the correct module. Especially since I needed to work out what the authentication details were to even access the endpoints that this module exposes.

PR #2478, opened 2017-08-21
BAEL-1080 Introduction to Future in Vavr

hunghoang:


vavr/src/test/java/com/baeldung/vavr/future/FutureUnitTest.java
```diff
@@ -0,0 +1,209 @@
...
+    @Test
+    public void givenAFuture_WhenCallCancel_ShouldReturnCancellationException() {
+        long waitTime = 1000;
+        Future<Integer> future = Future.of(() -> {
+            Thread.sleep(waitTime);
+            return 1;
+        });
+        future.cancel();
+        future.await();
+        assertEquals(CancellationException.class, future.getCause().get().getClass());
```
pivovarit: Does not Future simply rethrow the exception when you call get()?

vavr/src/test/java/com/baeldung/vavr/future/FutureUnitTest.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    @Test
+    public void givenAFuture_WhenAddAndThenConsumer_ShouldCallConsumerWithResultOfFutureAction() {
+        int[] store1 = new int[1];
+        int[] store2 = new int[1];
+        Future<Integer> future = Future.of(() -> 1);
+        future.andThen(i -> store1[0] = i.get() + 1)
+            .andThen(i -> store2[0] = store1[0] + 1);
+        while (store2[0] == 0) {
```
pivovarit: Use awaitility library for dealing with async tests

vavr/src/test/java/com/baeldung/vavr/future/FutureUnitTest.java
hunghoang: Yes, it does

vavr/src/test/java/com/baeldung/vavr/future/FutureUnitTest.java
hunghoang: with these tests, I cannot avoid using while, because I want to check the side effects of andThen, onSuccess... function. These functions executed after future completed. Cannot use future.await(), that return right away when future completed

hunghoang: Let me try another way to express these test

PR #2496, opened 2017-08-24
user: hugosama1@gmail.com

hugosama1:
Evaluation Post


hugosama1:
- Changed from XML to Annotation based injection
- Changed test to simple unit test following BDD convention
- Added AppConfig class to handle injection

PR #2497, opened 2017-08-25
BAEL-1075

shekhar-s:
Adding test cases for "how to delete a directory"


core-java/pom.xml
```diff
@@ -162,6 +162,12 @@
             <version>${avaitility.version}</version>
             <scope>test</scope>
         </dependency>
+		<dependency>
+		    <groupId>org.springframework</groupId>
```
pivovarit: Why do we need spring in the core-java module?

zhendrikse: It is one of the libraries that is used to show how to delete directories recursively in Java, and that article and code is part of java-core....

PR #2583, opened 2017-09-08
BAEL-1085: @RequestBody and @ResponseBody Annotation

Thoughtscript:


spring-rest/src/main/java/net/baeldung/transfer/ResponseTransfer.java
```diff
@@ -1,4 +1,4 @@
-package com.baeldung.transfer;
+package net.baeldung.transfer;
```
pivovarit: That should be "com.baeldung" :) here, and everywhere else - rest looks fine

PR #2585, opened 2017-09-08
BAEL-1041 Introduction to Caffeine

retroq:


libraries/src/test/java/com/baeldung/caffeine/CaffeineUnitTest.java
```diff
@@ -0,0 +1,156 @@
...
+import com.github.benmanes.caffeine.cache.AsyncLoadingCache;
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.Expiry;
+import com.github.benmanes.caffeine.cache.LoadingCache;
+
+public class CaffeineUnitTest {
+
+    @Test
+    public void given_Cache_whenPopulate_thenValueStored() {
```
KevinGilmore: Several of the test names begin with "given_".
Let's remove the trailing underscore. So this one, for example, would just be:
"givenCache_whenPopulate_thenValueStored"

retroq: Done

libraries/src/test/java/com/baeldung/caffeine/CaffeineUnitTest.java
```diff
@@ -0,0 +1,158 @@
...
+    @Test
+    public void givenCache_whenPopulate_thenValueStored() {
+        Cache<String, DataObject> cache = Caffeine.newBuilder()
+          .expireAfterWrite(1, TimeUnit.MINUTES)
+          .maximumSize(100)
+          .build();
+
+        String key = "A";
+        DataObject dataObject = cache.getIfPresent(key);
+        assertNull(dataObject);
```
KevinGilmore: Always add a blank line between what is being tested and the assertion(s).

libraries/src/main/java/com/baeldung/caffeine/DataObject.java
```diff
@@ -0,0 +1,28 @@
...
+    @Override
+    public String toString() {
+        return "DataObject{" +
+                "data='" + data + '\'' +
+                '}';
+    }
+
+    public static DataObject get(String data) {
+        objectCounter++;
+        System.out.println(String.format("Initializing DataObject#%d with data '%s'", objectCounter, data));
```
pivovarit: That looks like an overkill, let's use some logger here that can interpolate strings

PR #2587, opened 2017-09-08
BAEL-1069: Guide to diamond operator in Java

Sgitario:


core-java/src/test/java/org/baeldung/java/diamond/DiamondOperatorTest.java
```diff
@@ -0,0 +1,13 @@
+package org.baeldung.java.diamond;
+
+import static org.junit.Assert.assertNotNull;
+
+import org.junit.Test;
+
+public class DiamondOperatorTest {
```
KevinGilmore: rename to DiamondOperatorUnitTest

core-java/src/test/java/org/baeldung/java/diamond/DiamondOperatorTest.java
```diff
@@ -0,0 +1,13 @@
...
+    @Test
+    public void shouldCreateCarUsingDiamondWithoutWarnings() {
```
KevinGilmore: Not a very interesting test...but let's at least follow the naming convention for tests: givenX_whenY_thenZ (the givenX is optional). So you would have something like: whenCreateCarUsingDiamondOperator_thenSuccess

core-java/src/test/java/org/baeldung/java/diamond/DiamondOperatorTest.java
Sgitario: Done

core-java/src/test/java/org/baeldung/java/diamond/DiamondOperatorTest.java
Sgitario: Renamed. I know it's not a very interesting test, but taking into account the article...

KevinGilmore: That's fair.

PR #2645, opened 2017-09-19
BAEL-1098

sybraimah:
Testing a REST API with Karate


spring-mvc-webflow/src/main/webapp/WEB-INF/view/activation.jsp
```diff
@@ -1,22 +1,10 @@
 <html>
```
KevinGilmore: Why are there changes to this file here?

sybraimah:
changes have been pulled from upstream

PR #2658, opened 2017-09-22
From JDO to XML and Back Again article PR

shaimaa-hshalaby:


pivovarit:
@shaimaa-hshalaby this one is not valid anymore, right? let me know if this is not the case

PR #2659, opened 2017-09-22
BAEL-1155 crdt

tomekl007:


libraries/pom.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
KevinGilmore: Please do not alter spacing within or between existing lines. This makes the review take longer than it should, as I have to wade through what is an actual change vs what is only a whitespace change.

tomekl007: it was wrong formatted previously. Someone commited wrong style so I want to fix that issue :)

KevinGilmore: Understood. I've had difficulties with this file recently since the formatting was changed to the wrong style. Several PRs had merge conflicts in this file initially, as is the case here. You'll need to pull down the upstream changes and resolve the merge conflicts and push again :)

tomekl007: done

PR #2664, opened 2017-09-24
ASM Sample Project

k0l0ssus:


asm/pom.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
KevinGilmore: Since asm is a new module, you'll need to add it to the root-level pom.xml, so that it gets built during the main build

k0l0ssus: Done

PR #2675, opened 2017-09-26
Atomix Intro Project

abirkhan04:


atomix/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+		</dependency>
+	</dependencies>
+	<build>
+		<sourceDirectory>src</sourceDirectory>
+		<plugins>
+			<plugin>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<version>3.1</version>
+				<configuration>
+					<source>1.3</source>
```
pivovarit: @abirkhan04 why so low version numbers?

atomix/src/test/java/com/atomix/exampletest/ClientExampleTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.Test;
+
+public class ClientExampleTest {
+
+    @Test
+    public void ExampleTest() throws InterruptedException {
```
pivovarit: Let's use the naming convention specified in the documentation

atomix/src/test/java/com/atomix/exampletest/ClientExampleTest.java
```diff
@@ -0,0 +1,38 @@
...
+        AtomixClient client = AtomixClient.builder()
+            .withTransport(new NettyTransport())
+            .build();
+
+        List<Address> cluster = Arrays.asList(new Address("localhost", 8700), new Address("localhsot", 8701));
+        client.connect(cluster)
+            .thenRun(() -> System.out.println("Client Connected"));
+
+        Thread.sleep(5000);
```
pivovarit: Can't we replace this with Awaitility or any other form of testing things that happen asynchronously?

atomix/pom.xml
abirkhan04: I am updating version number.

atomix/src/test/java/com/atomix/exampletest/ClientExampleTest.java
abirkhan04: Need to check-back the documentation what I am missing here.

atomix/src/test/java/com/atomix/exampletest/ClientExampleTest.java
abirkhan04: I did forced synchronization here so that I could make sure client is connected to the cluster and I can retrieve the map value from there in the following line.

PR #2715, opened 2017-10-08
BAEL-719: Intro to Spring Cloud Stream

Sgitario:


spring-cloud/spring-cloud-stream/spring-cloud-stream-rabbit/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+    <parent>
+        <groupId>org.baeldung</groupId>
+        <artifactId>spring-cloud-stream</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>..</relativePath>
+    </parent>
+
+    <dependencies>
+   
+        <dependency>
```
KevinGilmore: use only spaces for indentation -- not tabs

spring-cloud/spring-cloud-stream/pom.xml
```diff
@@ -0,0 +1,70 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns="http://maven.apache.org/POM/4.0.0"
```
KevinGilmore: indent file with spaces instead of tabs

spring-cloud/spring-cloud-stream/spring-cloud-stream-rabbit/pom.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
KevinGilmore: Does this need to be a separate module? It seems premature to have a parent/aggregator module with only a single child module. If there are plans to have several spring-cloud-stream modules, then it's ok to leave as-is, otherwise let's just put it all in the spring-cloud-stream module.

spring-cloud/spring-cloud-stream/pom.xml
piksel-discovery: Ok

spring-cloud/spring-cloud-stream/spring-cloud-stream-rabbit/pom.xml
piksel-discovery: Ok

spring-cloud/spring-cloud-stream/spring-cloud-stream-rabbit/pom.xml
piksel-discovery: I think It's more than probably that we have a spring-cloud-stream-kafka for the future. For me, I would let it as it is. Do you agree?

spring-cloud/spring-cloud-stream/pom.xml
Sgitario: Ok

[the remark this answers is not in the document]
Sgitario: Ok

spring-cloud/spring-cloud-stream/spring-cloud-stream-rabbit/pom.xml
Sgitario: Ok

spring-cloud/spring-cloud-stream/spring-cloud-stream-rabbit/pom.xml
Sgitario: I think it's more than probably that we have a module spring-cloud-stream-kafka in the future. I would let it as it is. Do you agree?

KevinGilmore: Yes, I agree we can leave it as-is. This seems to be the way other spring-cloud modules are being organized in the repo.

spring-cloud/spring-cloud-stream/spring-cloud-stream-rabbit/src/test/java/com/baeldung/spring/cloud/stream/rabbit/MultipleOutputsServiceApplicationTests.java
```diff
@@ -0,0 +1,52 @@
...
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import com.baeldung.spring.cloud.stream.rabbit.processor.MyProcessor;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(classes = MultipleOutputsServiceApplication.class)
+@DirtiesContext
+public class MultipleOutputsServiceApplicationTests {
```
KevinGilmore: Class name should end in "UnitTest", "IntegrationTest", or "LiveTest", according to guidelines. If the tests can be run without any external dependencies (such as a server running a particular service), then go with "UnitTest". This comment applies to all of the test classes.

PR #2731, opened 2017-10-13
move security content from spring-security-rest-full

Doha2012:


spring-security-rest/src/test/java/org/baeldung/web/FooLiveTest.java
```diff
@@ -17,17 +17,20 @@
...
 
     private String cookie;
 
     private RequestSpecification givenAuth() {
         // return RestAssured.given().auth().form("user", "userPass", formConfig);
-        if (cookie == null) {
-            cookie = RestAssured.given().contentType("application/x-www-form-urlencoded").formParam("password", "userPass").formParam("username", "user").post(URL_PREFIX + "/login").getCookie("JSESSIONID");
-        }
-        return RestAssured.given().cookie("JSESSIONID", cookie);
+        // if (cookie == null) {
```
pivovarit: What about this code? why comment it out now?

spring-security-rest-full/.springBeans
```diff
@@ -1,18 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <beansProjectDescription>
 	<version>1</version>
-	<pluginVersion><![CDATA[3.8.2.201610040608-RELEASE]]></pluginVersion>
+	<pluginVersion><![CDATA[3.6.3.201411271034-RELEASE]]></pluginVersion>
```
pivovarit: Why are we changing to the 2014 version?

spring-security-rest/src/test/java/org/baeldung/web/FooLiveTest.java
Doha2012: When I was making sure everything is working, this test failed - so I fixed it
The security in this module is using http basic not form login

spring-security-rest-full/.springBeans
Doha2012: This is auto generated file, I have no idea why it changed

PR #2755, opened 2017-10-18
Mocking of private methods using PowerMock

sirishrenukumar:


pivovarit:
@sirishrenukumar since you opened a new one, I assume this one is not valid anymore - let me know if this is not the case

PR #2855, opened 2017-10-24
BAEL-1098: Testing a REST API with Karate

sybraimah:


rest-testing/pom.xml
```diff
@@ -150,6 +163,7 @@
         <rest-assured.version>2.9.0</rest-assured.version>
         <cucumber.version>1.2.5</cucumber.version>
         <wiremock.version>2.4.1</wiremock.version>
+        <karate.version>0.6.0</karate.version>
```
KevinGilmore: I see that 0.6.1 has just come out earlier this month. Let's go ahead and update the version.

rest-testing/src/test/java/com/baeldung/rest/cucumber/CucumberIntegrationTest.java
```diff
@@ -5,6 +5,6 @@
 import cucumber.api.junit.Cucumber;
 
 @RunWith(Cucumber.class)
-@CucumberOptions(features = "classpath:Feature")
+@CucumberOptions(features = "classpath:karate")
 public class CucumberIntegrationTest {
```
KevinGilmore: Does this class do anything? Looks like maybe a leftover from an earlier commit.

PR #2861, opened 2017-10-25
pazis.x@gmail.com

pazis:
This is the pull request for "Different Types of Bean Injection in Spring" as test article. I'm somehow unsure if it's the correct way to add code to your project. Please tell me if I had to do it in a different way.


pivovarit:
@pazis yes, the way is fine - I will close it now because we're not merging evaluation code anyway

PR #2890, opened 2017-10-28
group persistence modules

Doha2012:


pom.xml
```diff
@@ -243,11 +243,11 @@
 
         <module>rabbitmq</module>
         <module>vertx</module>
-        <module>spring-data-gemfire</module>
+        <module>persistence-modules/spring-data-gemfire</module>
```
pivovarit: @Doha2012 I think Spring Data articles probably deserve its own module. Let me double check

PR #2949, opened 2017-11-03
Testing with JSpec

Namozag:


testing/src/test/java/com/baeldung/jspec/JSpecTest.java
```diff
@@ -0,0 +1,57 @@
+package com.baeldung.jspec;
```
KevinGilmore: Rename class to JSpecUnitTest

testing/src/test/java/com/baeldung/jspec/CageTest.java
```diff
@@ -0,0 +1,114 @@
+package com.baeldung.jspec;
```
KevinGilmore: Rename class to CageUnitTest

testing/pom.xml
```diff
@@ -1,165 +1,171 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
KevinGilmore: Not sure why the huge difference -- could be different line endings? Just a guess, though. Main thing is to make sure your branch is up-to-date with the upstream repo.

PR #2952, opened 2017-11-04
BAEL-1272: Java String Pool

fpistritto:


core-java/src/test/java/com/baeldung/stringpool/StringPoolUnitTest.java
```diff
@@ -0,0 +1,43 @@
+package com.baeldung.stringpool;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class StringPoolUnitTest {
+
+    @Test
+    public void whenCreatingConstantStrings_thenTheirAddressesAreEqual() {
```
KevinGilmore: Very good test method names! Let's just switch to using AssertJ for assertions instead of JUnit, as per the recently announced policy.

fpistritto:
Thanks for the advice @KevinGilmore, I think I missed that policy change, still catching up :)

core-java/src/test/java/com/baeldung/stringpool/StringPoolUnitTest.java
fpistritto: Thanks! Just switched in the latest commit, could you please have a look? Thanks very much

fpistritto:
Hi @KevinGilmore, I updated the PR, please let me know your feedback, thanks very much!

PR #2954, opened 2017-11-04
BAEL-1296 Prime numbers till 'n'.

eralmas7:


pivovarit:
This one is not valid, right? let's close it then

PR #2985, opened 2017-11-07
BAEL 1185 Lazy Verification with Mockito 2 mini-article accompanying code

bahti:
BAEL 1185: Added unit test for the lazy verification mini-article to show multiple verifications can be executed and reported even if some of them fail.


mockito2/src/test/java/com/baeldung/mockito/java8/LazyVerificationTest.java
```diff
@@ -0,0 +1,34 @@
+package com.baeldung.mockito.java8;
```
pivovarit: mockito2 module is now in "testing-modules", let's move it there

mockito2/src/test/java/com/baeldung/mockito/java8/LazyVerificationTest.java
```diff
@@ -0,0 +1,34 @@
...
+    @Test
+    public void whenLazilyVerified_thenThrowsMultipleExceptions() {
+        VerificationCollector collector = MockitoJUnit.collector()
+            .assertLazily();
+
+        List mockList = mock(List.class);
+        verify(mockList).add("one");
+        verify(mockList).clear();
+
+        try {
```
pivovarit: Let's add the exception to the method signature and remove the try-catch

mockito2/src/test/java/com/baeldung/mockito/java8/LazyVerificationTest.java
bahti: I moved it testing-modules as suggested.

mockito2/src/test/java/com/baeldung/mockito/java8/LazyVerificationTest.java
bahti: Actually, the try-catch block there is to assert that both verification failures are reported when lazy verification is used, and is kind of part of the test. The try method is supposed to fail. I updated method name to make this intent clearer.

PR #3032, opened 2017-11-13
BAEL-1331 - Quick Example: Comparator vs Comparable in Java

gschambial:
Added Source code and Unit Test cases for article.


core-java/src/main/java/com/baeldung/comparator/PlayerAgeSorter.java
```diff
@@ -0,0 +1,26 @@
+package com.baeldung.comparator;
+
+import java.util.ArrayList;
+import java.util.Collections;
+
+public class PlayerAgeSorter {
+
+    public static void main(String[] args) {
+        
+        ArrayList<Player> footballTeam = new ArrayList<Player>();
```
lor6: let's use List<Player> for declaration in all examples

core-java/src/test/java/com/baeldung/comparator/ComparatorUnitTest.java
```diff
@@ -0,0 +1,42 @@
...
+    
+    @Test
+    public void whenUsingRankingComparator_thenSortedList() {
+        ArrayList<Player> footballTeam = new ArrayList<Player>();
+        Player player1 = new Player(59, "John", 20);
+        Player player2 = new Player(67, "Roger", 22);
+        Player player3 = new Player(45, "Steven", 24);
+        footballTeam.add(player1);
+        footballTeam.add(player2);
+        footballTeam.add(player3);
```
lor6: the code for creating the collection can be in a @Before method since it's the same for both tests

gschambial:
Added changes suggested by @lor6

PR #3099, opened 2017-11-22
BAEL-1328

andrethiago:
Code for mini-article about how to invert array in Java.


andrethiago:
Correcting indentation and added implementation in Commons Lang and Guava.

core-java/src/test/java/com/baeldung/array/ArrayInverterTest.java
```diff
@@ -0,0 +1,48 @@
...
+public class ArrayInverterTest {
+
+    private String[] fruits = { "apples", "tomatoes", "bananas", "guavas", "pineapples", "oranges" };
+
+    @Test
+    public void invertArrayWithForLoop() {
+        ArrayInverter inverter = new ArrayInverter();
+        inverter.invertUsingFor(fruits);
+
+        assertArrayEquals(new String[] { "oranges", "pineapples", "guavas", "bananas", "tomatoes", "apples" }, fruits);
```
KevinGilmore: Let's use AssertJ to do the assertions rather than the JUnit Assert class. Also, you could just create a field or constant to hold the inverted array, and use it in the tests.

core-java/src/test/java/com/baeldung/array/ArrayInverterTest.java
```diff
@@ -0,0 +1,48 @@
+package com.baeldung.array;
```
KevinGilmore: Rename to ArrayInverterUnitTest

PR #3148, opened 2017-11-27
fix parent.relativePath config for maven submodules in separate folders

hitme:
since some module have been moved to sub directories other than the root directory, pom.xml config should be adjusted accordingly.


pivovarit:
@hitme thanks! that's super helpful!

PR #3214, opened 2017-12-10
BAEL-1278

Namozag:


testing-modules/testing/src/test/java/com/baeldung/jspec/CageUnitTest.java
```diff
@@ -0,0 +1,126 @@
...
+public class CageUnitTest {
+
+    Cat tomCat = new Cat("Tom");
+    Cat felixCat = new Cat("Felix");
+    Dog boltDog = new Dog("Bolt");
+    Cage cage = new Cage();
+
+    
+    @Test
+    public void puttingAnimals_shouldIncreaseCageSize() {
```
KevinGilmore: Let's rename the tests according to convention. For example, you could call this one "whenPutAnimalInCage_thenShouldIncreaseCageSize".

PR #3247, opened 2017-12-14
Deep Dive into Default and Static Methods in Interfaces  - BAEL-1326

alejandrogervasio:


KevinGilmore:
Closing in lieu of #3281, which does not have the diff issue with the pom.xml

PR #3258, opened 2017-12-16
Added code for BAEL-1230: Getting log in JSON format

charro:
Code included into mini article BAEL-1230


charro:
Updated pull request after feedback from editor

PR #3333, opened 2017-12-31
[Bael 1209] - Java RMI files.

gautamshetty:
Java RMI files.


java-rmi/src/com/baeldung/rmi/MessengerServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.rmi.RemoteException;
+import java.rmi.registry.LocateRegistry;
+import java.rmi.registry.Registry;
+import java.rmi.server.UnicastRemoteObject;
+
+public class MessengerServiceImpl implements MessengerService {
+
+	@Override
+	public String sendMessage(String clientMessage) {
```
jzheaux: The code and the blog are mismatched here. Please change this method to "send".

java-rmi/src/com/baeldung/rmi/MessengerServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+		
+		String serverMessage = null;
+		if (clientMessage.equals("Client Message")) {
+			serverMessage = "Server Message";
+		}
+		
+		return serverMessage;
+	}
+	
+	public void createStubAndbind() throws RemoteException {
```
jzheaux: Camel case, e.g. createStubAndBind()

java-rmi/src/com/baeldung/rmi/MessengerServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+		return serverMessage;
+	}
+	
+	public static void main(String[] args) {
+		
+		MessengerServiceImpl msgInf = new MessengerServiceImpl();
+		try {
+			msgInf.createStubAndbind();
+		} catch (RemoteException e) {
+			e.printStackTrace();
```
jzheaux: Just my opinion (no action necessary), but the printing of the stack trace happens anyway, right? Not sure why the exception needs to be caught as opposed to putting RemoteException in the throws clause.

java-rmi/src/com/baeldung/rmi/RMIClientImpl.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class RMIClientImpl {
+
+	public static void main(String[] args) {
+		
+		try {
+			
+			Registry registry = LocateRegistry.getRegistry();
+			MessengerService server = (MessengerService) registry.lookup("MessengerService");
+			String responseMessage = server.sendMessage("Client Message");
```
jzheaux: Is this a helpful class without some kind of output to the console? When I run this, there would just be no response, right?

java-rmi/test/com/baeldung/rmi/JavaRMITest.java
```diff
@@ -0,0 +1,44 @@
...
+import static org.junit.Assert.fail;
+
+import java.rmi.NotBoundException;
+import java.rmi.RemoteException;
+import java.rmi.registry.LocateRegistry;
+import java.rmi.registry.Registry;
+
+import org.junit.Test;
+
+public class JavaRMITest {
```
jzheaux: This needs to be called JavaRMIIntegrationTest. XXXIntegrationTest is the naming convention we use for integration tests.

java-rmi/test/com/baeldung/rmi/JavaRMITest.java
```diff
@@ -0,0 +1,44 @@
...
+	
+	@Test
+	public void WhenRunServer_ServerIsStarted() {
```
jzheaux: Site standard is givenXXX_whenYYY_thenZZZ or whenYYY_thenZZZ, so this would be called: whenRunServer_thenServerStarts(). Note the camelCase as well as the additional "then"

java-rmi/test/com/baeldung/rmi/JavaRMITest.java
```diff
@@ -0,0 +1,44 @@
...
+		
+		try {
+			MessengerServiceImpl server = new MessengerServiceImpl();
+			server.createStubAndbind();
+			assertTrue(true);
```
jzheaux: This is unnecessary, right?

java-rmi/test/com/baeldung/rmi/JavaRMITest.java
```diff
@@ -0,0 +1,44 @@
...
+		} catch (RemoteException e) {
+			fail("Exception Occured");
```
jzheaux: (No action necessary) Is this typical of unit tests you have written? Usually, I will simply put RemoteException in the throws clause.

java-rmi/test/com/baeldung/rmi/JavaRMITest.java
```diff
@@ -0,0 +1,44 @@
...
+	
+	@Test
+	public void WhenClientSendsMessageToServer_ServerSendsResponseMessage() {
+		
+		try {
+			Registry registry = LocateRegistry.getRegistry();	
+			MessengerService server = (MessengerService) registry.lookup("MessengerService");			
+			String responseMessage = server.sendMessage("Client Message");			
+			String expectedMessage = "Server Message";			
+			assertEquals(responseMessage, expectedMessage);
```
jzheaux: The java doc says that expected goes first. Also, that will bring it into alignment with your post copy.

java-rmi/src/com/baeldung/rmi/Message.java
```diff
@@ -0,0 +1,36 @@
+package com.baeldung.rmi;
+
+import java.io.Serializable;
+
+public class Message implements Serializable {
+
```
jzheaux: Isn't it best practice to declare a serialVersionUID?

java-rmi/test/com/baeldung/rmi/JavaRMITest.java
jzheaux: Meaning the assertTrue(true) line

java-rmi/src/com/baeldung/rmi/Message.java
gautamshetty: Code guidelines mentions not to write it.

java-rmi/src/com/baeldung/rmi/MessengerServiceImpl.java
gautamshetty: Will do this.

java-rmi/src/com/baeldung/rmi/MessengerServiceImpl.java
gautamshetty: Will do this.

java-rmi/src/com/baeldung/rmi/MessengerServiceImpl.java
gautamshetty: The intention here is to let reader know that the exception needs to be handled from where the RMI APIs are called. Otherwise the program will crash. This is just an example.

java-rmi/src/com/baeldung/rmi/RMIClientImpl.java
gautamshetty: Just had written a client class. If only the integration test is to be included, this can be deleted.

java-rmi/test/com/baeldung/rmi/JavaRMITest.java
gautamshetty: Will do this.

java-rmi/test/com/baeldung/rmi/JavaRMITest.java
gautamshetty: Will do this.

java-rmi/test/com/baeldung/rmi/JavaRMITest.java
gautamshetty: Will do this.

java-rmi/test/com/baeldung/rmi/JavaRMITest.java
gautamshetty: According to code guidelines. Though there needs to be one line before the expected. Will do that.

java-rmi/src/com/baeldung/rmi/MessengerServiceImpl.java
gautamshetty: I think we should change the article from send() to sendMessage() for this.

java-rmi/src/com/baeldung/rmi/MessengerServiceImpl.java
gautamshetty: Sorry, this was just a temporary code I had written to test it. I will delete the main method from this class.

java-rmi/test/com/baeldung/rmi/JavaRMITest.java
gautamshetty: Either way works. This way just gives idea where it has failed.

java-rmi/src/com/baeldung/rmi/RMIClientImpl.java
gautamshetty: Code guidelines says not to write System.out.println(). Not sure that is just for articles or code in general. Thats why this has no output.

java-rmi/src/com/baeldung/rmi/MessengerServiceImpl.java
gautamshetty: Done.

gautamshetty:
I have updated the changes.

java-rmi/src/com/baeldung/rmi/Message.java
jzheaux: Oh, whoops! Okay, my mistake.

java-rmi/src/com/baeldung/rmi/RMIClientImpl.java
jzheaux: I'll check with @pivovarit to see what he thinks. Like you said, might be immaterial here since the class is likely unnecessary.

java-rmi/test/com/baeldung/rmi/JavaRMITest.java
jzheaux: Thank you for adding the "then". The "when" needs to be lowercased as well. Sorry for not being clearer.

java-rmi/test/com/baeldung/rmi/JavaRMIIntegrationTest.java
```diff
@@ -0,0 +1,44 @@
...
+import java.rmi.RemoteException;
+import java.rmi.registry.LocateRegistry;
+import java.rmi.registry.Registry;
+
+import org.junit.Test;
+
+public class JavaRMIIntegrationTest {
+	
+	@Test
+	public void WhenRunServer_thenServerStarts() {
```
jzheaux: Does it concern you that JUnit doesn't guarantee the order in which these methods will execute? Maybe converting this into a @Before or using the @FixMethodOrder annotation and then changing the name of the server one so that they can run in lexical order.

gautamshetty:
I have made changes regarding the integration tests.

PR #3353, opened 2018-01-04
fix start instead of run

lor6:


lor6:
Hey @pivovarit A very minor change here based on an issue. Also requires updating the article: http://www.baeldung.com/wp-admin/post.php?post=13871&action=edit

PR #3374, opened 2018-01-08
[BAEL 1209] - Java RMI Files.

gautamshetty:


java-rmi/src/test/java/com/baeldung/rmi/JavaRMIIntegrationTest.java
```diff
@@ -0,0 +1,45 @@
...
+import java.rmi.registry.LocateRegistry;
+import java.rmi.registry.Registry;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class JavaRMIIntegrationTest {
+	
+	@BeforeClass
+	public static void whenRunServer_thenServerStarts() {
```
jzheaux: Sorry, this was unclear, but doesn't this method name seem weird now that it isn't a test method? I guess I'd expect a name like "serverStartup" or "setUp". It also seems a little odd that we are asserting something inside a setup method.

java-rmi/src/main/java/com/baeldung/rmi/RMIClientImpl.java
```diff
@@ -0,0 +1,23 @@
+package com.baeldung.rmi;
+
+import java.rmi.NotBoundException;
+import java.rmi.RemoteException;
+import java.rmi.registry.LocateRegistry;
+import java.rmi.registry.Registry;
+
+public class RMIClientImpl {
```
jzheaux: I thought we were going to delete this? Let's go ahead and do that. I don't think it adds any value over and above your integration test.

java-rmi/pom.xml
```diff
@@ -0,0 +1,23 @@
...
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>3.8.1</version>
```
jzheaux: This needs to have the parent module in a <parent> tag. Go ahead and check a couple other poms for examples.

Also, the parent pom will include junit for you, so you can remove this.

Finally, this can't be part of the build unless this module "java-rmi" is listed in the module list in tutorials/pom.xml. Let's add that.

java-rmi/src/main/java/com/baeldung/rmi/RMIClientImpl.java
gautamshetty: I was waiting for your input, you did not reply so it was not deleted. I will do that.

You are yet to also give clarity on the system.out and similar code rules.

java-rmi/src/test/java/com/baeldung/rmi/JavaRMIIntegrationTest.java
gautamshetty: Doesn't satisfy the naming rules.

java-rmi/src/main/java/com/baeldung/rmi/RMIClientImpl.java
jzheaux: Thanks, Gautam. Personally, a standalone class that doesn't have an output doesn't have value--I think you would be safe to have a sysout, so long as it is just in the repo and not in the article.

java-rmi/src/test/java/com/baeldung/rmi/JavaRMIIntegrationTest.java
jzheaux: I appreciate your careful reading of the naming rules, but that rule only applies to tests. By annotating this with @BeforeClass, it is no longer a test, but is instead a setup method. If you want to change it back to a @Test, then that is one way to solve the issue as well; you simply need to configure each test so that it is executed in the correct order.

java-rmi/src/test/java/com/baeldung/rmi/JavaRMIIntegrationTest.java
```diff
@@ -0,0 +1,45 @@
...
+		
+		try {
+			Registry registry = LocateRegistry.getRegistry();	
+			MessengerService server = (MessengerService) registry.lookup("MessengerService");			
+			String responseMessage = server.sendMessage("Client Message");		
+			
+			String expectedMessage = "Server Message";			
+			assertEquals(responseMessage, expectedMessage);
+		} catch (RemoteException e) {
+			fail("Exception Occured");
```
jzheaux: Sorry to add one more thing: I just noticed this that Occurred is mispelled. It should be "Occurred" instead of "Occured". This is in multiple places. I only mention it here since you are already making a couple of other changes for the unit test.

pom.xml
```diff
@@ -86,7 +86,9 @@
         <module>jackson</module>
         <!--         <module>persistence-modules/java-cassandra</module> -->
         <module>vavr</module>
-		<module>java-lite</module>
+	<module>java-lite</module>
+	<module>java-lite</module>
```
jzheaux: Just checking, is this module accidentally listed twice?

java-rmi/src/test/java/com/baeldung/rmi/JavaRMIIntegrationTest.java
gautamshetty: There is no such mention in the guidelines. So, If you could please confirm it, I will do the change. Just don't want to do same changes again.

PR #3412, opened 2018-01-14
Refactoring-Thread Interrupt

khatwaniNikhil:


nawa:
@khatwaniNikhil Why `Thread.currentThread().interrupt()` was added and not `this.interrupt()` ?

PR #3420, opened 2018-01-15
BAEL-1464 Guava Memoizer

knightdna:
Sample codes for Guava Memoizer article.

### DIFF
--- a/guava/src/test/java/org/baeldung/guava/GuavaMemoizerUnitTest.java
+++ b/guava/src/test/java/org/baeldung/guava/GuavaMemoizerUnitTest.java
@@ -1,0 +1,101 @@
+package org.baeldung.guava;
+
+import com.google.common.base.Suppliers;
+import org.baeldung.guava.memoizer.CostlySupplier;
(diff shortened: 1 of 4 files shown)


guava/src/test/java/org/baeldung/guava/GuavaMemoizerUnitTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+        start = Instant.now();
+        bigNumber = memoizedSupplier.get().add(BigInteger.ONE);
+        duration = Duration.between(start, Instant.now()).toMillis();
+        assertThat(bigNumber, is(equalTo(new BigInteger("90840813214259257091017657299038202013732367616052353832494971785771209455315189760000000001"))));
+        assertThat(duration.doubleValue(), is(closeTo(0D, 100D)));
+
+        try {
+            TimeUnit.SECONDS.sleep(1);
+        } catch (InterruptedException e) {
```
pivovarit: Let's just add "throws" to the method signature

PR #3435, opened 2018-01-17
Added few config about KIE With Spring boot

pandiyan90:


lor6:
Thanks for the PR. We'll keep this as a non-Boot project for now. Closing.

PR #3459, opened 2018-01-19
BAEL-1148 earth001@gmail.com - Merged spring-mvc-push module in spring-mvc-simple

earth001:


spring-mvc-simple/pom.xml
```diff
@@ -8,83 +8,62 @@
...
+        <javax.servlet.jsp-api.version>2.3.2-b02</javax.servlet.jsp-api.version>
+        <javax.servlet-api.version>4.0.0</javax.servlet-api.version>
         <hibernate-validator.version>5.4.1.Final</hibernate-validator.version>
         <deploy-path>enter-location-of-server</deploy-path>
         <fileupload.version>1.3.2</fileupload.version>
         <java.version>1.8</java.version>
         <org.thymeleaf-version>3.0.7.RELEASE</org.thymeleaf-version>
         <groovy.version>2.4.12</groovy.version>
-        <freemarker.version>2.3.23</freemarker.version>
+        <freemarker.version>2.3.27-incubating</freemarker.version>
```
KevinGilmore: Why use an incubating version?

spring-mvc-simple/pom.xml
```diff
@@ -141,11 +133,18 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-compiler-plugin</artifactId>
-                   <version>3.7.0</version>
```
KevinGilmore: Why was this section removed?

spring-mvc-simple/src/main/java/com/baeldung/spring/configuration/JadeTemplateConfiguration.java
```diff
@@ -16,24 +16,24 @@
 public class JadeTemplateConfiguration {
```
KevinGilmore: The changes to this file and a couple of others below seem to be only spacing changes -- changing indents from tabs to spaces. Generally this is not advised, since it makes the change seem bigger than it actually is, and also because the reviewer has more files and lines of code to review.

PR #3475, opened 2018-01-21
add pact provider test

lor6:


spring-rest/src/main/java/org/baeldung/web/controller/PactController.java
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class PactController {
+
+    List<PactDto> pacts = new ArrayList<>();
+
+    @GetMapping(value = "/pact", produces = MediaType.APPLICATION_JSON_VALUE)
```
pivovarit: @lor6 isn't `MediaType.APPLICATION_JSON_VALUE` the default value here?

PR #3583, opened 2018-02-04
Priorityjobscheduler

rockoder:


core-java-concurrency/src/main/java/com/baeldung/concurrent/prioritytaskexecution/PriorityJobScheduler.java
```diff
@@ -9,21 +9,23 @@
...
+        priorityQueue = new PriorityBlockingQueue<Job>(queueSize, 
+          Comparator.comparing(Job::getJobPriority));
 
-        priorityJobScheduler = Executors.newSingleThreadExecutor();
         priorityJobScheduler.execute(()->{
             while (true) {
                 try {
                     priorityJobPoolExecutor.execute(priorityQueue.take());
                 } catch (InterruptedException e) {
+                    if (priorityQueue.size() > 0) {
```
pivovarit: Do we need that check here?

rockoder: I would prefer to remove it and just break. But then we go back to square one.

PR #3616, opened 2018-02-08
BAEL-1530 - Guide to java.lang.System

kashifma:
Code examples for article "Guide to java.lang.System"

BAEL-1530


jzheaux:
@kashifma, looks like we need to update the examples to match what is in the article. Try to convert what you can into unit tests. Anything that can't be a unit test (things that take input, System.gc, System.exit, etc.) go ahead and put the method from your article in a class and include a short usage description at the top in a comment block. I'll ask my senior editor for recommendations, too.

kashifma:
@jzheaux  yeah, I wanted that we finalize the examples in the article first so that we update all the code in one go in the end.

core-java/src/test/java/com/baeldung/system/SystemPropertiesTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.Properties;
+
+public class SystemPropertiesTest {
+
+    @Test
+    public void givenSystem_whenCalledGetProperties_thenReturnPropertiesinResult() {
```
jzheaux: Let's separate these out into separate tests. (The tests look great, by the way!) It will be easier for your readers to process if you are just testing one thing at a time.

core-java/src/test/java/com/baeldung/system/EnvironmentVariablesTest.java
```diff
@@ -0,0 +1,14 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+public class EnvironmentVariablesTest {
+
+    @Test
+    public void givenEnvVars_whenReadPath_thenGetValueinResult() {
+        EnvironmentVariables environmentVariables = new EnvironmentVariables();
+
+        Assert.assertNotNull(environmentVariables.getPath());
```
jzheaux: @pivovarit I don't want to make this harder than it needs to be, but this assertion bugs me. However, we can't rely on what folks are going to have as an environment variable.

My thought it to convert this to an Integration Test and then add instructions in the comments to run it like this:

```bash
NAME=value mvn integration-test
```

so that then this test could check for the `NAME` environment variable. 

Let me know if you can think of a better way, or if you are okay with this as a test.

core-java/src/test/java/com/baeldung/system/DateTimeServiceTest.java
```diff
@@ -0,0 +1,15 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+public class DateTimeServiceTest {
+
+    @Test
+    public void givenClass_whenCalledMethods_thenNotNullInResult() {
+        DateTimeService dateTimeService = new DateTimeService();
+
+        Assert.assertNotNull(dateTimeService.nowPlusOneHour());
```
jzheaux: @pivovarit same kind of question here, if you have any thoughts about making these more verifiable (they rely on `System.currentTimeMillis()`)

PR #3636, opened 2018-02-11
BAEL-1534: Detect the OS from Java

shouvikbhattacharya:


core-java/src/main/java/com/baeldung/system/DetectOS.java
```diff
@@ -0,0 +1,50 @@
+package com.baeldung.system;
+
+import java.util.Locale;
+
+import org.apache.commons.lang3.SystemUtils;
+
+public class DetectOS {
+
+    private static final DetectOS INSTANCE = new DetectOS();
+    private static OperatingSystem detectedOS;
```
pivovarit: Let's simplify the design and not store the result anywhere - let's simply return the result from the method

core-java/src/main/java/com/baeldung/system/DetectOS.java
```diff
@@ -0,0 +1,50 @@
+package com.baeldung.system;
+
+import java.util.Locale;
+
+import org.apache.commons.lang3.SystemUtils;
+
+public class DetectOS {
+
+    private static final DetectOS INSTANCE = new DetectOS();
```
pivovarit: No need to introduce a singleton here as well

core-java/src/main/java/com/baeldung/system/DetectOS.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public static DetectOS getInstance() {
+        return INSTANCE;
+    }
+
+    public OperatingSystem getOperatingSystem() {
+        if (detectedOS == null) {
+            String OS = System.getProperty("os.name", "generic")
+                .toLowerCase(Locale.ENGLISH);
+            if ((OS.indexOf("mac") >= 0)) {
```
pivovarit: Wouldn't it make sense to use "contains" instead of "indexOf" here?

core-java/src/main/java/com/baeldung/system/DetectOS.java
```diff
@@ -0,0 +1,50 @@
...
+                detectedOS = OperatingSystem.Linux;
+            } else {
+                detectedOS = OperatingSystem.Other;
+            }
+        }
+        return detectedOS;
+    }
+
+    public OperatingSystem getOperatingSystemSystemUtils() {
+        if (SystemUtils.IS_OS_WINDOWS)
```
pivovarit: That looks like a use-case for the switch-case construct

core-java/src/test/java/com/baeldung/system/WhenDetectingOS.java
```diff
@@ -0,0 +1,25 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import com.baeldung.system.DetectOS.OperatingSystem;
+
+public class WhenDetectingOS {
+
+    @Test
+    public void whenUsingSystemProperty_shouldReturnOS() {
```
pivovarit: Why is this working on Travis if it's linux-based?
After we sort it out, the test should be ignored so that it doesn't affect the build on various envs

shouvikbhattacharya: @pivovarit I am a bit lost here. I went through the build logs and didn't find anything related to this test case. I am under the impression that this didn't run otherwise the build should have failed. Can you help me out a bit? I will make the changes as suggested and ignore the tests in the test class so that it does lead to any build issues. I hope that should be fine?

pivovarit: I think you should try renaming the class to something that ends with "*Test" - this is how surefire picks up test classes

PR #3647, opened 2018-02-12
dennisca@gmail.com added spring-injection module

dennisca:
This is for evaluation article.


dennisca:
Close it since it is for evaluation article.

PR #3677, opened 2018-02-16
BAEL-1484: Intro to OpenCSV

Thoughtscript:
Code for BAEL-1484 - Using asynchronous methods to handle all CSV processes - will be adding a write example shortly too.

Also, as of 2/16/18 - the OpenCSV Sourceforge is being migrated - reference docs can be found for older versions here and there. Current spec is 4.1.


Thoughtscript:
Closing and reopening per request

PR #3678, opened 2018-02-16
 Sample code for BAEL-1159 - Working with Kotlin and JPA - earth001@gmail.com

earth001:


hibernate5/src/test/java/com/baeldung/hibernate/HibernateSpatialIntegrationTest.java
```diff
@@ -24,7 +24,7 @@
 
 import static org.junit.Assert.assertTrue;
 
-public class HibernateSpatialTest {
+public class HibernateSpatialIntegrationTest {
```
earth001: I renamed because test not passes, require a MySQL instance.

hibernate5/src/test/java/com/baeldung/hibernate/TemporalValuesIntegrationTest.java
```diff
@@ -15,7 +15,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-public class TemporalValuesTest {
+public class TemporalValuesIntegrationTest {
```
earth001: I renamed because test not passed, require a specific timezone.

PR #3683, opened 2018-02-16
BAEL-1556

kornelihno:
Spring Security with Thymeleaf


kornelihno:
need to remove previous code

PR #3736, opened 2018-02-26
BAEL-1226 Checkstyle introduction.

shreyasht:
Please review this commit for BAEL-1226. I have made it similar to Findbugs.

Regards,
Shreyash

### DIFF
--- a/spring-rest/pom.xml
+++ b/spring-rest/pom.xml
@@ -334,18 +334,5 @@
 		
 		<pact.version>3.5.11</pact.version>
 	</properties>
-	
-<reporting>
-    <plugins>
-        <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-checkstyle-plugin</artifactId>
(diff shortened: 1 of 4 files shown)


spring-rest/pom.xml
```diff
@@ -325,7 +325,8 @@
 		<!-- Maven plugins -->
 		<cargo-maven2-plugin.version>1.6.0</cargo-maven2-plugin.version>
 		<findbugs-maven-plugin.version>3.0.4</findbugs-maven-plugin.version>
-
+		<checkstyle-maven-plugin.version>3.0.0</checkstyle-maven-plugin.version>
+		<dependency.locations.enabled>false</dependency.locations.enabled>
```
shreyasht: Added this line so that build does not take a long time. If we remove this the following issue will happen:
https://stackoverflow.com/questions/26144541/why-am-i-getting-maven-error-unable-to-determine-if-resource-x-exists-in-http

PR #3740, opened 2018-02-27
Request for BAEL-1584

ankishagarwal:
Pull request for BAEL-1584 by ankishagarwal8@gmail.com

### DIFF
--- a/core-java-8/src/main/java/com/baeldung/findanelement/FindElementInAList.java
+++ b/core-java-8/src/main/java/com/baeldung/findanelement/FindElementInAList.java
@@ -2,23 +2,22 @@ package com.baeldung.findanelement;
 
 import java.util.List;
 import java.util.ListIterator;
-import java.util.Optional;
 import org.apache.commons.collections4.IterableUtils;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Iterables;
 
 public class FindElementInAList<T> {
 
-    public T findUsingIndexOf(T element, List<T> listOfIntegers) {
-        int index = listOfIntegers.indexOf(element);
+    public T findUsingIndexOf(T element, List<T> list) {
+        int index = list.indexOf(element);
         if (index >= 0) {
             return element;
         }
         return null;
     }
 
-    public T findUsingListIterator(T element, List<T> listOfIntegers) {
-        ListIterator<T> listIterator = listOfIntegers.listIterator();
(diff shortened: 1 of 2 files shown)


core-java-8/src/main/java/com/baeldung/findanelement/FindElementInAList.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.ListIterator;
+import java.util.Optional;
+import org.apache.commons.collections4.IterableUtils;
+import com.google.common.base.Predicate;
+import com.google.common.collect.Iterables;
+
+public class FindElementInAList<T> {
+
+    public T findUsingIndexOf(T element, List<T> listOfIntegers) {
+        int index = listOfIntegers.indexOf(element);
```
jzheaux: Since it is templated, these aren't really a `listOfIntegers`. Let's change the name to something more generic.

core-java-8/src/main/java/com/baeldung/findanelement/FindElementInAList.java
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+        return null;
+    }
+
+    public T findUsingStream(T element, List<T> listOfIntegers) {
+        Optional<T> foundElement = listOfIntegers.stream()
+            .filter(integer -> integer.equals(element))
+            .findFirst();
+        return foundElement.isPresent() ? foundElement.get() : null;
```
jzheaux: I kind of like the look of
```
T foundElement = listOfInteger.stream()
  .filter(integer -> integer.equals(element))
  .findFirst()
  .orElse(null);
```
Since it makes it match Guava. Thoughts?

core-java-8/src/main/java/com/baeldung/findanelement/FindElementInAList.java
ankishagarwal: Sure, will update the name to list

core-java-8/src/main/java/com/baeldung/findanelement/FindElementInAList.java
ankishagarwal: Yes, that's same.
Updating to orElse ,this would be helpful when use compares between different approaches.

ankishagarwal:
@jzheaux thanks for the suggestions. Updated pull request.

PR #3750, opened 2018-03-02
Bael 1585

alv21:


core-java-8/src/test/java/com/baeldung/java8/Java8MaxMinUnitTest.java
```diff
@@ -44,4 +46,29 @@ public void whenListIsOfPersonObjectThenMinCanBeDoneUsingCustomComparatorThrough
...
+
+        int min = Arrays.stream(integers)
+            .min()
+            .getAsInt();
+
+        assertEquals(7, min);
+    }
+
+    @Test
+    public void whenArrayIsOfPersonObjectThenMaxCanBeDoneUsingCustomComparatorThroughLambda() {
```
jzheaux: I edited the names of these unit tests in the article so they didn't bleed across the margin. Would you please update these names to match?

PR #3858, opened 2018-03-19
BAEL-681 (3)

knightdna:
New DynamoDB integration test sample without Spring


knightdna:
Changes will be moved to aws module. Closing PR.

PR #4010, opened 2018-04-15
Refactor Ignite samples

pivovarit:


lor6:
@pivovarit  I pushed a fix for the core-java tests, but I think you have to pull from master

PR #4016, opened 2018-04-15
Bael 1601 kotlin

pauljervis:
Code for Kotlin Collection Filter article


pauljervis:
Named the branch improperly. Should be 1601.

PR #4122, opened 2018-04-28
BAEL-1734 add different methods to find file extension

rajat-garg:
Add three methods to find file extension:
1. Using Apache Commons IO Library
2. Using String Handler
3. Using Guava Library

### DIFF
--- a/core-java/src/main/java/com/baeldung/extension/Extension.java
+++ b/core-java/src/main/java/com/baeldung/extension/Extension.java
@@ -5,19 +5,19 @@ import org.apache.commons.io.FilenameUtils;
 
 public class Extension {
     //Instead of file name we can also specify full path of a file eg. /baeldung/com/demo/abc.java
-    public String getExtensionByApacheCommonLib(String fileName) {
-        return FilenameUtils.getExtension(fileName);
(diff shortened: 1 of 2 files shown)


core-java/src/main/java/com/baeldung/extension/Extension.java
```diff
@@ -0,0 +1,23 @@
+package com.baeldung.extension;
+
+import com.google.common.io.Files;
+import org.apache.commons.io.FilenameUtils;
+
+public class Extension {
+    //Instead of file name we can also specify full path of a file eg. /baeldung/com/demo/abc.java
+    public String getExtensionByApacheCommonLib(String fileName) {
```
pauljervis: please update the argument name "fileName" to be "filename" to match the rest of the file.

rajat-garg: Done. I have changed the other names because it violates the camel case convention.

pauljervis: Camel Case only counts for multiple words. ["filename"]](http://www.dictionary.com/browse/filename) is a single word, not two.

PR #4192, opened 2018-05-06
BAEL-1388 Introduction to Java MSF4J Microservices

pxai:
PR for the BAEL-1388 Introduction to Java MSF4J Microservices article


spring-thymeleaf/pom.xml
```diff
@@ -12,6 +12,30 @@
         <version>1.0.0-SNAPSHOT</version>
     </parent>
 
+<<<<<<< HEAD
```
jzheaux: Looks like you might have accidentally committed a conflict

pom.xml
```diff
@@ -97,6 +97,7 @@
         <module>metrics</module>
         <module>maven</module>
         <module>mesos-marathon</module>
+		<module>msf4j</module>
```
jzheaux: whitespacing

msf4j/pom.xml
```diff
@@ -0,0 +1,16 @@
...
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>msf4j</artifactId>
+    <packaging>pom</packaging>
+    <parent>
+        <artifactId>parent-modules</artifactId>
+        <groupId>com.baeldung</groupId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+    <modules>
```
jzheaux: Typically, it's preferred to differentiate examples by package instead of by maven module. Would you be able to condense this to one project, `msf4j`, and then separate the examples by package?

msf4j/msf4jspring/mvnw.cmd
```diff
@@ -0,0 +1,143 @@
+@REM ----------------------------------------------------------------------------
```
jzheaux: We don't need the maven wrappers at this level. What we have at the parent level is fine, so let's remove these.

msf4j/msf4jspring/.gitignore
```diff
@@ -0,0 +1,25 @@
+/target/
```
jzheaux: I'm guessing these might have been added by start.spring.io? You should be able to remove them all, relying on the .gitignore in the parent.

pxai:
Hi there, I've put together the three projects in a single one, using different package names. Maven wrappers and other files are gone now.

PR #4228, opened 2018-05-12
A short example of real-time event streaming using Spring Webflux by inbox.saikat@gmail.com

psysane:
A quick and practical example of the Spring Webflux support on the server side, and on the client side, a simple Reactive Web client.

The endpoint produces one event (i.e. current time stamp) per second infinitely and the client side simply consumes and logs the events as they arrive.

`This PR is only for demo purpose and will be closed shortly.`


psysane:
This PR is only for demo purpose hence closing it.

PR #4241, opened 2018-05-13
Bael 4461

amit2103:


spring-mvc-simple/pom.xml
```diff
@@ -115,6 +115,13 @@
             <groupId>org.json</groupId>
             <artifactId>json</artifactId>
             <version>${json.version}</version>
+        </dependency>
+		<dependency>
```
pivovarit: @amit2103 why do we need this change?

amit2103: @pivovarit This PR is too old. We don't need it anymore.

PR #4427, opened 2018-06-07
Bootstrapping a JPA Implementation Programatically in Java

alejandrogervasio:


eugenp:
Quick note guys - keep in mind that we need to include the JIRA number in the name/message here, so that we know what this PR is related to over on JIRA.

PR #4453, opened 2018-06-10
Update download servlet to try-with-resources block.

pauljervis:


eugenp:
Hey @pauljervis - quick note - remember we need to have the JIRA id in the name of the PR - to know which JIRA this was related to. 
Cheers, 
Eugen.

PR #4477, opened 2018-06-13
rozagerardo@gmail.com - [No Jira ticket] - ServerSentEvent Streaming Using Webflux

rozagerardo:
* Add code for initial post about SSE streaming with webflux


rozagerardo:
Post link: http://inprogress.baeldung.com/?p=92027&preview=true

In this PR:
- I created an endpoint to retrieve a Flux<ServerSentEvent>
- In order to simulate a client consuming this stream of events, I created additional endpoints (in a separate package and application, to keep it clean), which just trigger an Async task
- The tasks launch a WebClient that retrieves and logs the stream of events from the endpoint
- I also used a WebTestClient which binds to the running server, in order to show the assertions that can be used with it

PR #4508, opened 2018-06-18
BAEL-1783

andrea-ligios:
BAEL-1783


gson/pom.xml
```diff
@@ -1,67 +1,74 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
KevinGilmore: Why are all the gson files in this PR? Are these leftover from another article/PR?

spring-boot-logging-log4j2/.mvn/wrapper/maven-wrapper.properties
```diff
@@ -0,0 +1 @@
+distributionUrl=https://repo1.maven.org/maven2/org/apache/maven/apache-maven/3.5.3/apache-maven-3.5.3-bin.zip
```
KevinGilmore: Do we need the maven-wrapper files in the repository?

spring-boot-logging-log4j2/src/test/java/com/baeldung/springbootlogging/SpringBootLoggingApplicationTests.java
```diff
@@ -0,0 +1,16 @@
+package com.baeldung.springbootlogging;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+public class SpringBootLoggingApplicationTests {
```
KevinGilmore: Integration test class names should end with "IntegrationTest" as per convention.

andrea-ligios: Those are the default test classes generated by Spring Initializr. I've not added anything, we could delete them because useless or leave them to match what is generated by start.spring.io, but not renaming them, it would be unnecessary misleading IMHO

PR #4533, opened 2018-06-23
[BAEL-1900] Calculate age using Java 8, Java 7 and Joda time libraries

psysane:
http://jira.baeldung.com/browse/BAEL-1900

### DIFF
--- a/core-java-8/src/main/java/com/baeldung/datetime/AddHoursToDate.java
+++ b/core-java-8/src/main/java/com/baeldung/datetime/AddHoursToDate.java
@@ -1,0 +1,54 @@
+package com.baeldung.datetime;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
+import java.util.Calendar;
+import java.util.Date;
+
+import org.apache.commons.lang3.time.DateUtils;
+
+public class AddHoursToDate {
+
+    public Date addHoursToJavaUtilDate(Date date, int hours) {
+        Calendar calendar = Calendar.getInstance();
+        calendar.setTime(date);
+        calendar.add(Calendar.HOUR_OF_DAY, hours);
+        return calendar.getTime();
+    }
+
+    public Date addHoursToDateUsingInstant(Date date, int hours) {
+        return Date.from(date.toInstant()
+            .plus(Duration.ofHours(hours)));
+    }
+
+    public LocalDateTime addHoursToLocalDateTime(LocalDateTime localDateTime, int hours) {
+        return localDateTime.plusHours(hours);
+    }
+
+    public LocalDateTime subtractHoursToLocalDateTime(LocalDateTime localDateTime, int hours) {
+        return localDateTime.minusHours(hours);
+    }
+
+    public ZonedDateTime addHoursToZonedDateTime(ZonedDateTime zonedDateTime, int hours) {
+        return zonedDateTime.plusHours(hours);
+    }
+
+    public ZonedDateTime subtractHoursToZonedDateTime(ZonedDateTime zonedDateTime, int hours) {
+        return zonedDateTime.minusHours(hours);
+    }
+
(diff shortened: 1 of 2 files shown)


core-java-8/pom.xml
```diff
@@ -1,200 +1,206 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
jzheaux: Looks like your editor may have changed the pom from spaces to tabs.

Let's change this back so that only the changes necessary for your article are present.

core-java-8/src/main/java/com/baeldung/datetime/AgeCalculator.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private AgeCalculator() {
+    }
+
+    public static int calculateAge(java.time.LocalDate birthDate, java.time.LocalDate currentDate) {
+        if (birthDate == null || currentDate == null) {
+            throw new IllegalArgumentException("Please enter a valid date!");
+        }
+        return Period.between(birthDate, currentDate)
+            .getYears();
```
jzheaux: Continuations should be indented in 2 spaces, not 4. However, you could also safely put this on the line above, if you wanted.

psysane:
@jzheaux Done 👍

core-java-8/pom.xml
```diff
@@ -94,6 +94,11 @@
             <artifactId>streamex</artifactId>
             <version>${streamex.version}</version>
         </dependency>
+        <dependency>
+			<groupId>joda-time</groupId>
```
jzheaux: The whitespacing is still incorrect here. Should be spaces, like the rest of the file.

psysane: Done 👍

jzheaux:
Looks great, @psysane 

We'll just wait for Grzegorz to chime in on the article. If he doesn't recommend any code changes, then we'll merge.

core-java-8/src/main/java/com/baeldung/datetime/AgeCalculator.java
herokuappsdevelopment: Provide separate null check for date of birth and current date

core-java-8/src/main/java/com/baeldung/datetime/AgeCalculator.java
```diff
@@ -0,0 +1,41 @@
...
+        }
+        Years age = Years.yearsBetween(birthDate, currentDate);
+        return age.getYears();
+    }
+
+    public static int calculateAgeWithJava7(Date birthDate, Date currentDate) {
+        if (birthDate == null || currentDate == null) {
+            throw new IllegalArgumentException("Please enter a valid date!");
+        }
+        DateFormat formatter = new SimpleDateFormat("yyyyMMdd");
```
jzheaux: That's a curious approach. Of course, there are probably dozens of ways to skin this cat without a library.

I think it would fail once we get to 10000 AD, but it appears safe until then. :)

core-java-8/src/test/java/com/baeldung/datetime/AgeCalculatorUnitTest.java
```diff
@@ -0,0 +1,49 @@
...
+    private final int DOB_YEAR = 1947;
+    private final int DOB_MONTH = 8;
+    private final int DOB_DAY = 15;
+
+    private final int YEAR = 2020;
+    private final int MONTH = 7;
+    private final int DAY = 12;
+    
+    private final String DATE_FORMAT = "yyyyMMdd";
+
```
jzheaux: Might be good to add a unit test to show what happens when currentDate precedes birthDate, or add a validation check in your methods to prevent it.

psysane: Done :)

psysane: Done :-)

psysane:
I think this PR should be merged to master. 
@eugenp

PR #4606, opened 2018-07-01
[BAEL-1930] Encryption and decryption of files using JDK

kiview:
PR for post http://jira.baeldung.com/browse/BAEL-1930


kiview:
Whoops, not sure how the deletion of the submodule got in here. 
The actual submodule seems wrong though, so should be okay to remove this here as well?

PR #4614, opened 2018-07-02
Added some sample code for example of event stream at the rate of One event per second.

sandy03934:


sandy03934:
Closing pull request as this is a code related to evaluation article.

PR #4619, opened 2018-07-03
BAEL-1889 - Let's move the Java Number articles into a new module

JonCook:


java-numbers/pom.xml
```diff
@@ -0,0 +1,281 @@
...
+			<groupId>org.openjdk.jmh</groupId>
+			<artifactId>jmh-core</artifactId>
+			<version>${jmh-core.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.openjdk.jmh</groupId>
+			<artifactId>jmh-generator-annprocess</artifactId>
+			<version>${jmh-generator-annprocess.version}</version>
+		</dependency>
+		<dependency>
```
jzheaux: Double-check some of the whitespacing here. Looks like these few dependencies don't have the same indentation as the reast.

java-numbers/pom.xml
```diff
@@ -0,0 +1,281 @@
...
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>${org.slf4j.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>${logback.version}</version>
+            <!-- <scope>runtime</scope> -->
```
jzheaux: Is there a reason this comment-out needs to be here (like for an article that references it?). If not, let's clean it up.

java-numbers/pom.xml
```diff
@@ -0,0 +1,281 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>${assertj.version}</version>
+            <scope>test</scope>
+        </dependency>
+	</dependencies>
+	
+	<build>
```
jzheaux: The indentation looks off here. Note the closing one and how `</plugin>` is more indented than `<plugin>` is here under `<builld>`

java-numbers/pom.xml
```diff
@@ -0,0 +1,281 @@
...
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+	
+	<properties>
+		<commons-math3.version>3.6.1</commons-math3.version>
+        <decimal4j.version>1.0.3</decimal4j.version>
```
jzheaux: Whitespacing

java-numbers/src/main/java/com/baeldung/numberofdigits/NumberOfDigitsDriver.java
```diff
@@ -1,35 +1,36 @@
...
-import org.apache.log4j.Logger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class NumberOfDigitsDriver {
     private static NumberOfDigits numberOfDigits;
-    
-    private static Logger LOG = Logger.getLogger(NumberOfDigitsDriver.class);
-    
+
```
jzheaux: Hmm, while I agree that `slf4j` is totally the way to go, I think this task should just be about moving the classes over. Would you be okay with moving this back to `log4j` and logging an enhancement?

java-numbers/src/main/java/com/baeldung/numberofdigits/NumberOfDigitsDriver.java
```diff
@@ -1,35 +1,36 @@
...
         LOG.info("Testing all methods...");
-        
+
         long length = numberOfDigits.stringBasedSolution(602);
-        LOG.info("String Based Solution : " + length);
-        
+        LOG.info("String Based Solution : {}", +length);
+
         length = numberOfDigits.logarithmicApproach(602);
-        LOG.info("Logarithmic Approach : " + length);
```
jzheaux: Same comment about just moving the tests and not improving their contents.

java-numbers/src/main/java/com/baeldung/prime/BruteForcePrimeChecker.java
```diff
@@ -1,16 +1,14 @@
-package com.baeldung.algorithms.primechecker;
+package com.baeldung.prime;
```
jzheaux: What is your reasoning for the package name change?

If it is something like "readability" or "cosmetics", then let's try and avoid this for this improvement. (While I do think these changes are great, they can be done in a different ticket)

JonCook:
Thanks for the comments. I will resubmit the PR.

JonCook:
@jzheaux - Thanks for the comments! I updated the PR. I'm not sure what happened with the formatting in the pom.xml but I think it should be OK now. Probably a tab setting in my xml editor in eclipse.

Sorry for including additional improvements that weren't part of the ticket. It is just the way I work, to clean often, frequently and in small incrementations. As I was creating a new module I thought it was a good opportunity to do this following the recommendations in the dev docs. But of course happy if this is done under a different improvement. But the tendency is this type of cleaning/improvements tend to get forgotten ;-)

I reverted to log4j and put the original package names back. Thanks again for the comments.

jzheaux:
@JonCook The issue is that associated code in the articles also needs to be edited as part of that. Changing package names is one thing since those don't really end up in the article code samples, but changing more substantial things is a slippery slope that can end in confusing readers who see one thing in the article and another in the codebase.

Feel free to ping Jira Support about adding another enhancement task to update articles to slf4j (etc.), and it's possible they will say that it is easier to do in this PR--it just means you'd be editing the contents of those articles, too.

JonCook:
@jzheaux  - Thanks for the comments again ;-) I'm still learning and quite new so agree with your suggestions. I don't think the formatting changes should have much impact on the articles. I had a quick look and most of it seems to be the same. Thanks again.

PR #4644, opened 2018-07-06
Delete the root element from the tree.

bi-:


eugenp:
Hey @bi-  - what is this one about? Is it related to a specific JIRA?

bi-:
Hi,

No jira issue, It just fixes the delete method. 

It updates the tree if the delete methods gets the same value as the root element holds, earlier it did not update it. See the test method.

eugenp:
I see, thanks for clearing that up, @bi-  
Quick question - are you an author with the site?

eugenp:
Oh, OK, thanks for clearing that up, @bi-  :)

lor6:
Hi @bi-  Thanks for the PR. I want to merge but there's a conflict.

bi-:
fixed, please check again.

lor6:
@bi-  Merged, thank you.

PR #4668, opened 2018-07-09
Fix Trie implementation

pivovarit:


data-structures/src/main/java/com/baeldung/trie/Trie.java
```diff
@@ -51,7 +51,7 @@ private boolean delete(TrieNode current, String word, int index) {
         if (node == null) {
             return false;
         }
-        boolean shouldDeleteCurrentNode = delete(node, word, index + 1);
+        boolean shouldDeleteCurrentNode = delete(node, word, index + 1) && !node.isEndOfWord();
```
pivovarit: @fatosmorina ping for visibility

fatosmorina: @pivovarit Thanks :)

PR #4707, opened 2018-07-14
Bael 1273 Spring RSS Feed View

aurasphere:


jzheaux:
@aurasphere it looks like the build is failing, possibly due to RssFeedApplication conflicting with another Application class. Will you please look into this?

PR #4884, opened 2018-08-01
Add Subject alternative name

rzucareli:
I add a Subject alternative name (SAN) parameter in all commands. This make the generated certificate work in Google Chrome. Before that change Chrome presents a ERR_CERT_COMMON_NAME_INVALID even when you had add the certificate as a know Authority.


lor6:
@rzucareli  Thanks :) Merged.

PR #4947, opened 2018-08-11
Spring Security Taglibs

nnhai1991:
Spring Security Taglibs demo code


spring-boot-security-taglibs/README.md
```diff
@@ -0,0 +1 @@
+
```
jzheaux: You can probably just remove this file as Grzegorz typically comes around and adds these after.

spring-boot-security-taglibs/src/main/java/org/baeldung/security/ApplicationConfig.java
```diff
@@ -0,0 +1,40 @@
...
+
+@SpringBootApplication
+@Configuration
+@EnableWebSecurity
+public class ApplicationConfig extends WebSecurityConfigurerAdapter {
+    public static final String DEFAULT_PASSWORD = "password";
+
+    @Override
+    protected void configure(AuthenticationManagerBuilder auth) throws Exception {
+        PasswordEncoder encoder = PasswordEncoderFactories.createDelegatingPasswordEncoder();
```
jzheaux: I applaud your extra efforts to teach good security practices.

It is my opinion, however, that this extra work is unnecessary because I think that the typical reader sees that you are using inMemoryAuthentication.

If you take a look at the [Java Docs for InMemoryUserDetailsManager](https://docs.spring.io/spring-security/site/docs/4.2.6.RELEASE/apidocs/org/springframework/security/provisioning/InMemoryUserDetailsManager.html), it says:

> Mainly intended for testing and demonstration purposes

And I don't believe that folks will use this part of your code for the long-term basis of their app.

Further, since the password is in-memory, there is no password at rest to be stolen, and so the encoding used is immaterial. If they were to steal it from the string pool, they have it already anyway in the way of `DEFAULT_PASSWORD`.

Again, I like you security awareness, but it is just noise in this article.

Keep in mind that the very authors of the library you are demonstrating don't add this complexity to their intro examples.

spring-boot-security-taglibs/src/main/java/org/baeldung/security/ApplicationConfig.java
```diff
@@ -0,0 +1,40 @@
...
+            .withUser("testUser")
+            .password(encoder.encode(DEFAULT_PASSWORD))
+            .roles("ADMIN");
+    }
+
+    @Override
+    protected void configure(HttpSecurity http) throws Exception {
+        http.csrf();
+
+        http.authorizeRequests()
```
jzheaux: Consider the formatting guidelines published by Spring Security: https://spring.io/blog/2013/07/11/spring-security-java-config-preview-readability/

spring-boot-security-taglibs/src/main/java/org/baeldung/security/HomeController.java
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@Controller
+@RequestMapping("/")
+public class HomeController {
+    @RequestMapping("")
+    public String home(HttpServletRequest request, HttpServletResponse response) {
```
jzheaux: `request` and `response` here are unnecessary. Let's remove them so that readers can spend more time looking at the relevant code in your article and less at the boilerplate.

spring-boot-security-taglibs/src/main/java/org/baeldung/security/HomeController.java
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@Controller
+@RequestMapping("/")
+public class HomeController {
+    @RequestMapping("")
```
jzheaux: Would @GetMapping be more appropriate here? Also, I'm unclear whether or not "" is actually necessary here. Could you remove the annotation altogether?

spring-boot-security-taglibs/src/main/resources/application.properties
```diff
@@ -0,0 +1,8 @@
+#jsp config
+spring.mvc.view.prefix: /WEB-INF/views/
+spring.mvc.view.suffix: .jsp
+spring.http.encoding.charset=UTF-8
```
jzheaux: If the encoding here is not strictly necessary, let's add a comment here that separates these from what is necessary for supporting JSPs. Something like:

```yaml
# The encoding settings here are not strictly necessary 
# for adding spring security tags to your JSPs, but they 
# are a good practice for secure and resilient JSPs themselves.
```

The reason for this is because many readers are probably already using Spring Security in a way that they like, but they need to know how to add JSP support. If they already have encoding settings that they are satisfied with, we don't want to miscommunicate that these are required (e.g. that they need to change their existing settings).

I'm sure you've been there before. You are trying something new, following a tutorial, something doesn't work, and so you change your code to be exactly like theirs. Well, let's help them out by differentiating what is necessary and what isn't.

And if they are necessary, then go ahead and ignore all of that. :)

spring-boot-security-taglibs/src/main/webapp/WEB-INF/views/home.jsp
```diff
@@ -0,0 +1,28 @@
+<%@ page language="java" contentType="text/html; charset=ISO-8859-1" pageEncoding="ISO-8859-1"%>
```
jzheaux: I'm curious why you use a pageEncoding here of ISO-8859-1 while taking the time to set up the app for UTF-8?

spring-boot-security-taglibs/src/test/java/org/baeldung/security/HomeControllerTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(webEnvironment = WebEnvironment.RANDOM_PORT)
+public class HomeControllerTest {
+
+    @Autowired
+    private TestRestTemplate restTemplate;
+
+    @Test
+    public void home() throws Exception {
```
jzheaux: Tests on baeldung should follow the naming convention givenXXX_whenYYY_thenZZZ or whenYYY_thenZZZ.

Also, please let's have a different test method for each. You are actually already doing this because of your comment above each assertion. Instead you can achieve both by doing:

```java
@Test
public void whenUserIsAuthenticatedThenAuthenticatedSectionsShowOnSite() {}
```

etc.

Also, please demonstrate what happens when they are not authenticated.

spring-boot-security-taglibs/src/test/java/org/baeldung/security/HomeControllerTest.java
```diff
@@ -0,0 +1,42 @@
...
+        String body = this.restTemplate.withBasicAuth("testUser", ApplicationConfig.DEFAULT_PASSWORD)
+            .getForEntity("/", String.class)
+            .getBody();
+        System.out.println(body);
```
jzheaux: According to the site guidelines, please remove all Sysouts before merging.

spring-boot-security-taglibs/src/main/webapp/WEB-INF/views/home.jsp
```diff
@@ -0,0 +1,28 @@
...
+        AUTHENTICATED
+    </sec:authorize>
+    <sec:authorize access="hasRole('ADMIN')">
+        ADMIN ROLE
+    </sec:authorize>
+    <h2>
+        principal.username:
+        <sec:authentication property="principal.username" />
+    </h2>
+    <form method="post" action="/do/something">
```
jzheaux: Could we have this go somewhere legit in the app? I'm okay with:

```html
<form method="post">
```
In which case @GetMapping might not be the right fit (see an earlier comment).

spring-boot-security-taglibs/src/main/webapp/WEB-INF/views/home.jsp
```diff
@@ -0,0 +1,28 @@
...
+        <sec:csrfInput />
+        Text Field:
+        <br />
+        <input type="text" name="textField" />
```
jzheaux: Consider changing this to a submit button, like maybe "Submit CSRF Token".

I'm not convinced we should do this since it distracts a bit from the app's purpose, but otherwise they won't be able to see that the csrfInput tag really works.

spring-boot-security-taglibs/src/main/webapp/WEB-INF/views/home.jsp
```diff
@@ -0,0 +1,33 @@
...
+    <sec:authorize access="isAuthenticated()">
+        AUTHENTICATED Content 
+        <sec:authorize access="hasRole('ADMIN')">
+            Content for users who have the "ADMIN" role.
+        </sec:authorize>
+        <h2>
+            Welcome back, <sec:authentication property="name" />
+        </h2>
+        <form>
+            <sec:csrfInput />
```
jzheaux: What I mean here is that without an endpoint that actually requires CSRF protection, then they can't really see this one in action. So, what thinking, what if you did this:

```html
<form method="post">
  <sec:csrfInput />
  <input type="submit" value="Do a post">
</form>
```

Then, folks can interact with the form.

Let me know if I'm misunderstanding or not being clear.

PR #5059, opened 2018-08-25
BAEL-1973

Thoughtscript:
Updated some older WebSocket config and improved the GUI for both socket examples.

### DIFF
--- a/spring-security-mvc-socket/pom.xml
+++ b/spring-security-mvc-socket/pom.xml
@@ -11,9 +11,9 @@
 
     <parent>
         <groupId>com.baeldung</groupId>
-        <artifactId>parent-spring-5</artifactId>
+        <artifactId>parent-spring-4</artifactId>
         <version>0.0.1-SNAPSHOT</version>
-        <relativePath>../parent-spring-5</relativePath>
+        <relativePath>../parent-spring-4</relativePath>
     </parent>
 
     <dependencies>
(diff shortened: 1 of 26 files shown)


spring-security-mvc-socket/pom.xml
```diff
@@ -11,9 +11,9 @@
 
     <parent>
         <groupId>com.baeldung</groupId>
-        <artifactId>parent-spring-5</artifactId>
+        <artifactId>parent-spring-4</artifactId>
```
KevinGilmore: Are we downgrading from Spring 5.x to 4.x?

spring-security-mvc-socket/src/main/java/com/baeldung/springsecuredsockets/controllers/SocketController.java
```diff
@@ -19,10 +27,20 @@
...
+    public OutputMessage sendAll(Message msg) throws Exception {
         OutputMessage out = new OutputMessage(msg.getFrom(), msg.getText(), new SimpleDateFormat("HH:mm").format(new Date()));
         return out;
     }
+
+    /**
+     * Example of sending message to specific user using 'convertAndSendToUser()' and '/queue'
+     */
+    @MessageMapping(SECURED_CHAT_ROOM)
+    //@SendToUser(SECURED_CHAT_SPECIFIC_USER)
```
KevinGilmore: Is this line commented out intentionally? If so, is there a reason to leave the line in there, or could we just delete it?

spring-security-mvc-socket/pom.xml
Thoughtscript: @KevinGilmore - just saw your comment on here. I replied over on JIRA - there was an unintentional upgrading from the original dependencies used by the other WebSocket article in here. It should have stayed at Spring-4. Spring-5 encourages SSE and Reactive WebSockets. This article specifies how to improve Spring WebSockets to send messages to specific users and not broadcast them to everyone. Sorry for any confusion!

PR #5081, opened 2018-08-27
BAEL-1726 - Serverless Spring Boot with Spring Cloud Functions 

aravindram87:


spring-cloud/spring-cloud-functions/spring-cloud-function-local/src/test/java/com/baeldung/spring/cloudfunction/CloudFunctionApplicationTests.java
```diff
@@ -0,0 +1,16 @@
+package com.baeldung.spring.cloudfunction;
```
jzheaux: What purpose does this class serve? Can we remove it?

spring-cloud/spring-cloud-functions/spring-cloud-function-local/src/test/java/com/baeldung/spring/cloudfunction/CloudFunctionApplicationTests.java
```diff
@@ -0,0 +1,16 @@
...
+import org.junit.runner.RunWith;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+public class CloudFunctionApplicationTests {
+
+    @Test
+    public void contextLoads() {
```
jzheaux: Or can we add a test that stands up the endpoint and invokes it?

spring-cloud/spring-cloud-functions/spring-cloud-function-aws/src/main/java/com/baeldung/spring/cloudfunction/CloudFunctionAwsApplication.java
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.Bean;
+
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+@SpringBootApplication
+public class CloudFunctionAwsApplication {
```
jzheaux: The duplication bugs me a bit.

One thing to remember that is different about writing code in a tutorial setting is that we may have many more tutorials on Spring Cloud Function, each with their own setup. 

Is it possible to have the web and aws dependencies in the same pom and then make sure that your package naming convention leaves room for other demos in the future to use your same module, but isolate themselves to their own package? This tends to be a bit more ideal long-term.

PR #5096, opened 2018-08-28
Task/bael 8504

dkapil:


dkapil:
@lor6 I had to leave the surefire configs of spring-rest and spring-rest-simple because those configs were used in explaining maven cargo plugin article, but there is nothing to worry as these configs are aligned with our main configs.

PR #5152, opened 2018-09-03
BAEL-1868

shreyasht:


shreyasht:
hey @jzheaux 
I did not get this part "What would you think about removing the parent pom directly here? Is the extra layer necessary?"

Let me try removing the logback.xml
Also, adding the module to parent.

Expect the changes by tomorrow.

shreyasht:
Hi @jzheaux 
I have done the below changes:

Removed the logback.xml
Added the module to parent.

PR #5165, opened 2018-09-04
BAEL-971 - An Introduction to Apache Commons Lang

alejandrogervasio:


libraries/src/main/java/com/baeldung/commons/lang3/application/Application.java
```diff
@@ -0,0 +1,8 @@
+package com.baeldung.commons.lang3.application;
+
+public class Application {
```
pivovarit: We don't need that class, right?

alejandrogervasio: Actually, we don't. I just kept it just in case the reader want's to drop the library's utility classes in it, rather than in the unit tests.

PR #5176, opened 2018-09-06
Fixed assignment of restTemplate

smic1909:
Fixed assignment of restTemplate


lor6:
@smic1909  Thanks, merged.

PR #5185, opened 2018-09-07
BAEL-2075: Lombok @Getter for boolean field

codehunter34:
BAEL-2075: Lombok @Getter for boolean field


codehunter34:
I will create a new PR.

PR #5205, opened 2018-09-09
BAEL-2070 Qualifier used instead of Primary 

xamcross:
Qualifier used instead of Primary to distinguish between different repository beans


xamcross:
BAEL-2095 CrudRepository save() method has been added to the pull request

PR #5222, opened 2018-09-10
Proxy in Hibernate load() method

fscorro:


hibernate5/src/test/java/com/baeldung/hibernate/proxy/HibernateProxyUnitTest.java
```diff
@@ -0,0 +1,75 @@
...
+        if(session != null) {
+            session.close();
+        }
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void givenAnInexistentEmployeeId_whenUseGetMethod_thenReturnNull() {
+        Employee employee = session.get(Employee.class, new Long(14));
+        assertNull(employee);
+        employee.getId();
```
jzheaux: I don't understand the purpose of this line. Didn't we just assert that employee is null? So why are we calling it?

hibernate5/src/test/java/com/baeldung/hibernate/proxy/HibernateProxyUnitTest.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Test
+    public void givenAnInexistentEmployeeId_whenUseLoadMethod_thenReturnAProxy() {
+        Employee employee = session.load(Employee.class, new Long(14));
+        assertNotNull(employee);
```
jzheaux: We should actually confirm that it is a proxy in this test.

hibernate5/src/test/java/com/baeldung/hibernate/proxy/HibernateProxyUnitTest.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
```
jzheaux: I think we are missing a couple of tests here.

In the tutorial, there is a statement that when using load and then invoking a method, we get an exception. That would be an informative test.

More importantly, though, the examples regarding what happens with relationships, with batch size, etc. aren't represented in the tests. We should add those so the reader can see those in action.

hibernate5/src/test/java/com/baeldung/hibernate/proxy/HibernateProxyUnitTest.java
```diff
@@ -0,0 +1,75 @@
...
+    @Test
+    public void givenABatchEmployeeList_whenSaveOne_thenSaveTheWholeBatch() {
```
jzheaux: What does this test have to do with the tutorial? I don't see how this test relates to understanding Hibernate.load.

hibernate5/src/main/java/com/baeldung/hibernate/proxy/Boss.java
```diff
@@ -0,0 +1,49 @@
+package com.baeldung.hibernate.proxy;
+
+import javax.persistence.*;
+import java.io.Serializable;
+
+@Entity
+public class Boss implements Serializable {
```
jzheaux: Could we change this to "Company"? My reasoning is that it offers a cleaner topological separation between the two entities. In the real work, a boss _is_ an employee, too, so there is a bit more cognitive load when the reader is thinking about a boss and an employee.

PR #5238, opened 2018-09-13
BAEL-1726 - Serverless Spring Boot with Spring Cloud Functions

aravindram87:


aravindram87:
@jzheaux My other PR #5238 went into a weird state. So, I closed it and created a new one. I have also implemented your review comments.

jzheaux:
@aravindram87 Looks good. Just one more thing, then, and that is to make sure that this new module is part of the build. You'll probably need to edit the `spring-cloud` parent pom to include your `spring-cloud-function-aws` module.

PR #5254, opened 2018-09-14
BAEL-2000 - Shreyas Mahajan - Adding a project for dockerizing Spring boot application with Jib

shreyasm:


jzheaux:
@shreyasm Let's do just one more adjustment, and that is regarding the module name.

`spring-boot-jib` implies that we are talking about some Spring Boot Jib integration, but there is no such thing. It would be better to have this module called `jib`. Then, if Spring Boot does ever have some kind of special support for Jib, it won't confuse folks.

shreyasm:
This is old pull request, not needed anymore. Hence, closing it.

PR #5287, opened 2018-09-18
BAEL-2164 added password hashing code and tests

MrSamMillington:


core-java/src/main/java/com/baeldung/passwordhashing/PBKDF2Hasher.java
```diff
@@ -0,0 +1,159 @@
...
+    byte[] dk = pbkdf2(password, salt, 1 << cost);
+    byte[] hash = new byte[salt.length + dk.length];
+    System.arraycopy(salt, 0, hash, 0, salt.length);
+    System.arraycopy(dk, 0, hash, salt.length, dk.length);
+    Base64.Encoder enc = Base64.getUrlEncoder().withoutPadding();
+    return ID + cost + '$' + enc.encodeToString(hash);
+  }
+
+
+  public String hashSimple(String password) throws Exception{
```
jzheaux: Can we put this elsewhere and make it static? We could maybe put it down towards the end of the file or in a different file. My concern is that it gets in the way of the hash/authenticate/pbkdf2 combination of methods.

core-java/src/main/java/com/baeldung/passwordhashing/PBKDF2Hasher.java
```diff
@@ -0,0 +1,159 @@
...
+    byte[] hash = f.generateSecret(spec).getEncoded();
+    return String.valueOf(hash);
+  }
+
+  /**
+   * Authenticate with a password and a stored password token.
+   *
+   * @return true if the password and token match
+   */
+  public boolean authenticate(char[] password, String token)
```
jzheaux: I might call this `checkPassword` or `matches` instead. `authenticate` is clear enough, but it has a broader meaning (granting privileges, etc.).

core-java/src/main/java/com/baeldung/passwordhashing/SHA512Hasher.java
```diff
@@ -0,0 +1,35 @@
...
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+
+/** A really simple SHA_512 Encryption example.
+ *
+ */
+public class SHA512Hasher {
+
+  public String get_SHA_512_SecurePassword(String passwordToHash, byte[] salt){
```
jzheaux: Could we name and design this the same as the equivalent in PBKDF2Hasher? I think the symmetry will help readers find commonalities.

core-java/src/main/java/com/baeldung/passwordhashing/PBKDF2Hasher.java
```diff
@@ -0,0 +1,159 @@
...
+    Matcher m = layout.matcher(token);
+    if (!m.matches())
+      throw new IllegalArgumentException("Invalid token format");
+    int iterations = iterations(Integer.parseInt(m.group(1)));
+    byte[] hash = Base64.getUrlDecoder().decode(m.group(2));
+    byte[] salt = Arrays.copyOfRange(hash, 0, SIZE / 8);
+    byte[] check = pbkdf2(password, salt, iterations);
+    int zero = 0;
+    for (int idx = 0; idx < check.length; ++idx)
+      zero |= hash[salt.length + idx] ^ check[idx];
```
jzheaux: Thank you for not exiting early, (e.g. timing attacks). Nice!

MrSamMillington:
Made a few changes as per your request. Let me know if I can tidy it up further.

jzheaux:
@MrSamMillington I noticed that you changed the method names for `SHA512Hasher` but not for `PBKDF2Hasher`. Would you be okay with changing those, too, so that the method names align?

MrSamMillington:
@jzheaux So 'checkPassword' instead of 'authenticate' for PBKDF2Hasher too?

jzheaux:
Let me know if I'm off my rocker, here, too. What I'm saying is that your readers will want to compare the two hashers, and if they have the same method names, it will be easier for them to follow.

PR #5374, opened 2018-10-01
Class Refactor

fscorro:
New tests in Hibernate with proxy


hibernate5/src/test/java/com/baeldung/hibernate/proxy/HibernateProxyUnitTest.java
```diff
@@ -36,40 +43,68 @@ public void close(){
...
+    @Test
     public void givenAnInexistentEmployeeId_whenUseGetMethod_thenReturnNull() {
-        Employee employee = session.get(Employee.class, new Long(14));
+        Employee employee = session.get(Employee.class, 14L);
+        assertNull(employee);
+    }
+
+    @Test(expected = ObjectNotFoundException.class)
+    public void givenAnInexistentEmployeeId_whenUseLoadMethod_thenThrowObjectNotFoundException() {
+        Employee employee = session.load(Employee.class, 999L);
```
jzheaux: This test is a bit confusing. The expected exception actually gets thrown by the `assertNull`, not by `employee.getId()` (indeed calling `getId` won't hit the database, so it wouldn't induce such an exception).

Really, shouldn't we confirm that the proxy is *not* null? And then call a method that would cause the exception? e.g.

```java
assertNotNull(employee);
employee.getFirstName();
```

PR #5375, opened 2018-10-02
BAEL-2000 - Shreyas Mahajan - Adding a project for demonstrating Spring Boot Jib Maven Plugin

shreyasm:


spring-boot-jib/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
jzheaux: So, there are some test dependencies here, but no tests.

First, do you think we are okay here without tests? The app is simple enough that I'm a bit impartial.

Second, if we aren't doing tests, let's drop the test dependencies.

In fact, let's make the pom as minimal as possible, just to keep the focus on the tutorial's topic.

spring-boot-jib/pom.xml
```diff
@@ -0,0 +1,68 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>spring-boot-jib</artifactId>
+    <version>0.1-SNAPSHOT</version>
+    <name>spring-boot-jib</name>
+    <packaging>war</packaging>
```
jzheaux: Why are we producing a war? Could we do a jar instead?

spring-boot-jib/checkstyle.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jzheaux: We have PMD already configured for the entire project, so unless this checkstyle is necessary for jib to work (e.g. jib will fail if we have any star imports), then let's remove it.

spring-boot-jib/.gitignore
```diff
@@ -0,0 +1,13 @@
+*.class
```
jzheaux: This seems unnecessary. I realize several modules have this, but we have a .gitignore at the top-level of the project. So, unless we are adding things here that are unique to jib (e.g. that we don't want to add to the main .gitignore), let's remove this file.

spring-boot-jib/src/main/resources/logback.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jzheaux: Is this file necessary for Spring Boot to work, or is it just a good practice? [The documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/howto-logging.html) seems to suggest that we don't need this.

If we don't need it, then lets remove it, just so the focus of the example is on jib.

spring-boot-jib/src/main/resources/application.properties
```diff
@@ -0,0 +1,2 @@
+server.port= 8086
+server.servlet.context-path=/spring-boot-jib
```
jzheaux: Is this custom context path necessary? e.g. I think without it, we'd just do localhost:8086/greeting and we'd be fine right?

spring-boot-jib/src/main/resources/application.properties
```diff
@@ -0,0 +1,2 @@
+server.port= 8086
```
jzheaux: It is important to set the port to 8086 instead of the default 8080? If not, let's remove it.

spring-boot-jib/src/main/java/com/baeldung/GreetingController.java
```diff
@@ -0,0 +1,20 @@
...
+
+import java.util.concurrent.atomic.AtomicLong;
+
+@RestController
+public class GreetingController {
+
+    private static final String template = "Hello Docker, %s!";
+    private final AtomicLong counter = new AtomicLong();
+
+    @RequestMapping("/greeting")
```
jzheaux: This should be `@GetMapping` instead. It's a bit simpler to read and it also doesn't expose methods unintentionally.

spring-boot-jib/pom.xml
shreyasm: Sure. I don't think we need tests here, since the focus is on configuration anyway. I will remove the test dependencies

spring-boot-jib/pom.xml
shreyasm: Yup, will modify it

spring-boot-jib/checkstyle.xml
shreyasm: Sure, its not necessary here. Will remove it

spring-boot-jib/.gitignore
shreyasm: Yes, nothing special to Jib, will remove it

spring-boot-jib/src/main/resources/logback.xml
shreyasm: Yes, its not needed. Anyways, any failure to build using Jib is going to be logged at console, so I will remove it. Thanks

spring-boot-jib/pom.xml
shreyasm: Removed the test dependencies

spring-boot-jib/pom.xml
shreyasm: Removed the packaging

spring-boot-jib/checkstyle.xml
shreyasm: Removed the file

spring-boot-jib/.gitignore
shreyasm: Removed the file

spring-boot-jib/src/main/resources/logback.xml
shreyasm: Removed the file

spring-boot-jib/src/main/resources/application.properties
shreyasm: Removed the file itself, since both properties are not needed. Actually, I added it so that the port and context are very independent of all other applications in source code. But, I guess nobody is going to deploy source code on a single machine, so we can delete it.

spring-boot-jib/src/main/resources/application.properties
shreyasm: Deleted the file. Reason behind adding it is mentioned above

PR #5379, opened 2018-10-02
BAEL - 2166

the-java-guy:
Code snippets for BAEL 2166


java-strings/src/test/java/com/baeldung/string/password/StringPasswordUnitTest.java
```diff
@@ -0,0 +1,198 @@
...
+import org.passay.CharacterData;
+import org.passay.CharacterRule;
+import org.passay.EnglishCharacterData;
+import org.passay.PasswordGenerator;
+
+/**
+ * Examples of passwords conforming to various specifications, using different libraries.
+ * 
+ * 
+ * @author tritty
```
KevinGilmore: Let's omit the author tag from the JavaDoc. Also, not sure we really need the JavaDoc on the test methods below. Better to let the method names speak for themselves.

the-java-guy: Sure thing. Updated the PR

PR #5393, opened 2018-10-06
BAEL-2167 Added classes for article

priyeshmashelkar:


xml/src/test/java/com/baeldung/xml/XMLDocumentWriterUnitTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.Test;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+
+public class XMLDocumentWriterUnitTest {
+
+    @Test
+    public void givenXMLDocumentWhenWriteIsCalledThenXMLIsWrittenToFile() throws Exception {
+        Document document = createSampleDocument();
+        new XMLDocumentWriter().write(document, "company_simple.xml", false, false);
```
jzheaux: My only concern here is that these tests will leave artifacts on the filesystem, right? Can we clean them up in an `@After` or similar?

xml/src/main/java/com/baeldung/xml/XMLDocumentWriter.java
```diff
@@ -0,0 +1,42 @@
...
+            if(prettyPrint) {
+                transformer.setOutputProperty(OutputKeys.INDENT, "yes");
+                transformer.setOutputProperty("{http://xml.apache.org/xslt}indent-amount", "2");
+            }
+            DOMSource source = new DOMSource(document);
+            FileWriter writer = new FileWriter(new File(fileName));
+            StreamResult result = new StreamResult(writer);
+            transformer.transform(source, result);
+        } catch (IOException e) {
+            e.printStackTrace();
```
jzheaux: Let's not encourage this practice. At least, let's do `throw new IllegalStateException(e)` or `throw new IllegalArgumentException(e)`, depending on whether we think it is a system error or user error.

xml/src/test/java/com/baeldung/xml/XMLDocumentWriterUnitTest.java
priyeshmashelkar: Added cleanUp method

xml/src/main/java/com/baeldung/xml/XMLDocumentWriter.java
priyeshmashelkar: Updated code as per suggestion

PR #5397, opened 2018-10-07
BAEL-2142 article code

codewarrior2000:
Two Java classes and one POM file update for BAEL-2142


java-strings/src/main/java/com/baeldung/string/WordIndexer.java
```diff
@@ -0,0 +1,46 @@
...
+            if (index != -1) {
+                indexes.add(index);
+            }else{
+                done = true;
+            }
+        }
+        return indexes;
+    }
+
+    public ArrayList findWordUpgrade(String textString, String word) {
```
KevinGilmore: The two methods are identical. Isn't the first one supposed to be the naive approach?

java-strings/src/main/java/com/baeldung/string/WordIndexer.java
```diff
@@ -0,0 +1,46 @@
...
+        int index = -1;
+        ArrayList<Integer> indexes = new ArrayList<Integer>();
+        StringBuilder output = new StringBuilder();
+        boolean done = false;
+        String lowerCaseTextString = textString.toLowerCase();
+        String lowerCaseWord = word.toLowerCase();
+        int wordLength = word.length();
+
+        while(!done){
+            index = lowerCaseTextString.indexOf(lowerCaseWord, index + wordLength);  // Slight improvement
```
KevinGilmore: What happens if the string ends with the word you're searching for? Would the next iteration of the loop cause an error if index + wordLength is greater than the last position of the string? Be sure to test this scenario.

codewarrior2000: Thank you, Kevin. They are all very good points. 
Java's indexOf() method gracefully responds to "... if the string ends with the word you're searching for" (i.e., no error should happen).  I added a JUnit test anyway to test that scenario.

codewarrior2000:
@KevinGilmore The merge conflict with java-strings/pom.xml is due to my addition of the following two dependencies for JUnit 5 however it could be that I don't have write access to this repository that I therefore cannot resolve the merge.

<dependency>
            <groupId>org.junit.jupiter</groupId>
            <artifactId>junit-jupiter-api</artifactId>
             <version>5.3.1</version>
             <scope>test</scope>
         </dependency>
 
         <dependency>
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest-library</artifactId>
             <version>1.3</version>
            <scope>test</scope>
        </dependency>

java-strings/src/main/java/com/baeldung/string/WordIndexerUnitTest.java
```diff
@@ -0,0 +1,45 @@
+package com.baeldung.string.searching;
```
KevinGilmore: Unit tests need to be in src/test/java, not in src/main/java

PR #5468, opened 2018-10-15
[BAEL-2255] Guide to BufferedReader

mprevisic:


core-java/src/main/java/com/baeldung/bufferedreader/BufferedReaderExample.java
```diff
@@ -0,0 +1,108 @@
...
+
+        while ((line = reader.readLine()) != null) {
+            content.append(line);
+            content.append(System.lineSeparator());
+        }
+
+        return content.toString();
+    }
+
+    public String readAllLines2(BufferedReader reader) {
```
jzheaux: Instead of `readAllLines2` could we use the method name to describe the different approach. e.g. `readAllLinesWithStreams`.

core-java/src/main/java/com/baeldung/bufferedreader/BufferedReaderExample.java
```diff
@@ -0,0 +1,108 @@
...
+
+        int value;
+        while ((value = reader.read()) != -1) {
+            content.append((char) value);
+        }
+
+        return content.toString();
+    }
+
+    public String readAllCharacters2(BufferedReader reader) throws IOException {
```
jzheaux: Same comment about the method name being more specific.

core-java/src/main/java/com/baeldung/bufferedreader/BufferedReaderExample.java
```diff
@@ -0,0 +1,108 @@
...
+        StringBuilder content = new StringBuilder();
+
+        String title = "text: ";
```
jzheaux: I'd be wary of adding additional logic in this method. The reason is that folks are going to want to compare `readAllCharacters` to `readAllCharacters2` and folks will probably see things like a title and line separators as noise.

core-java/src/main/java/com/baeldung/bufferedreader/BufferedReaderExample.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        return content.toString();
+    }
+
+    public void markAndReset(BufferedReader reader) throws IOException {
+        reader.mark(512);
+
+        for (int i = 0; i < 3; i++) {
+            System.out.println(reader.readLine());
```
jzheaux: Is there a way to do this without a sysout?

core-java/src/main/java/com/baeldung/bufferedreader/BufferedReaderExample.java
```diff
@@ -0,0 +1,108 @@
...
+            reader.reset();
+            reader.mark(512);
+        }
+    }
+
+    public void readFile() {
```
jzheaux: It would be great if this returned a value so that it can be verified in your unit test.

core-java/src/main/java/com/baeldung/bufferedreader/BufferedReaderExample.java
```diff
@@ -0,0 +1,108 @@
...
+                if (reader != null) {
+                    reader.close();
+                }
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+        }
+    }
+
+    public void readFileTryWithResources() {
```
jzheaux: Same comment about returning a value, if possible.

core-java/src/test/java/com/baeldung/bufferedreader/BufferedReaderExampleUnitTest.java
```diff
@@ -0,0 +1,122 @@
...
+            reader.reset();
+            result = reader.readLine();
+        }
+
+        assertEquals("Lorem ipsum dolor sit amet.", result);
+    }
+
+    @Test
+    public void whenCreatesNewBufferedReader_thenOk() throws IOException {
+        BufferedReader reader = Files.newBufferedReader(Paths.get(FILE_PATH));
```
jzheaux: Is there a reason this one isn't in a try-with-resources? e.g. 

```java
try ( BufferedReader reader = Files.newBufferedReader(Paths.get(FILE_PATH)) ) {
    assertNotNull(reader);
    assertTrue(reader.ready());
}
```

core-java/src/test/java/com/baeldung/bufferedreader/BufferedReaderExampleUnitTest.java
```diff
@@ -0,0 +1,122 @@
...
+public class BufferedReaderExampleUnitTest {
+
+    public static final String FILE_PATH = "src/main/resources/input.txt";
+
+    @Test
+    public void givenBufferedReader_whenReadAllLines_thenReturnsContent() throws IOException {
+        BufferedReader reader = new BufferedReader(new FileReader(FILE_PATH));
+
+        BufferedReaderExample bre = new BufferedReaderExample();
+        String content = bre.readAllLines(reader);
```
jzheaux: It feels a bit odd that some of these are delegated to `BufferedReaderExample` while others aren't. Is there a way that we could do one or the other? Or, is there a reason that it is this way?

mprevisic: I'm aware of this problem. Initially all tests were written against the BufferedReaderExample class but then, after I rewrote some example from BufReadExample in a unit-test-like style I moved them to BufReadExTest because they are now tests (although they aren't testing BufReadExample).. Would it be fine if a put them in a new BufferedReaderTest class in the same package? Or shall I rather puth them back into BufReadExample (than we will have a mix of normal example methods and unit-test methods)?

PR #5477, opened 2018-10-17
correct java main method signature, otherwise performance can't be ru…

shidongwa:
…n in IDE


lor6:
@shidongwa Thanks, merged.

PR #5517, opened 2018-10-22
BAEL-2236: Reading a CSV file into a array

PranayVJain:


core-java-io/src/test/java/com/baeldung/csv/ReadCSVInArrayUnitTest.java
```diff
@@ -0,0 +1,106 @@
...
+    public void givenCSVFile_whenBufferedReader_thenContentsAsExpected() throws IOException {
+        List<List<String>> records = new ArrayList<List<String>>();
+        try (BufferedReader br = new BufferedReader(new FileReader(CSV_FILE))) {
+            String line = "";
+            while ((line = br.readLine()) != null) {
+                String[] values = line.split(COMMA_DELIMITER);
+                records.add(Arrays.asList(values));
+            }
+        } catch (Exception e) {
+            e.printStackTrace();
```
jzheaux: I'm confused about why we want to proceed with the test if we've gotten an exception. Why not just allow it to be thrown?

core-java-io/src/test/java/com/baeldung/csv/ReadCSVInArrayUnitTest.java
```diff
@@ -0,0 +1,106 @@
...
+                    add("Kapil Isapuari");
+                    add("Farishta");
+                }
+            });
+        }
+    });
+
+    @Test
+    public void givenCSVFile_whenBufferedReader_thenContentsAsExpected() throws IOException {
+        List<List<String>> records = new ArrayList<List<String>>();
```
jzheaux: Can we remove the extra diamond? I think it will read more cleanly, and it's what developers are used to these days.

core-java-io/src/test/java/com/baeldung/csv/ReadCSVInArrayUnitTest.java
PranayVJain: Ok. I will make the changes.

core-java-io/src/test/java/com/baeldung/csv/ReadCSVInArrayUnitTest.java
PranayVJain: Agreed.

Issue #5497, opened 2018-10-19
UnzipFile is vulnerable to Zip Slip


Marcono1234:
The class [`UnzipFile`](https://github.com/eugenp/tutorials/blob/master/core-java-io/src/main/java/com/baeldung/unzip/UnzipFile.java) is affected by the [Zip Slip vulnerability](https://github.com/snyk/zip-slip-vulnerability) because it creates the target file path by concatenating the folder name with the zip entry name.

PR #5524, opened 2018-10-23
UnzipFile is vulnerable to Zip Slip #5497

eric-martin:


lor6:
@eric-martin  I don't think we need to push the zip file.

PR #5529, opened 2018-10-24
bael-2022 initialize array in kotlin

Vizsoro:


core-kotlin/src/test/kotlin/com/baeldung/kotlin/ArrayInitializationTest.kt
```diff
@@ -0,0 +1,48 @@
+package com.baeldung.kotlin
+
+import org.junit.Test
+import kotlin.test.assertEquals
+
+class ArrayInitializationTest {
+
```
echeyne: A few general notes:

1. Please review the code formatting guidelines within **baeldung - for authors - 2 - code in the article** and **baeldung - for authors - 2 - implementation code**. You need to indent your code with spaces instead of tabs and make the use of new lines within your tests consistent with one another and the overall code base.

2. semicolons should not be used in Kotlin -> https://stackoverflow.com/questions/39318457/what-are-the-rules-of-semicolon-inference-in-kotlin

Vizsoro: I got some problems with formatting because Eclipse Kotlin plugin (as far as I see) doesn't support formatting rules import from XML.  I have formatted the code as Java code. I hope it is ok now.
Also, I made the other requested changes.

core-kotlin/src/test/kotlin/com/baeldung/kotlin/ArrayInitializationTest.kt
```diff
@@ -0,0 +1,48 @@
...
+		for (i in array.indices) {
+			array[i] = i * i;
+		}
+
+		assertEquals(16, array[4])
+	}
+
+	@Test
+	fun whenGeneratorUsed_thenValuesPresent() {
+		val generatedArray = IntArray(10, { i -> i * i })
```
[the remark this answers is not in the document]
Vizsoro: Not for nitpicking but for learning, could you explain the difference between the two solutions highlighting why the second one is preferable?

echeyne:
Hey @Vizsoro I noticed that at one point (maybe your comment got deleted?) you asked about why the lambda should be outside of the parenthesis. I think that is a great question and I always encourage you to ask questions about my comments if something isn't clear.

According to the Kotlin's coding conventions if a lambda is the last parameter of a method it should go outside the parenthesis. Both ways of formatting will allow the code to compile, but we should try and follow the language's coding conventions whenever possible.

General coding conventions: https://kotlinlang.org/docs/reference/coding-conventions.html
Lambda specific: https://kotlinlang.org/docs/reference/lambdas.html#passing-a-lambda-to-the-last-parameter

core-kotlin/src/test/kotlin/com/baeldung/kotlin/ArrayInitializationTest.kt
echeyne: Changes look great, thanks!

echeyne: Your code changes look good. I will merge them in once we are ready to pass off your article to our senior editor.

PR #5541, opened 2018-10-26
BAEL-2250: Adding files for the article on SSL handshake failure.

kcacademic:
BAEL-2250: Adding files for the article on SSL handshake failure.


kcacademic:
Thanks, @echeyne, I usually have the formatter in my eclipse and use that before saving the file. I see that you have made changes to a class "SecureConnection" with respect to formatting but this has not been added as part of my original pull request. This is a class used by some other article.

Yes, the URISyntexException was a residue from some other experiment and I forgot to remove it.

As for the certificates, I agree we can get rid of them and expect the reader to generate as par the article, let me you if yo agree and I will remove them.

kcacademic:
Hi @echeyne, I have now reimported the formatter from https://github.com/eugenp/tutorials/blob/master/eclipse/formatter.xml and reapplied to my classes. 
I have removed unnecessary exceptions in the throws clause.
I have got rid of the generated certificates.

echeyne:
Hey @kcacademic, thanks for making those changes and thanks for alerting me to the fact that I changed `SecureConnection.java` unnecessarily, I have reverted the change. It looks like your code is good to go and I will merge it in once we are done going through your article.

PR #5557, opened 2018-10-28
[BAEL-2183] dupirefr/dupire.francois+pro@gmail.com Using String array for Stream section and adding array copy

dupirefr:


dupirefr:
This one wasn't supposed to be triggered as the article went live one or two weeks ago.

PR #5558, opened 2018-10-28
BAEL-2302 Spring Data REST HTTP Method customization example

gmconte:


gmconte:
Please don't consider this pull request as it was based on a very old version of the upstream codebase.
I created https://github.com/eugenp/tutorials/pull/5594 instead

PR #5559, opened 2018-10-28
BAEL-2218 Example of sorting in Kotlin

laurentiud:


core-kotlin/src/test/kotlin/com/baeldung/sorting/SortingExampleKtTest.kt
```diff
@@ -0,0 +1,14 @@
+package com.baeldung.sorting
+
+import org.junit.jupiter.api.Test
+
+import org.junit.jupiter.api.Assertions.*
+
+class SortingExampleKtTest {
+
+    @Test
+    fun naturalOrderComparator_ShouldBeAscendingTest() {
```
jzheaux: I think it would be helpful to readers and also more stable for the code if we had tests for each one of the methods in SortingExample.kt. I realize that this means that we may need to remove the println from those and turn them into asserts in the test, but, again, that's an improvement for stability and readability (e.g. I don't actually have to run the code to see what the output will be).

PR #5571, opened 2018-10-29
BAEL-2193 - Merge 2 java.util.Properties objects

JonCook:


JonCook:
Yes you are right, good catch. I've updated the PR. I think it was the automatic creation of the test methods in eclipse using a shortcut. Thanks for pointing that out. ;-)

PR #5583, opened 2018-10-30
use of static variables is concurrency problem

simbo1905:
Studying this tutorial code I noticed that it stores the current users request for a page of results in class static variables then reads those variables to render the page. If we get two concurrent requests we could have a second request clobber the first requests passed values and render the wrong page to the first user. 

If I try to make those values local variables the compiler says that "variable used in lamba expression should be final or effectively final". Intellij then suggests making them static, which is a concurrency bug, or atomic, and it performs the refactor to use local atomic variables which touches a lot of code to both declare, set and get the atomics. 

Using `orElse` to default the value is less code and does the job. Making them final also makes it clearer that they are parameters passed by the user and that allowing parameters to change in a method makes it harder to maintain code.


lor6:
Thanks @simbo1905  Merged.

PR #5621, opened 2018-11-05
FileOutputStream is never closed

holgerl:
The output file stream is not closed, so it is still in use by the Java process after the code runs. I noticed this when trying to delete the output files right afterwards.


lor6:
@holgerl Good point, thanks :) Merged.

PR #5630, opened 2018-11-06
Add BAEL-2272 persisting DDD aggregates examples

mikewojtyna:
Add persisting DDD aggregates examples (BAEL-2272). I created a completely new "ddd" module, which will contain all examples from the DDD series (BAEL-2271). I'm not really sure about the profiles in the parent module and the modules organization policy, so I simply used spring-boot parent. Let me know if that's acceptable.


echeyne:
Your code looks great, just one question; why do all of your unit tests throw an `Exception`?  It doesn't seem necessary?

echeyne:
Alright sounds good. I am going to merge your PR in and update the github link in the article

PR #5633, opened 2018-11-07
[BAEL-2322] Guide to Apache BookKeeper

smmansoor:


apache-bookkeeper/pom.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
```
echeyne: When we create a new `pom.xml` file we always include a reference to the parent module. Check out a module like `core-java` for an example

apache-bookkeeper/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+            </plugin>
+        </plugins>
+    </build>
+
+    <properties>
+        <!-- in your <properties> block -->
+        <bookkeeper.version>4.7.2</bookkeeper.version>
+    </properties>
+
+    <!-- in your <dependencies> block -->
```
echeyne: This comment should be removed

apache-bookkeeper/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+                <configuration>
+                    <source>8</source>
+                    <target>8</target>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+    <properties>
+        <!-- in your <properties> block -->
```
echeyne: This comment should be removed

apache-bookkeeper/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+        <bookkeeper.version>4.7.2</bookkeeper.version>
+    </properties>
+
+    <!-- in your <dependencies> block -->
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.bookkeeper</groupId>
+            <artifactId>bookkeeper-server</artifactId>
+            <version>${bookkeeper.version}</version>
```
echeyne: You never define `bookkeeper.version` anywhere

apache-bookkeeper/src/main/java/com/baeldung/BookkeeperLedgerExample.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.bookkeeper.client.api.BookKeeper;
+import org.apache.bookkeeper.client.api.DigestType;
+import org.apache.bookkeeper.client.api.LedgerEntries;
+import org.apache.bookkeeper.client.api.LedgerEntry;
+import org.apache.bookkeeper.client.api.ReadHandle;
+import org.apache.bookkeeper.client.api.WriteHandle;
+import org.apache.bookkeeper.conf.ClientConfiguration;
+
+
+public class BookkeeperLedgerExample {
```
echeyne: I think this class can be renamed to just `BookkeeperLedger`, we don't typically include `Example` in the class name because majority of our code is examples

apache-bookkeeper/src/main/java/com/baeldung/BookkeeperLedgerExample.java
```diff
@@ -0,0 +1,72 @@
...
+
+  public static void main(String[] args) throws Exception {
+
```
echeyne: I think a better way to format this code would be to create your objects and have some functional methods within `BookkeeperLedger` and then have some test code that verifies the contents of the objects. Typically, we format the code for our tutorials using that method. There are lots of examples of this pattern within the `core-java` module.

PR #5649, opened 2018-11-09
[BAEL-9549] reduce core-java module migrating code to core-java-lang and core-java-arrays

rozagerardo:
JIRA: http://team.baeldung.com/browse/BAEL-9549
Each commit will contain code to migrate around 15 articles' code from core-java to core-java-lang and core-java-arrays.
Each commit has a message indicating which articles it includes.
`mvn clean install` and `mvn clean install -P integration` commands ran before pushing, and I will contemplate if there are _*LiveTest.java_ classes.


core-java/src/main/java/org/baeldung/equalshashcode/entities/ComplexClass.java
```diff
@@ -2,7 +2,7 @@
```
rozagerardo: I will move org.baeldung.* code to com.baeldung.* in the new module, so that we have everything under the same base package

core-java/src/main/java/org/baeldung/equalshashcode/entities/ComplexClass.java
```diff
@@ -2,7 +2,7 @@
 
 import java.util.List;
 import java.util.Set;
-
+// all classes in package moved to core-java-lang as part of com.baeldung.* package
```
rozagerardo: As you can see here: https://github.com/eugenp/tutorials/pull/5649/files#diff-ff9a246bf080d2677d59ed9e474a87e9L1
The code for this particular article was duplicated (in com.baeldung). I will delete the code on both packages after successfully migrating the code to the new module (y)

rozagerardo:
Just some final notes:
* test pass successully in core-java-lang and core-java-arrays
* in core-java I have 4 failing tests, all seem to be platform-dependent tests (1 relies on the default locale, and the other 3 are making operation on an absolute path obtained programatically erasing the first character. Locally - with Linux - that change is not needed, the path is ok as it's provided)
* `mvn clean install` and `mvn clean install -P integration` finish succesfully in all modules
* 2 articles that I was planning to move to _core-java-lang_ are not part of _core-java_ module anymore:
https://www.baeldung.com/java-floating-point-sum-order (in here: https://github.com/eugenp/tutorials/blob/master/java-numbers/src/main/java/com/baeldung/maths/FloatingPointArithmetic.java)
https://www.baeldung.com/java-try-with-resources (in here: https://github.com/eugenp/tutorials/blob/master/core-java-8/src/test/java/com/baeldung/java8/JavaTryWithResourcesLongRunningUnitTest.java)
They were removed from the README file
* Some other articles were duplicated in the core-java README file, this was corrected
* functionality in org.baeldung.* was moved to com.baeldung.* when moved to other modules
* functionality in the root package com.baeldung was moved to a proper package when moved to a new module

PR #5655, opened 2018-11-10
BAEL-2239 - JPA error "java.lang.String cannot be cast to [Ljava.lang.String;"

asturcon:
IP: http://inprogress.baeldung.com/wp-admin/post.php?post=116693&action=edit


JavaCodeMood:
#5655

PR #5670, opened 2018-11-12
BAEL-2809 Get all Data in a table with Hibernate

enpy303:


persistence-modules/hibernate5/src/main/java/com/baeldung/hibernate/findall/FindAll.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private Session session;
+
+    public FindAll(Session session) {
+        super();
+        this.session = session;
+    }
+
+    public List<Student> findAllWithJpql() {
+        return session.createQuery("Select a From Student a", Student.class)
```
echeyne: Can you clean up the capitalization of this query? I believe the JPQL standard is to use the same conventions as SQL/MySQL, isn't it?

persistence-modules/hibernate5/src/test/java/com/baeldung/hibernate/findall/FindAllTest.java
```diff
@@ -0,0 +1,65 @@
...
+    private Session session;
+    private Transaction transaction;
+    
+    private FindAll findAll;
+
+    @Before
+    public void setUp() throws IOException {
+
+        session = HibernateUtil.getSessionFactory()
+            .openSession();
```
echeyne: In this case and on line 33 you can safely keep these fluent api calls on the same line. We only break up these calls when they are following out of the predefined page area, or if needed in our articles

echeyne: Same thing for `getResultList()` in `FindAll`

enpy303: Baeldung formatter breaks it down on save. I'll do it here, we need to update that formatter ...

persistence-modules/hibernate5/src/main/java/com/baeldung/hibernate/findall/FindAll.java
enpy303: JPQL doesn't have capitalization convention as far as i know. Some projects use uppercase, others lowercase etc. I guess we should define unique style for our tutorials. I uppercased both keywords ...

PR #5676, opened 2018-11-13
[BAEL-10301] core-java | fix unit tests

rozagerardo:
Fixed:
* SimpleDateFormatUnitTest test, which formerly relied on the system locale to format a date.
* UppercaseFileReaderUnitTest, LowercaseFileReaderUnitTest and StandardFileReaderUnitTest, which were using an approach to generate a resource path that seems to be efficient only on Windows; on Linux it raises an exception when loading the resource. The new approach should be good for all cases, but we need further testing on a Windows environment

### DIFF
--- a/core-java/src/test/java/com/baeldung/abstractclasses/LowercaseFileReaderUnitTest.java
+++ b/core-java/src/test/java/com/baeldung/abstractclasses/LowercaseFileReaderUnitTest.java
@@ -2,17 +2,22 @@ package com.baeldung.abstractclasses;
 
 import com.baeldung.abstractclasses.filereaders.BaseFileReader;
 import com.baeldung.abstractclasses.filereaders.LowercaseFileReader;
+
+import java.net.URL;
+import java.nio.file.Paths;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


alejandrogervasio:
@rozagerardo Hey Gerardo,

Thanks for the comments. Now I see why I was getting the errors in the unit tests, so thanks for making it clear. It just was an operating system issue related to a resource path. 

No worries. I'll checkout your branch early this morning and let you know how it goes :)

alejandrogervasio:
@rozagerardo Hey Gerardo,

I just checked out your branch and ran the core-java tests on my Windows machine.  All of them ran OK.

rozagerardo:
Awesome, thanks for the help and the quick response @alejandrogervasio , much appreciated :)

alejandrogervasio:
@rozagerardo Great! I'm constantly struggling to write better code, thus I didn't feel quite right with the unit test cases failing on Travis CI. But I thought the issue had to do with the way Travis was configured to get a resource path, and not with an operating system-related issue.

So thanks again for finding a working, cross-platform solution.

PR #5689, opened 2018-11-15
BAEL-2368 convert string array to string

gwidel:
Tests that show converting with different methods.


echeyne:
Can you put this in its own package within  the `com.baeldung` folder? Typically, we are trying to follow that folder structure

PR #5694, opened 2018-11-15
Bael 2330

kyleandari:
Removing repeated Characters from a string using various ways.


echeyne:
You also need to try running your code with Strings like `there are some mooose on the loose` and `geeks for geeks`. Right now not all of your test cases would pass. 

You are also missing a call to `removeDuplicatesUsingSorting` in your main method

PR #5702, opened 2018-11-16
BAEL-2325 Convert between String and Timestamp

laurentiud:


java-dates/src/test/java/com/baeldung/timestamp/TimestampToStringConverterTest.java
```diff
@@ -0,0 +1,18 @@
+package com.baeldung.timestamp;
+
+import org.junit.Assert;
+import org.junit.jupiter.api.Test;
+
+import java.sql.Timestamp;
+
+class TimestampToStringConverterTest {
```
jzheaux: Can we make this `public`?

java-dates/src/test/java/com/baeldung/timestamp/TimestampToStringConverterTest.java
```diff
@@ -0,0 +1,18 @@
...
+
+    @Test
+    public void testConversion() {
```
jzheaux: I mentioned this in my article review, but just to place it here, too, make sure that the method names adhere to the site standard of givenXXX_whenYYY_thenZZZ or whenYYY_thenZZZ

PR #5704, opened 2018-11-16
BAEL-2312 - Abstract Classes in Java

alejandrogervasio:


core-java/src/main/java/com/baeldung/abstractclasses/filereaders/BaseFileReader.java
```diff
@@ -1,19 +1,26 @@
...
     }
     
-    public String getFilePath() {
+    public Path getFilePath() {
         return filePath;
     }
     
-    public abstract List<String> readFile() throws IOException;
+    public List<String> readFile() throws IOException {
+        return Files.lines(filePath).collect(Collectors.toList());
```
jzheaux: @alejandrogervasio just because this is an intro article, I'm going to ask for just one more simplification. What would you think of this:

```java
return Files.lines(filePath).map(this::mapFileLine).collect(Collectors.toList());
```

And then

```java
protected abstract String mapFileLine(String line);
```

alejandrogervasio: @jzheaux Sound really good. I'll make the requested change today.

alejandrogervasio: @jzheaux I refactored the sample codes, so that they now reflect your suggestion. Definitively, this approach is better structured for a classic base abstract class/subclasses hierarchy. Nicely done!

PR #5724, opened 2018-11-18
BAEL - 1511

tinomthomas:
Added code related to the issue BAEL - 1511


rxjava-2/pom.xml
```diff
@@ -34,12 +34,20 @@
             <artifactId>rxrelay</artifactId>
             <version>${rxrelay.version}</version>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/com.github.akarnokd/rxjava2-extensions -->
+<dependency>
```
jzheaux: Can we fix the whitespacing here?

rxjava-2/src/test/java/com/baeldung/rxjava/AsyncAndSyncToObservableIntegrationTest.java
```diff
@@ -0,0 +1,103 @@
...
+            source.test()
+                .awaitDone(5, TimeUnit.SECONDS)
+                .assertResult(1);
+
+            assertEquals(1, counter.get());
+        }
+    }
+
+    @Test
+    public void givenAsyncMethod_whenConvertedWithFromFuture_thenRetrunObservble() { // method will execute only once and cache its result.
```
jzheaux: Let's shorten these long lines. Would recommend a /* */ comment above the method instead

PR #5748, opened 2018-11-21
Calculate Factorial in Java

dhananjay12:


algorithms-miscellaneous-1/src/test/java/com/baeldung/algorithms/factorial/FactorialUnitTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+    Factorial factorial;
+
+    @Before
+    public void setup() {
+        factorial = new Factorial();
+    }
+
+    @Test
+    public void givenNumber_whenCallingFactorialUsingForLoop_calcualeFactorial() {
```
KevinGilmore: The test names are missing the "then" component, and for most of them, the "given" isn't really necessary. This one, for example, could be "whenCalculateFactorialUsingForLoop_thenCorrect"

PR #5766, opened 2018-11-24
BAEL-2337: Find Substring which are palindromes

PranayVJain:


java-strings/src/main/java/com/baeldung/string/SubstringPalindrome.java
```diff
@@ -0,0 +1,90 @@
+package com.baeldung.string;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class SubstringPalindrome {
+
+    public Set<String> isPalindromeUsingCenter(String input) {
```
KevinGilmore: When a method name begins with "is", one expects it to return boolean. A better method name would be _findAllPalindromesUsingCenter_.

java-strings/src/main/java/com/baeldung/string/SubstringPalindrome.java
```diff
@@ -0,0 +1,90 @@
...
+        Set<String> result = new HashSet<>();
+        while (low >= 0 && high < input.length() && input.charAt(low) == input.charAt(high)) {
+            result.add(input.substring(low, high + 1));
+            low--;
+            high++;
+        }
+        return result;
+    }
+
+    public Set<String> isPalindromeUsingSubstring(String input) {
```
KevinGilmore: A better method name would be _findAllPalindromesUsingBruteForceApproach_

java-strings/src/main/java/com/baeldung/string/SubstringPalindrome.java
```diff
@@ -0,0 +1,90 @@
...
+        return palindromes;
+    }
+
+    private boolean isPalindrome(String input) {
+        StringBuilder plain = new StringBuilder(input);
+        StringBuilder reverse = plain.reverse();
+        return (reverse.toString()).equals(input);
+    }
+
+    public Set<String> isPalindromeUsingManachersAlgo(String input) {
```
KevinGilmore: Rename method: _findAllPalindromesUsingManachersAlgorithm_

java-strings/src/main/java/com/baeldung/string/SubstringPalindrome.java
```diff
@@ -0,0 +1,90 @@
...
+                    k++;
+                }
+                max = Math.max(max - k, 0);
+                i += k;
+            }
+        }
+        for (int i = 1; i <= input.length(); i++) {
+            for (int j = 0; j <= 1; j++) {
+                for (max = radius[j][i]; max > 0; max--) {
+                    palindromes.add(input.substring(i - max - 1, (2 * max + j) + (i - max - 1)));
```
KevinGilmore: The second expression could be simplified. Instead of "(2 * max + j) + (i - max - 1)",
we could write "max + j + i - 1"

java-strings/src/test/java/com/baeldung/string/SubstringPalindromeUnitTest.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.HashSet;
+import java.util.Set;
+import org.junit.Test;
+
+public class SubstringPalindromeUnitTest {
+
+    private static final String INPUT_NITIN = "nitin";
+    private static final String INPUT_CIVIC = "civic";
+    private static final String INPUT_LEVEL = "level";
+
```
KevinGilmore: These three inputs are pretty much the same: they're all five letters, the entire word is a palindrome, and the three letters in the middle form a palindrome. I don't see any reason to show all three. Choose one ("civic"). Then, let's include different kinds of inputs. Here are some examples of things to test to prove the algorithms are correct:
1. a string that contains one or more non-trivial palindromes, but is not itself a palindrome, such as "ababac"
2. a string that contains a palindrome of even length, such as "indeed" -- this also covers an edge case, where a string ends with a palindrome
3. a string that begins with a palindrome, such as "bubble"

java-strings/src/main/java/com/baeldung/string/SubstringPalindrome.java
PranayVJain: Changed named

java-strings/src/main/java/com/baeldung/string/SubstringPalindrome.java
PranayVJain: Changed name

java-strings/src/main/java/com/baeldung/string/SubstringPalindrome.java
PranayVJain: Changed name

java-strings/src/main/java/com/baeldung/string/SubstringPalindrome.java
PranayVJain: Expression simplified

java-strings/src/test/java/com/baeldung/string/SubstringPalindromeUnitTest.java
PranayVJain: Added different test cases

java-strings/src/test/java/com/baeldung/string/SubstringPalindromeUnitTest.java
```diff
@@ -0,0 +1,83 @@
...
+            add("aba");
+            add("bab");
+            add("ababa");
+        }
+    };
+
+    private SubstringPalindrome palindrome = new SubstringPalindrome();
+
+    @Test
+    public void whenApproachUsingManachersAlgo_shouldBePalindrome() {
```
KevinGilmore: These are pretty good names for tests, but let's strictly follow the BDD-style naming convention. For this one, it could be _whenUsingManachersAlgorithm_thenFindsAllPalindromes_. Use a similar name for the other tests.

KevinGilmore:
```suggestion
    public void whenUsingManachersAlgorithm_thenFindsAllPalindromes() {
```

java-strings/src/test/java/com/baeldung/string/SubstringPalindromeUnitTest.java
```diff
@@ -0,0 +1,83 @@
...
+        assertEquals(EXPECTED_PALINDROME_BUBBLE, palindrome.findAllPalindromesUsingManachersAlgorithm(INPUT_BUBBLE));
+        assertEquals(EXPECTED_PALINDROME_INDEED, palindrome.findAllPalindromesUsingManachersAlgorithm(INPUT_INDEED));
+        assertEquals(EXPECTED_PALINDROME_CIVIC, palindrome.findAllPalindromesUsingManachersAlgorithm(INPUT_CIVIC));
+        assertEquals(EXPECTED_PALINDROME_ABABAC, palindrome.findAllPalindromesUsingManachersAlgorithm(INPUT_ABABAC));
+    }
+
+    @Test
+    public void whenApproachUsingCenter_shouldBePalindrome() {
```
KevinGilmore:
```suggestion
    public void whenUsingCenterApproach_thenFindsAllPalindromes() {
```

java-strings/src/test/java/com/baeldung/string/SubstringPalindromeUnitTest.java
```diff
@@ -0,0 +1,83 @@
...
+        assertEquals(EXPECTED_PALINDROME_BUBBLE, palindrome.findAllPalindromesUsingCenter(INPUT_BUBBLE));
+        assertEquals(EXPECTED_PALINDROME_INDEED, palindrome.findAllPalindromesUsingCenter(INPUT_INDEED));
+        assertEquals(EXPECTED_PALINDROME_CIVIC, palindrome.findAllPalindromesUsingCenter(INPUT_CIVIC));
+        assertEquals(EXPECTED_PALINDROME_ABABAC, palindrome.findAllPalindromesUsingCenter(INPUT_ABABAC));
+    }
+
+    @Test
+    public void whenApproachUsingBruteForce_shouldBePalindrome() {
```
KevinGilmore:
```suggestion
    public void whenUsingBruteForceApproach_thenFindsAllPalindromes() {
```

java-strings/src/test/java/com/baeldung/string/SubstringPalindromeUnitTest.java
PranayVJain: Ok sure

PranayVJain: Changed test case name

java-strings/src/test/java/com/baeldung/string/SubstringPalindromeUnitTest.java
PranayVJain: Changed test case name

java-strings/src/test/java/com/baeldung/string/SubstringPalindromeUnitTest.java
PranayVJain: Changed test case name to whenUsingBruteForceApproach_thenFindsAllPalindromes

PR #5775, opened 2018-11-25
BAEL-2213 Added custom dialect for SQLite database.

sandy03934:


spring-data-rest/pom.xml
```diff
@@ -33,6 +33,15 @@
             <groupId>com.h2database</groupId>
             <artifactId>h2</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-autoconfigure</artifactId>
+        </dependency>
+        <!--<dependency>-->
```
jzheaux: Commenting this out confuses me. Can you explain?

spring-data-rest/pom.xml
```diff
@@ -43,6 +52,8 @@
     </build>
 
     <properties>
+        <start-class>com.baeldung.SpringDataRestApplication</start-class>
```
jzheaux: Does this break other demos in the same module?

sandy03934: No it doesn't. The Spring-boot Maven plugin in parent-boot-2 module expects the start-class property as well. So added the value

spring-data-rest/pom.xml
sandy03934: The project is by default working on H2 Database. So leaving this commented. We can uncomment if we want to use the SQLite database.

sandy03934:
Added some comments to reflect the reason behind the pom changes

jzheaux:
Closed in favor of #5822

PR #5786, opened 2018-11-26
There is no PasswordEncoder mapped for the id "null"

mlizbeth:
I was reading this tutorial https://www.baeldung.com/spring-boot-security-autoconfiguration
and spring boot will crash if you don't specify {noop} in the password since the tutorial is using in memory auth & there is no encoder present.

Fixes: java.lang.IllegalArgumentException: There is no PasswordEncoder mapped for the id "null"
	at org.springframework.security.crypto.password.DelegatingPasswordEncoder$UnmappedIdPasswordEncoder.matches(DelegatingPasswordEncoder.java:244) ~[spring-security-core-5.1.1.RELEASE.jar:5.1.1.RELEASE]


lor6:
Hey @mlizbeth Thanks for the note. The code in this module is still using Spring Boot 1. We do have a plan to migrate to Boot 2, and we'll add a PasswordEncoder bean at that point.

PR #5791, opened 2018-11-27
BAEL-2360

MajewskiKrzysztof:


spring-boot-mvc/src/main/java/com/baeldung/shedlock/ShedlockScheduler
```diff
@@ -0,0 +1,15 @@
+package com.baeldung.shedlock;
```
jzheaux: Why isn't this a Java file?

spring-boot-mvc/pom.xml
```diff
@@ -72,6 +72,18 @@
             <version>${spring.fox.version}</version>
         </dependency>
 
+        <!-- ShedLock -->
```
jzheaux: It feels weird that this is in `spring-boot-mvc`. Is there a more general one? My concern is this has nothing to do with MVC.

spring-boot-mvc/src/main/java/com/baeldung/shedlock/ShedlockScheduler
jzheaux: Also, can we make sure it matches the article?

MajewskiKrzysztof:
@jzheaux Sorry for that mistake :/
I made changes. I also deleted Application class and created new one: SchedulerConfiguration. I moved all to spring-all project. 
Is it ok now?

PR #5805, opened 2018-11-29
BAEL-2174 How to use proxies in core java

chrisoberle:
* move samples from core-java to new core-java-net module


core-java-net/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
KevinGilmore: Is this an IDE-specific file, perhaps from Visual Studio? If so, then let's omit it from the repo.

chrisoberle: removed and rebase squash

PR #5903, opened 2018-12-12
[BAEL-10845] | Upgrade set of modules to use boot 2.1 - part 3

rozagerardo:
JIRA: http://team.baeldung.com/browse/BAEL-10845
Notes & observations:

1- **spring-rest-template**
It seems that this module was already migrated to the spring-rest module. but only one class still remains here (it's also in spring-rest, it seems they just forgot to delete this). Maybe we have to remove this module
https://github.com/eugenp/tutorials/blob/master/spring-rest-template/src/main/java/com/baeldung/web/upload/client/MultipartFileUploadClient.java
https://github.com/eugenp/tutorials/blob/master/spring-rest/src/main/java/com/baeldung/web/upload/client/MultipartFileUploadClient.java

2- **spring-resttemplate**
The app starts, but for some reason it finishes right away, as if it wasn't a web-app. LiveTests of course don't work
We should revisit the project structure (I believe the main class is not scanning all packages) and the pom file (its really complex, there are lots of dependencies and configurations that could probably be removed)

3-  **spring-security-mvc-boot**
SpringDataWithSecurityUnitTest seems to be an Integration test actually.

4- **spring-vault**
All Integration tests (VaultIntegrationTest and SpringContextIntegrationTest) throw exception. It is not related to this change, it seems to be trying to run a 'vault' command to start a vault server in the host machine. Of course it needs the environment to be configured for this, I would rename them it as a 'ManualTest', or change the approach of the test.

Even with the command available, 2 out of 3 tests fail (not related to the spring-boot version upgrade, I checked using the former version, and tests still failed).

5- **spring-webflux-amqp**
SpringContextIntegrationTest requires a RabbitMQ running insttance to pass. Maybe we want to rename it as LiveTest or ManualTest?

Also, I changed the spring.rabbitmq.host property. It had a hard IP value, now its using 'localhost' so that tests pass successfuly. The reason why I mention this is because in the related article it also has the fixed IP. I didn't read the whole article, but that IP is not mentioned when setting up the rabbitMQ server, so we might want to analize the possibility of changing it to 'localhost' as well

6- **vaadin**
I'm having an error on this module when bootstraping this app, not related to this change (I tried with the 2.0.x version too):
```
[ERROR] Failed to execute goal org.springframework.boot:spring-boot-maven-plugin:2.0.5.RELEASE:run (default-cli) on project vaadin: An exception occurred while running. null: InvocationTargetException: Unable to start web server; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'tomcatServletWebServerFactory' defined in class path resource [org/springframework/boot/autoconfigure/web/servlet/ServletWebServerFactoryConfiguration$EmbeddedTomcat.class]: Initialization of bean failed; nested exception is org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'org.springframework.boot.autoconfigure.web.servlet.error.ErrorMvcAutoConfiguration': Unsatisfied dependency expressed through constructor parameter 1; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.boot.autoconfigure.web.servlet.DispatcherServletPath' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {} -> [Help 1]
```
Also, it seems spring 2.1 is supported from version 8 onwards (we're using v7):
https://github.com/vaadin/spring/issues/331#issuecomment-435128475
The APIs change quite a lot in that version, the process (and probably the related article) will need quite a lot of refactoring if we want to upgrade to the latest versions (it seems V10 was a big milestone).

7- **vavr**
This has a vavr-validation folder in the root directory, which also has a maven project structure, but of course, it is not an actual sub-module, it has no pom file, and it doesn't seem very useful as it is. Maybe we should make this a module with 2 submodules?


vaadin/src/main/resources/application.properties
```diff
@@ -0,0 +1,2 @@
+#Vaadin supports spring-boot 2.1 properly from V8 onwards (according to this comment https://github.com/vaadin/spring/issues/331#issuecomment-435128475)
+spring.main.allow-bean-definition-overriding=true
```
rozagerardo: As I mentioned in the notes, vaadin supports spring 2.1 properly from version 8 onwards, that's why I decided to 'bypass' a bean overriding exception (caused by spring and vaadin autoconfigurations, nothing that we can handle) using this property

PR #5908, opened 2018-12-13
BAEL-2068: Using hibernate-c3p0

eric-martin:


persistence-modules/hibernate5/pom.xml
```diff
@@ -56,6 +61,11 @@
             <artifactId>jackson-databind</artifactId>
             <version>${jackson.version}</version>
         </dependency>
+        <dependency>
```
eric-martin: Unrelated to this PR but required for tests to run

PR #5954, opened 2018-12-19
feature/hexagonal architecture by davecranesunwheel@gmail.com

DaveTheCrane:
This is a new spring boot project used to demonstrate refactoring from layered to hexagonal architecture, and to provide the sample code for the article 

http://inprogress.baeldung.com/?p=124488&preview=true

I had a look in the repo for an existing module to use, but couldn't find anything quite right. It was created in Spring Initializr, and I haven't tried to add this into the parent pom as a module. Let me know if you want to do this.

The maven wrappers have gone into git too - should they be there, or should I excise them?

There are three commits beyond the initial app, representing three stages of refactoring. Code in the article borrows from all three commits.

Regards, Dave Crane


DaveTheCrane:
Closing immediately after opening, as requested in email

PR #5959, opened 2018-12-20
[BAEL-2270] Guide to XMPP Smack Client

freddyaott:


libraries/src/main/java/com/baeldung/smack/StanzaThread.java
```diff
@@ -0,0 +1,57 @@
...
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smack.tcp.XMPPTCPConnection;
+import org.jivesoftware.smack.tcp.XMPPTCPConnectionConfiguration;
+import org.jxmpp.jid.impl.JidCreate;
+import org.jxmpp.stringprep.XmppStringprepException;
+
+import java.io.IOException;
+
+public class StanzaThread extends Thread{
+
```
jzheaux: Let's clean up the whitespacing here.

libraries/src/main/java/com/baeldung/smack/StanzaThread.java
```diff
@@ -0,0 +1,57 @@
...
+import org.jivesoftware.smack.chat2.ChatManager;
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smack.tcp.XMPPTCPConnection;
+import org.jivesoftware.smack.tcp.XMPPTCPConnectionConfiguration;
+import org.jxmpp.jid.impl.JidCreate;
+import org.jxmpp.stringprep.XmppStringprepException;
+
+import java.io.IOException;
+
+public class StanzaThread extends Thread{
```
jzheaux: Extending `Thread` is an antipattern - is it necessary in this case? If not, let's change this to implement `Runnable`.

jzheaux: Also, let's be consistent with our whitespacing on the curly at the end of lines. Sometimes you do it without a space between the final token and the brace, e.g. `Thread{` and other times you have a space, e.g. `run() {`. Let's use a space.

libraries/src/main/java/com/baeldung/smack/StanzaThread.java
```diff
@@ -0,0 +1,57 @@
...
+
+
+            ChatManager chatManager = ChatManager.getInstanceFor(connection);
+
+            Chat chat = chatManager.chatWith(JidCreate.from("baeldung@jabb3r.org").asEntityBareJidOrThrow());
+
+            chat.send("Hello!");
+
+        } catch (XmppStringprepException e) {
+            e.printStackTrace();
```
jzheaux: Let's not print out the stack trace. Instead I recommend either logging it or rethrowing an `IllegalArgumentException` if it is the user's fault (bad input, etc.) or an `IllegalStateException` if it is the environment's fault.

libraries/src/test/java/com/baeldung/smack/SmackIntegrationTest.java
```diff
@@ -0,0 +1,92 @@
...
+                .build();
+
+        connection = new XMPPTCPConnection(config);
+        connection.connect();
+        connection.login();
+
+    }
+
+    @Test
+    public void whenSendMessageWithChatThenReceiveMessage() throws XmppStringprepException, InterruptedException {
```
jzheaux: Thank you for using the when/then convention. This is easy to miss, but would you please do it with an underscore, like whenXXX_thenYYY, please?

libraries/src/main/java/com/baeldung/smack/StanzaThread.java
```diff
@@ -0,0 +1,57 @@
...
+import org.jivesoftware.smack.tcp.XMPPTCPConnectionConfiguration;
+import org.jxmpp.jid.impl.JidCreate;
+import org.jxmpp.stringprep.XmppStringprepException;
+
+import java.io.IOException;
+
+public class StanzaThread extends Thread{
+
+
+    public StanzaThread(){
```
jzheaux: Let's remove this, if possible.

libraries/src/test/java/com/baeldung/smack/SmackIntegrationTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+        Object cv = new Object();
+        ChatManager chatManager = ChatManager.getInstanceFor(connection);
+        final String[] expected = {null};
+
+        new StanzaThread().start();
+
+
+        chatManager.addIncomingListener((entityBareJid, message, chat) -> {
+            System.out.println("Message arrived: " + message.getBody());
```
jzheaux: It would be better if this were a logger. Is that possible here? I'm guessing the module has `slf4j`.

libraries/src/test/java/com/baeldung/smack/SmackIntegrationTest.java
```diff
@@ -0,0 +1,92 @@
...
+            expected[0] = message.getBody();
+            synchronized (cv) {
+                cv.notify();
```
jzheaux: Huh, this is clever. :) I've also seen people use a `CountDownLatch`, which you might find to have less boilerplate in the end. (No need to change, just something I've found to be effective.)

diegomoreira001:
Build looks good now. 
CountDownLatch was a nice touch. Thanks for pointing that out, Josh.

PR #6002, opened 2018-12-26
BAEL-2444 Sample code

earth001:


earth001:
I used the 2.0.0-RC.4 JMockit version because its the only version that works with lastest Mockito version.

PR #6035, opened 2018-12-31
username - rajeshbhojwani "Java toString() method" 

RajeshBhojwani:
This pull request is to merge code for "Java toString() method" Article


core-java/src/main/java/com/baeldung/tostring/CustomerS5.java
```diff
@@ -0,0 +1,49 @@
+package com.baeldung.tostring;
+
+import java.util.List;
+
+import org.apache.commons.lang.builder.ReflectionToStringBuilder;
+
+public class CustomerS5  {
+    private String firstName;
```
echeyne: There is a lot of duplicate code across all of these customer classes. Can't we have on parent `Customer` class and then extend it for all of these other classes? Also, lets name these classes a little bit better to describe the differences of the `toString` methods in each (i.e. this class could be called `CustomerReflectionToString`)

echeyne:
I think the more approprite module for this is https://github.com/eugenp/tutorials/tree/master/java-strings/src/main/java/com/baeldung/string. We are trying to split out `java-core` a little bit because it is quite full at the moment. You can create a `tostring` package within the above module.

core-java/src/main/java/com/baeldung/tostring/CustomerS1.java
```diff
@@ -0,0 +1,32 @@
+package com.baeldung.tostring;
+
+public class CustomerS1  {
+	private String firstName;
```
echeyne: We have an Eclipse formatter that you can find in our author documentation that we ask all of our authors run their code through prior to creating a PR. Can you apply it to your code changes? It will fix up random spaces like this one.

RajeshBhojwani:
Closing this pull request as I had a repository corruption issue. Created a new repository and raising new pull request.

RajeshBhojwani:
I have accommodated all the feedback in the new pull request.

PR #6071, opened 2019-01-04
New section in InputStream to byte array article and recreating branch.

eaudet:
@echeyne I had to delete the first branch since it contained commits from the first article written during my onboarding evaluation. I started fresh with a new fork from the base. 

Here's your last comment on my old pull request:

> Hey @eaudet it seems like there are a lot of files in this PR that aren't related to your changes. Can you remove them, please? We only want to have files changed that are relevant to your article. Formatting changes to modules, in general, are fine, but we want them to be in a separate pull request.


echeyne:
Thanks for making that change @eaudet :)

Can you run your code through our formatter (you can find it in the author docs) the indentation seems to be off

eaudet:
My test will always work. I will modify the code accordingly.

eaudet:
@echeyne do you have other requests/comments on this code? Can you proceed?

echeyne:
I am not sure that `core-java-8` is the right place for this. What about `core-java-io` instead?

eaudet:
@echeyne since cicd passed all check, the error was only on Mac OSx Oracle JDK. So I would do anything about this. ;-)

PR #6080, opened 2019-01-05
BAEL-2500 || Changes to enable TLSv1.2 on Java 7.

svkirans:
How to enable TLS 1.2 in Java 7

### DIFF
--- a/core-java-security/src/main/java/com/baeldung/ssl/EnableTLSv12.java
+++ b/core-java-security/src/main/java/com/baeldung/ssl/EnableTLSv12.java
@@ -7,6 +7,9 @@ import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.net.URL;
+import java.net.UnknownHostException;
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
 
 import javax.net.ssl.HttpsURLConnection;
@@ -15,7 +18,12 @@ import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLSocket;
 import javax.net.ssl.SSLSocketFactory;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class EnableTLSv12 {
+
+    private final Logger logger = LoggerFactory.getLogger(EnableTLSv12.class);
 
     public String url = "";
     public Integer port = null;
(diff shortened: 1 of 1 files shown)


core-java-security/src/main/java/com/baeldung/ssl/EnableTLSv12.java
```diff
@@ -0,0 +1,134 @@
...
+                System.exit(-1);
+            }
+            enableTLSv12.getHost(args);
+            enableTLSv12.getPort(args);
+            enableTLSv12.enableTLSv12UsingHttpConnection();
+            enableTLSv12.enableTLSv12UsingProtocol();
+            enableTLSv12.enableTLSv12UsingSSLContext();
+            enableTLSv12.enableTLSv12UsingSSLParameters();
+        } catch (Exception e) {
+            e.printStackTrace();
```
jzheaux: If we just want to print the stack trace, then go ahead and change the `main` method signature to throw the appropriate exceptions.

core-java-security/src/main/java/com/baeldung/ssl/EnableTLSv12.java
```diff
@@ -0,0 +1,134 @@
...
+        url = args[0];
+    }
+
+    private void getHost(String[] args) {
+        String portNumber = args[1];
+        port = Integer.parseInt(portNumber);
+    }
+
+    private void handleCommunication(SSLSocket socket, String usedTLSProcess) throws IOException {
+        System.out.print("Enabled TLS v1.2 on " + usedTLSProcess);
```
jzheaux: Can you use a logger instead or remove this (as well as other locations that use a sysout)?

core-java-security/src/main/java/com/baeldung/ssl/EnableTLSv12.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public void enableTLSv12UsingProtocol() throws IOException {
+        try {
+            SSLSocketFactory socketFac = (SSLSocketFactory) SSLSocketFactory.getDefault();
+            SSLSocket sslSocket = (SSLSocket) socketFac.createSocket(url.trim(), port);
+            sslSocket.setEnabledProtocols(new String[] { "TLSv1.2" });
+            sslSocket.startHandshake();
+            handleCommunication(sslSocket, "SSLSocketFactory-EnabledProtocols");
+        } catch (Exception e) {
+            e.printStackTrace();
```
jzheaux: Let's either throw the exception the signature or wrap this exception in something like an `IllegalStateException`. Same for the other places where you have an `e.printStackTrace`

core-java-security/src/main/java/com/baeldung/ssl/EnableTLSv12.java
```diff
@@ -0,0 +1,134 @@
...
+        if (out.checkError())
+            System.out.println("SSLSocketClient:  java.io.PrintWriter error");
+
+        BufferedReader in = new BufferedReader(new InputStreamReader(socket.getInputStream()));
+
+        String inputLine;
+        while ((inputLine = in.readLine()) != null)
+            System.out.println(inputLine);
+
+        in.close();
```
jzheaux: These won't get closed in the case of an exception, right? We want to try and show good engineering practice, even though closing streams isn't so important in an example like this.

`try-with-resources` is available in Java 7, so I think these could be cleaned up nicely.

core-java-security/src/main/java/com/baeldung/ssl/EnableTLSv12.java
```diff
@@ -0,0 +1,134 @@
...
+        port = Integer.parseInt(portNumber);
+    }
+
+    private void handleCommunication(SSLSocket socket, String usedTLSProcess) throws IOException {
+        System.out.print("Enabled TLS v1.2 on " + usedTLSProcess);
+        PrintWriter out = new PrintWriter(new BufferedWriter(new OutputStreamWriter(socket.getOutputStream())));
+        out.println("GET / HTTP/1.0");
+        out.println();
+        out.flush();
+        if (out.checkError())
```
jzheaux: I'm not completely clear on why, if there is indeed an error, we still continue on with the method?

core-java-security/src/main/java/com/baeldung/ssl/EnableTLSv12.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    public static void main(String[] args) {
+        try {
+            EnableTLSv12 enableTLSv12 = new EnableTLSv12();
+            if (args.length != 2) {
+                System.out.println("Provide the server url and the secure port:");
+                System.exit(-1);
+            }
+            enableTLSv12.getHost(args);
```
jzheaux: Wouldn't it be more common to call these `setHost` and `setPort`?

jzheaux: Or, if it's easy, you might consider something like MockWebServer so that you can control the output and convert these methods to tests. (Up to you)

PR #6086, opened 2019-01-06
Check if a year/date is a leap year in Java

cupendra:
@pivovarit I have added a small program with assertions for checking leap year. Please check.


cupendra:
Thanks, @pivovarit.

PR #6116, opened 2019-01-09
BAEL-2559: Creating a custom annotation in Java

ChitranshSaurabh18:


core-java/src/main/java/com/baeldung/annotations/AnnotationParser.java
```diff
@@ -0,0 +1,85 @@
...
+            method.setAccessible(true);
+            if (method.isAnnotationPresent(MethodDescription.class)) {
+                MethodDescription annotation = method.getAnnotation(MethodDescription.class);
+                System.out.println(index + ") " + "Description: " + annotation.value() + ", Author: " + annotation.author() + ", Version: " + annotation.version());
+                index++;
+            }
+        }
+    }
+
+    public static void main(String[] args) throws Exception {
```
echeyne: Instead of a main class let's create a unit test for this

echeyne:
Hey @ChitranshSaurabh18,

Your code looks good :) The reason that the build is failing is because you aren't following our naming convention for unit tests. Your test class names should end with `UnitTest` so `MethodDescriptionParserTest` needs to be `MethodDescriptionParserUnitTest`. Can you rename your two unit tests?

Also, we follow the BDD test naming conventions. Can you please rename your tests accordingly?

More info on both of these issues can be found in "baeldung - for authors - 2 - implementation code" and "baeldung - for authors - 2 - code in the article".

Thanks,
Emily

echeyne:
@ChitranshSaurabh18 the unit test names still aren't correct. Can you please take a look at the documentation that I referenced in my above comment as well as in my review video.

PR #6123, opened 2019-01-10
BAEL-2447: Added unit test cases to list files within directory

PranayVJain:


core-java-io/src/test/resources/listFilesUnitTestFolder/country.txt
```diff
@@ -0,0 +1 @@
+<xml></xml>
```
jzheaux: This appears to be an xml file, not a txt file. Can we adjust one or the other please?

core-java-io/src/test/resources/listFilesUnitTestFolder/employe.json
```diff
@@ -0,0 +1 @@
+{}
```
jzheaux: Let's fix the spelling on employee

core-java-io/src/test/resources/listFilesUnitTestFolder/country.txt
PranayVJain: Ok done

PR #6189, opened 2019-01-21
[BAEL-11598] Update and move code for the "REST Pagination in Spring" article

rozagerardo:
moved Rest Pagination article code from spring-rest-full to spring-boot-rest

JIRA: http://team.baeldung.com/browse/BAEL-11598

### DIFF
--- a/spring-boot-rest/src/test/java/com/baeldung/common/web/AbstractBasicLiveTest.java
+++ b/spring-boot-rest/src/test/java/com/baeldung/common/web/AbstractBasicLiveTest.java
@@ -1,0 +1,103 @@
+package com.baeldung.common.web;
+
+import static com.baeldung.web.util.HTTPLinkHeaderUtil.extractURIByRel;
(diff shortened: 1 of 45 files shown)


spring-boot-rest/src/test/java/com/baeldung/common/web/AbstractBasicLiveTest.java
```diff
@@ -0,0 +1,103 @@
...
+        final Response response = RestAssured.get(getURL() + "?page=0&size=10");
+
+        assertFalse(response.body().as(List.class).isEmpty());
+    }
+
+    @Test
+    public void whenFirstPageOfResourcesAreRetrieved_thenSecondPageIsNext() {
+        create();
+        create();
+        create();
```
rozagerardo: Some of these tests changed a little bit, creating o couple of entities before making the assertions.
In the old module, I think persistence tests were carried out as part of the test suite before running these tests, therefore they didn't need to be created explicitly. Now each live test should run successfully even if they are run alone.

PR #6218, opened 2019-01-26
BAEL-2466 Spring soap

alv21:
http://jira.baeldung.com/browse/BAEL-2466


spring-soap/src/main/java/com/baeldung/springsoap/gen/Country.java
```diff
@@ -0,0 +1,144 @@
+//
+// Questo file è stato generato dall'architettura JavaTM per XML Binding (JAXB) Reference Implementation, v2.2.7 
+// Vedere <a href="http://java.sun.com/xml/jaxb">http://java.sun.com/xml/jaxb</a> 
+// Qualsiasi modifica a questo file andrà persa durante la ricompilazione dello schema di origine. 
+// Generato il: 2019.01.25 alle 06:06:58 PM PST 
+//
+
+
+package com.baeldung.springsoap.gen;
```
KevinGilmore: Don't we simply want these to be generated? Why include them in the PR?

alv21: If I was just reading the article and I want to check the project out, without actually download it and run it, I would like to see the generated classes.
Anyway ff you think it's not necessary I can delete them, definitely I'm going to change the comment in Italian, I don't know why my jdk wrote them in Italian.

eric-martin: It looks like you have deleted the generated code.

PR #6225, opened 2019-01-27
Added tutorial sample for BAEL-2550

er-han:


thombergs:
reopening, since this has still to be merged

thombergs:
@er-han do we need the BarBean and FooBarBean classes? They are not referenced in the article

er-han:
@thombergs 
I thought three different component types would be better to demonstrate field, setter and constructor injection in one abstract class. The other option was one component class vs three abstract class. I could still use just one component type for three kinds of injection in one abstract class. But I think that wouldn't be a good code example.
At first I referenced the BarBean and FooBarBean classes in the article, too. I changed them with FooBean later, to clarify the article. I think the reader could be confused to see different component types for each example. So, code in github and in the article is not exactly the same now.

PR #6234, opened 2019-01-27
BAEL-774

the-java-guy:
Files corresponding to Ratpack with RxJava (BAEL-774)


pom.xml
```diff
@@ -9,6 +9,7 @@
 	<version>1.0.0-SNAPSHOT</version>
     <modules>
         <module>lombok-custom</module>
+        <module>reactive-modules</module>
```
jzheaux: Looks like lombok custom is in the wrong place. :) You'll see a much longer list of modules later on in the pom. Please place `reactive-modules` there.

ratpack/src/main/java/com/baeldung/model/Movie.java
```diff
@@ -0,0 +1,47 @@
+package com.baeldung.model;
```
jzheaux: Just thinking ahead a bit, would you please put these in something like `com.baeldung.ratpack.intro.model` just so that if we do other articles on ratpack, then other authors can get the namespace isolation they need?

ratpack/src/main/java/com/baeldung/model/Movie.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * 
+ *POJO class for Movie object
+ */
+public class Movie {
```
jzheaux: Also, do we need both `Movie` classes? (I'm hoping not!)

reactive-modules/ratpack-rxjava/src/main/java/com/baeldung/rxjava/RatpackErrorHandlingApp.java
```diff
@@ -0,0 +1,21 @@
+package com.baeldung.rxjava;
+
+import ratpack.error.ServerErrorHandler;
+import ratpack.rx.RxRatpack;
+import ratpack.server.RatpackServer;
+import rx.Observable;
+
+public class RatpackErrorHandlingApp {
```
jzheaux: Since this is an app that the user can run, can we place a class-level comment stating what endpoint they should hit to see the behavior in action? e.g.:

```java
/**
  * Try hitting http://localhost:8080/some-url to see the error handler in action
  */
```

reactive-modules/ratpack-rxjava/src/main/java/com/baeldung/rxjava/RatpackObserveApp.java
```diff
@@ -0,0 +1,39 @@
...
+import com.baeldung.rxjava.service.MoviePromiseService;
+import com.baeldung.rxjava.service.impl.MoviePromiseServiceImpl;
+
+import ratpack.exec.Promise;
+import ratpack.handling.Handler;
+import ratpack.jackson.Jackson;
+import ratpack.rx.RxRatpack;
+import ratpack.server.RatpackServer;
+
+public class RatpackObserveApp {
```
jzheaux: Same comment about class-level docs

reactive-modules/ratpack-rxjava/src/main/java/com/baeldung/rxjava/service/MovieService.java
```diff
@@ -0,0 +1,13 @@
+package com.baeldung.rxjava.service;
+
+import com.baeldung.model.Movie;
+
+import rx.Observable;
+
+public interface MovieService {
```
jzheaux: Since the other is called `MoviePromiseService`, should this one be called `MovieObservableService` for clarity?

reactive-modules/ratpack-rxjava/src/main/java/com/baeldung/rxjava/service/impl/MovieServiceImpl.java
```diff
@@ -0,0 +1,35 @@
+package com.baeldung.rxjava.service.impl;
```
jzheaux: If possible, let's place all of the classes for the same tutorial in the same shared package. Having subpackages is fine, but could we do something like `com.baeldung.ratpack.intro.impl` or `com.baeldung.ratpack.intro.rxjava.service` if we want to be clear that this is rxjava-specific. The important piece is that everything from our tutorial be namespaced to our tutorial so we don't clobber future authors using the same module.

pom.xml
the-java-guy: Is profile, 'default-first' the right place to place the module?

PR #6259, opened 2019-02-01
BAEL-2502

myluckagain:


persistence-modules/spring-data-jpa/src/main/java/com/baeldung/dao/repositories/user/UserRepository.java
```diff
@@ -67,4 +67,8 @@
     @Modifying
     @Query(value = "UPDATE USERS.Users u SET u.status = ? WHERE u.name = ?", nativeQuery = true)
     int updateUserSetStatusForNameNative(Integer status, String name);
+    
+
```
ashleyfrieze: `Minor` - extra whitespace. Probably best having just a blank line between methods.

persistence-modules/spring-data-jpa/src/test/java/com/baeldung/dao/repositories/UserRepositoryIntegrationTest.java
```diff
@@ -390,6 +391,28 @@ public void givenUsersInDBWhenFindByEmailsWithDynamicQueryThenReturnCollection()
...
+        userRepository.save(user1);
+
+        User user2 = new User();
+        user2.setName(USER_NAME_PETER);
+        user2.setEmail(USER_EMAIL2);
+        userRepository.save(user2);
+
+        List<String> names = new ArrayList<>();
+        names.add(USER_NAME_ADAM);
+        names.add(USER_NAME_PETER);
```
ashleyfrieze: `MINOR` - consider whether it might be cleaner to go:

`List<String> names = Arrays.asList(USER_NAME_ADAM, USER_NAME_PETER);`

persistence-modules/spring-data-jpa/src/test/java/com/baeldung/dao/repositories/UserRepositoryIntegrationTest.java
```diff
@@ -390,6 +391,28 @@ public void givenUsersInDBWhenFindByEmailsWithDynamicQueryThenReturnCollection()
...
+
+        Collection<User> usersWithNames = userRepository.findUserByNameList(names);
```
ashleyfrieze: This is correct Java. It's a good thing to use a more abstract type on the left of an operation. However, as the repo returns `List`, casting it to `Collection` is something extra for the reader to process. I think it might be better to receive the `LIst` as `List<User>` just to make it more obvious.

PR #6260, opened 2019-02-01
Refactored the test names and used backtick identifiers.

alimate:
Resolved BAEL-2658

### DIFF
--- a/core-kotlin/src/test/kotlin/com/baeldung/kotlin/junit5/CalculatorTest5.kt
+++ b/core-kotlin/src/test/kotlin/com/baeldung/kotlin/junit5/CalculatorTest5.kt
@@ -7,12 +7,12 @@ class CalculatorTest5 {
     private val calculator = Calculator()
 
     @Test
-    fun whenAdding1and3_thenAnswerIs4() {
+    fun `Adding 1 and 3 should be eqaul to 4`() {
         Assertions.assertEquals(4, calculator.add(1, 3))
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


core-kotlin/src/test/kotlin/com/baeldung/kotlin/junit5/CalculatorTest5.kt
```diff
@@ -7,12 +7,12 @@ class CalculatorTest5 {
     private val calculator = Calculator()
 
     @Test
-    fun whenAdding1and3_thenAnswerIs4() {
+    fun `Adding 1 and 3 should be eqaul to 4`() {
```
KevinGilmore:
```suggestion
    fun `Adding 1 and 3 should be equal to 4`() {
```

core-kotlin/src/test/kotlin/com/baeldung/kotlin/junit5/CalculatorTest5.kt
```diff
@@ -21,7 +21,7 @@ class CalculatorTest5 {
     }
 
     @Test
-    fun whenSquaringNumbers_thenCorrectAnswerGiven() {
+    fun `The square of a number should be eqaul to that number multiplied in itself`() {
```
KevinGilmore:
```suggestion
    fun `The square of a number should be equal to that number multiplied by itself`() {
```

eric-martin:
I will fix the two issues in a separate PR

PR #6274, opened 2019-02-03
[BAEL-2467] Introduction to Spring-cloud-openfeign

freddyaott:


spring-cloud/spring-cloud-openfeign/pom.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>com.baeldung.cloud</groupId>
+    <artifactId>openfeign</artifactId>
+    <version>0.0.1-SNAPSHOT</version>
+    <name>openfeign</name>
+    <description>OpenFeing project for Spring Boot</description>
```
ashleyfrieze: `OpenFeign` is misspelt here.

spring-cloud/spring-cloud-openfeign/src/main/java/com/baeldung/cloud/openfeign/service/impl/JSONPlaceHolderServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+    private JSONPlaceHolderClient jsonPlaceHolderClient;
+
+    @Override
+    public List<Post> getPosts() {
+        return jsonPlaceHolderClient.getPosts();
+    }
+
+    @Override
+    public List<Post> getPostsWrong() {
+        return jsonPlaceHolderClient.getPostsWrong();
```
ashleyfrieze: It's common to have each function of a DAO or low-level client reflected by a service. When that's the case, it is usually caused by the service not adding any business logic, and merely being a conduit to the lower level client.

I think here we have an example of a `getPostsWrong` service method that's purely there to test the error handling behaviour of a client. Is this the best way to make the point? Do we want to just write the unit test that shows the error handling by testing the client directly, rather than via a service?

Just a suggestion.

spring-cloud/spring-cloud-openfeign/src/main/resources/application.properties
```diff
@@ -0,0 +1,3 @@
+spring.application.name= openfeign
+logging.level.com.baeldung.cloud.openfeign.client.JSONPlaceHolderClient: DEBUG
```
ashleyfrieze: Reading the article, I was curious about whether you had to set logging level on a per-interface basis, or whether something like:

`logging.level.com.baeldung.cloud.openfeign.client: DEBUG`

Would set a whole package's debug level. If it can, then perhaps that's a more useful example?

spring-cloud/spring-cloud-openfeign/src/test/java/com/baeldung/cloud/openfeign/OpenfeignUnitTest.java
```diff
@@ -0,0 +1,50 @@
...
+    @Test
+    public void whenSpringContextIsBootstrapped_thenNoExceptions() {
+    }
+
+    @Test
+    public void whenGetPosts_thenListPostSizeGreaterThanZero() {
+
+        List<Post> posts = jsonPlaceHolderService.getPosts();
+
+        assert posts.size() > 0;
```
ashleyfrieze: The assertions in this test are using `assert` as opposed to JUnit's `assertEquals` or `assertThat`

I would recommend moving to using the assertions library from JUnit. See `org.junit.Assert`.

For testing the size of a list, I'd recommend something like `assertFalse(posts.isEmpty())`

spring-cloud/spring-cloud-openfeign/src/test/java/com/baeldung/cloud/openfeign/OpenfeignUnitTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+        assert post != null;
+    }
+
+    @Test
+    public void whenGetPostsWithWrongUrl_thenListPostSizeEqualsZero() {
+
+        List<Post> posts = jsonPlaceHolderService.getPostsWrong();
+
+        assert posts.size() == 0;
```
ashleyfrieze: Similarly, this could be `assertTrue(posts.isEmpty())`

spring-cloud/spring-cloud-openfeign/src/main/java/com/baeldung/cloud/openfeign/exception/BadRequestException.java
```diff
@@ -0,0 +1,29 @@
+package com.baeldung.cloud.openfeign.exception;
+
+public class BadRequestException extends Exception{
```
ashleyfrieze: Need a space before `{`

spring-cloud/spring-cloud-openfeign/src/main/java/com/baeldung/cloud/openfeign/exception/BadRequestException.java
```diff
@@ -0,0 +1,29 @@
...
+
+    private String message;
+
+    public BadRequestException() {
+        super();
```
ashleyfrieze: You don't need to call `super()` it's called by default.

spring-cloud/spring-cloud-openfeign/src/main/java/com/baeldung/cloud/openfeign/exception/BadRequestException.java
```diff
@@ -0,0 +1,29 @@
+package com.baeldung.cloud.openfeign.exception;
+
+public class BadRequestException extends Exception{
+
+    private String message;
```
ashleyfrieze: Why can't you use the `message` property of `Exception`?

spring-cloud/spring-cloud-openfeign/src/main/java/com/baeldung/cloud/openfeign/exception/BadRequestException.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public BadRequestException() {
+        super();
+    }
+
+    public BadRequestException(String message) {
+        super();
+        this.message = message;
```
ashleyfrieze: Why not call `super(message)`?

spring-cloud/spring-cloud-openfeign/src/main/java/com/baeldung/cloud/openfeign/exception/BadRequestException.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "BadRequestException: "+message;
+    }
+
+    @Override
+    public String getMessage() {
+        return message;
```
ashleyfrieze: You wouldn't need this override if you passed the message up to the `Exception` base class

spring-cloud/spring-cloud-openfeign/src/main/java/com/baeldung/cloud/openfeign/exception/BadRequestException.java
```diff
@@ -0,0 +1,29 @@
...
+        this.message = message;
+    }
+
+    public BadRequestException(Throwable cause) {
+        super(cause);
+    }
+
+    @Override
+    public String toString() {
+        return "BadRequestException: "+message;
```
ashleyfrieze: Watch out for spacing around `+` - you could use `getMessage()` instead of this class's `message`

spring-cloud/spring-cloud-openfeign/src/main/java/com/baeldung/cloud/openfeign/exception/NotFoundException.java
```diff
@@ -0,0 +1,30 @@
+package com.baeldung.cloud.openfeign.exception;
+
+
+public class NotFoundException extends Exception{
```
ashleyfrieze: See comments from other exception class

spring-cloud/spring-cloud-openfeign/src/main/java/com/baeldung/cloud/openfeign/service/JSONPlaceHolderService.java
```diff
@@ -0,0 +1,14 @@
+package com.baeldung.cloud.openfeign.service;
+
+
```
ashleyfrieze: extraneous blank line

PR #6297, opened 2019-02-07
BAEL-2738 - Convert String to JsonObject with GSON

dcalap:


dcalap:
Ready for editor's review

PR #6333, opened 2019-02-14
 [BAEL-2718] - Spring JPA Batch Inserts

blockchain-trainer:
New directory under root - could not find a suitable module


spring-data-jpa-batch-dml/src/main/resources/application.properties
```diff
@@ -0,0 +1,16 @@
+# ===============================
+# = DATA SOURCE
+# ===============================
+spring.jpa.hibernate.ddl-auto=create
+spring.datasource.url=jdbc:mysql://localhost:3306/baeldung_db?rewriteBatchedStatements=true
```
jzheaux: I think it would be better to use H2 here so that the user doesn't need to stand up a database to see your examples work.

spring-data-jpa-batch-dml/pom.xml
```diff
@@ -0,0 +1,85 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jzheaux: Let's use an existing module instead, like `spring-data-jpa`.

spring-data-jpa-batch-dml/README.md
```diff
@@ -0,0 +1,6 @@
+## Spring Data JPA batch insert example
```
jzheaux: We'll add this for you afterwards - please don't include this file in the PR.

spring-data-jpa-batch-dml/src/main/java/com/baeldung/jpa/entity/Customer.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Customer Entity class
+ * @author ysharma2512
+ *
+ */
+@Entity
+public class Customer {
+
+    @Id
```
jzheaux: The file here feels a little disorganized. It starts with a field, and then all the setters, followed by fields, and then the constructor. Could we rearrange this to have a more typical setup?

persistence-modules/spring-data-jpa/src/main/java/com/baeldung/controller/CustomerController.java
```diff
@@ -0,0 +1,43 @@
+package com.baeldung.controller;
```
jzheaux: Regarding packages, let's keep all the code for a given tutorial in the same package. It makes things easier to find. So, `src/main/java/com/baeldung/batchinserts` for the code and `src/test/java/com/baeldung/batchinserts` for the tests.

persistence-modules/spring-data-jpa/pom.xml
```diff
@@ -19,6 +19,15 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-data-jpa</artifactId>
         </dependency>
+        <dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-web</artifactId>
+		</dependency>
```
jzheaux: The whitespacing looks a bit funny here.

persistence-modules/spring-data-jpa/pom.xml
```diff
@@ -55,14 +64,14 @@
             <artifactId>junit-jupiter-engine</artifactId>
             <scope>test</scope>
         </dependency>
-        <!-- JUnit platform launcher -->
-        <!-- To be able to run tests from IDE directly -->
+        JUnit platform launcher
```
jzheaux: Let's not have stray words in the pom here. Once they uncomment, then this file will no longer parse.

persistence-modules/spring-data-jpa/pom.xml
```diff
@@ -45,7 +54,7 @@
         </dependency>
         
         <!-- JUnit Jupiter dependencies -->
-        <dependency>
+       <!--  <dependency>
```
jzheaux: Further, I guess I don't understand why this is being commented out?

spring-data-jpa-batch-dml/src/main/java/com/baeldung/jpa/controller/CustomerController.java
```diff
@@ -0,0 +1,41 @@
+package com.baeldung.jpa.controller;
```
jzheaux: Why do we have two `CustomerController`s now? (As well as two of several other things) The duplication seems unnecessary.

spring-ehcache/src/main/java/com/baeldung/cachetest/config/CacheConfig.java
```diff
@@ -1,10 +1,57 @@
 package com.baeldung.cachetest.config;
 
+import java.math.BigDecimal;
```
jzheaux: What is this doing in this PR?

persistence-modules/spring-data-jpa/pom.xml
```diff
@@ -37,6 +37,15 @@
             <artifactId>spring-security-test</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+			<groupId>org.springframework.boot</groupId>
```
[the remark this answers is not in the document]
blockchain-trainer: @jzheaux  Addressed...thanks

PR #6350, opened 2019-02-17
[BAEL-2538] Spring Data JPA @Modifying annotation

dupirefr:


baeldungci:

PR #6358, opened 2019-02-17
[BAEL-12669] - Check modules that don't use our standard parents

amit2103:


baeldungci:

PR #6415, opened 2019-02-26
An example program to make JSON Post request with HttpURLConnection

cupendra:
@pivovarit I created a new pull request for BAEL-2737 (HttpUrlConnection POST Request with JSON). Please review the same

### DIFF
--- a/core-java/src/main/java/com/baeldung/urlconnection/PostJSONWithHttpURLConnection.java
+++ b/core-java/src/main/java/com/baeldung/urlconnection/PostJSONWithHttpURLConnection.java
@@ -9,44 +9,37 @@ import java.net.URL;
 
 public class PostJSONWithHttpURLConnection {
 	
-	public static void main (String []args) {
-		HttpURLConnection con = null;
-		try {
-			//Change the URL with any other publicly accessible POST resource, which accepts JSON request body
-			URL url = new URL ("https://reqres.in/api/users");
-			
-			con = (HttpURLConnection)url.openConnection();
-			con.setRequestMethod("POST");
-			
-			con.setRequestProperty("Content-Type", "application/json; utf-8");
-			con.setRequestProperty("Accept", "application/json");
-			
-			con.setDoOutput(true);
-			
-			//JSON String need to be constructed for the specific resource. 
-			//We may construct complex JSON using any third-party JSON libraries such as jackson or org.json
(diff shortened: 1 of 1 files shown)


core-java/src/main/java/com/baeldung/urlconnection/PostJSONWithHttpURLConnection.java
```diff
@@ -0,0 +1,53 @@
...
+				String responseLine = null;
+				while ((responseLine = br.readLine()) != null) {
+					response.append(responseLine.trim());
+				}
+				System.out.println(response.toString());
+			}
+		}catch (IOException ex) {
+			ex.printStackTrace();
+		}finally {
+			con = null;
```
pivovarit: Not sure why to null it out at the end, it's the end of the main() anyway.

core-java/src/main/java/com/baeldung/urlconnection/PostJSONWithHttpURLConnection.java
```diff
@@ -0,0 +1,53 @@
...
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+public class PostJSONWithHttpURLConnection {
+	
+	public static void main (String []args) {
+		HttpURLConnection con = null;
```
pivovarit: No need to keep that outside the try-catch I think

core-java/src/main/java/com/baeldung/urlconnection/PostJSONWithHttpURLConnection.java
```diff
@@ -0,0 +1,53 @@
...
+			try(BufferedReader br = new BufferedReader(new InputStreamReader(con.getInputStream(), "utf-8"))){
+				StringBuilder response = new StringBuilder();
+				String responseLine = null;
+				while ((responseLine = br.readLine()) != null) {
+					response.append(responseLine.trim());
+				}
+				System.out.println(response.toString());
+			}
+		}catch (IOException ex) {
+			ex.printStackTrace();
```
pivovarit: Would be good to rethrow unchecked exception and not just swallow it

cupendra: Hi, @pivovarit I changed the code as suggested. Please take look it now.

PR #6435, opened 2019-02-27
Bael 2712 lombok builder singular

DaveTheCrane:
Two classes, plus tests, demonstrating use of @Singular for the mini-article


lombok/src/test/java/com/baeldung/lombok/builder/singular/BuilderWithSingularSupportForCollectionsUnitTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.*;
+
+public class BuilderWithSingularSupportForCollectionsUnitTest {
+
+    @Test
+    public void canAddMultipleElementsAsNewCollection() throws Exception {
+        Person person = Person.builder()
+                .givenName("Aaron")
```
KevinGilmore: Let's adjust the indent on all of these examples. They're at six spaces over from the parent line. The guideline is two, although four is also acceptable with fluent APIs like this.

PR #6484, opened 2019-03-08
Update CourseService.java

worldpeacez0991:
From beanutils.java,
The method copyProperties(Object dest, Object orig) is improperly used in this example.


lor6:
Good point @worldpeacez0991 Thanks, merged.

PR #6489, opened 2019-03-08
Update ProxyAcceptCookiePolicy.java

cz111000:
The guide says: When we create an instance of ProxyAcceptCookiePolicy, we pass in a String of the domain address we would like to accept cookies from **in addition to** the original server.


lor6:
Good point @cz111000 Merged, thanks.

cz111000:
> 
> 
> Good point @cz111000 Merged, thanks.

Not at all.

PR #6493, opened 2019-03-09
Bael 2562

Maiklins:
Bael-2562 New section in Generics article


ashleyfrieze:
@Maiklins - please squash these commits into a single commit and then this will be good to go.

PR #6547, opened 2019-03-16
BAEL-2769 Added MockK examples

del7a1:


kotlin-libraries/src/test/kotlin/com/baeldung/mockk/TestableService.kt
```diff
@@ -0,0 +1,12 @@
+package com.baeldung.mockk
+
+class TestableService {
+    fun geDataFromDb(testParameter: String): String {
```
KevinGilmore: Should this be "getDataFromDb"?

kotlin-libraries/src/test/kotlin/com/baeldung/mockk/AnnotationMockKUnitTest.kt
```diff
@@ -0,0 +1,44 @@
...
+    fun setUp() = MockKAnnotations.init(this)
+
+    @Test
+    fun givenServiceMock_whenCallingMockedMethod_thenCorrectlyVerified(){
+        // given
+        every { service1.geDataFromDb("Test Param") } returns "No"
+        // when
+        val result = objectUnderTest.invokeService1()
+        // then
+        assertEquals(result, "No")
```
KevinGilmore: The order of params in assertEquals should be reversed in all the tests, so that it's assertEquals(expected, actual)

kotlin-libraries/src/test/kotlin/com/baeldung/mockk/TestableService.kt
del7a1: Fixed

kotlin-libraries/src/test/kotlin/com/baeldung/mockk/AnnotationMockKUnitTest.kt
del7a1: Good point. I use AssertJ for assertions, so I used to put the expected value at the end 🙃 
Fixed

PR #6559, opened 2019-03-18
BAEL-1931 Kerberos integration in Java

den314:


spring-security-sso/spring-security-sso-kerberos/src/main/java/kerberos/client/SampleService.java
```diff
@@ -0,0 +1,24 @@
...
+@Service
+class SampleService {
+
+	@Value("${app.access-url}")
+	private String endpoint;
+
+	private RestTemplate restTemplate;
+
+	@Autowired
+	public void setRestTemplate(RestTemplate restTemplate) {
```
pivovarit: Use constructor-based injection instead? :)

spring-security-sso/spring-security-sso-kerberos/src/main/java/kerberos/kdc/MiniKdcConfigBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+import java.nio.file.Paths;
+import java.util.Objects;
+
+class MiniKdcConfigBuilder {
+
+	private static final int FIELDS_COUNT = 3;
+	private String workDir;
+	private String confDir;
+	private String keytabName;
+	private String[] principals = { "principal/localhost" };
```
pivovarit: Why not a List? wouldn't it be easier to handle?

spring-security-sso/spring-security-sso-kerberos/src/main/java/kerberos/kdc/MiniKdcConfigBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+
+	String[] build() {
+
+		String[] args = new String[FIELDS_COUNT + principals.length];
+
+		args[0] = workDir;
+		args[1] = confDir;
+		args[2] = keytabName;
+
+		for (int i = 0, j = 3; i < principals.length; i++, j++) {
```
pivovarit: Why not handle that manually? those are just three fields

spring-security-sso/spring-security-sso-kerberos/src/main/java/kerberos/kdc/MiniKdcConfigBuilder.java
den314: refactored to list usage, thank you

spring-security-sso/spring-security-sso-kerberos/src/main/java/kerberos/kdc/MiniKdcConfigBuilder.java
den314: using list we get rid of this part

spring-security-sso/spring-security-sso-kerberos/src/main/java/kerberos/client/SampleService.java
den314: thx - adjusted. Will leave setter for the sake of test showcase.

PR #6579, opened 2019-03-23
BAEL-1970 spring logging level while testing

mprevisic:


spring-boot-testing/src/main/java/com/baeldung/testlog/TestLog.java
```diff
@@ -0,0 +1,22 @@
+package com.baeldung.testlog;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TestLog {
+
+	Logger LOG = LoggerFactory.getLogger(this.getClass());
```
ashleyfrieze: This is using tabs, rather than spaces for indentation. Please reformat using the Baeldung format guidelines.

ashleyfrieze: Generally speaking a logger would be `private static final` - especially if it has an uppercase name.

spring-boot-testing/src/main/java/com/baeldung/testloglevel/TestLogLevelController.java
```diff
@@ -0,0 +1,29 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import com.baeldung.testlog.TestLog;
+
+@RestController
+public class TestLogLevelController {
+
+	Logger LOG = LoggerFactory.getLogger(this.getClass());
```
ashleyfrieze: `private static final` here too please

spring-boot-testing/src/main/java/com/baeldung/testloglevel/TestLogLevelController.java
```diff
@@ -0,0 +1,29 @@
...
+	@GetMapping("/testLogLevel")
+	public String testLogLevel() {
+		LOG.trace("This is a TRACE log");
+		LOG.debug("This is a DEBUG log");
+		LOG.info("This is an INFO log");
+		LOG.error("This is an ERROR log");
+
+		new TestLog().trace("This is a TRACE log in another package");
+		new TestLog().info("This is an INFO log in another package");
+		new TestLog().error("This is an ERROR log in another package");
```
ashleyfrieze: I think that `TestLog` is a bit misleading. You're using another bean to show logging across classes. Why not have a class called `OtherComponent` with the `@Component` annotation and a single entry point - `void processData()` which does logging inside itself. This will perhaps explain your point more precisely?

spring-boot-testing/src/test/java/com/baeldung/testloglevel/LogbackMultiProfileTestLogLevelIntegrationTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.boot.test.web.client.TestRestTemplate;
+import org.springframework.http.ResponseEntity;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(webEnvironment = WebEnvironment.RANDOM_PORT, classes = TestLogLevelApplication.class)
+@EnableAutoConfiguration(exclude = SecurityAutoConfiguration.class)
+@ActiveProfiles("logback-test2")
+public class LogbackMultiProfileTestLogLevelIntegrationTest {
```
ashleyfrieze: This has been formatted with tabs, rather than spaces

PR #6588, opened 2019-03-25
[BAEL-2073] Java 9 Migration Issues and Resolution

macroscopic64:


prejpms/src/main/java/com/baeldung/prejpms/App.java
```diff
@@ -0,0 +1,79 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.sun.crypto.provider.SunJCE;
+
+import sun.misc.BASE64Encoder;
+import sun.reflect.Reflection;
+
+public class App {
+
+    private static final Logger logger = LoggerFactory.getLogger(App.class);
```
ashleyfrieze: This should be `LOGGER` according to usual style guides...

prejpms/src/main/java/com/baeldung/prejpms/App.java
```diff
@@ -0,0 +1,79 @@
...
+        } catch (Throwable e) {
+            logger.error(e.toString());
+        }
+    }
+
+    private static void getCallStackClassNames() {
+        try {
+            int i = 0;
+            StringBuffer sbStack = new StringBuffer();
+            while (true) {
```
ashleyfrieze: it may be easier if you replace this with a `do` / `while` loop, as `while(true)` is harder to reason about

prejpms/src/main/java/com/baeldung/prejpms/App.java
```diff
@@ -0,0 +1,79 @@
...
+    private static void getCrytpographyProviderName() {
+        try {
+            logger.info("1. Java Cryptography Extension - Provider Name: " + new SunJCE().getName() + "\n");
+        } catch (Throwable e) {
+            logger.error(e.toString());
+        }
+    }
+
+    private static void getCallStackClassNames() {
+        try {
```
ashleyfrieze: For the purposes of separating the callstack walker from the boilerplate, can I suggest you extract a function that does the insides of this try block up to the logger statement, and returns a String.

prejpms/src/main/java/com/baeldung/prejpms/App.java
```diff
@@ -0,0 +1,79 @@
...
+            logger.error(e.toString());
+        }
+
+    }
+
+    private static void getBase64EncodedString(String inputString) {
+        try {
+            String encodedString = new BASE64Encoder().encode(inputString.getBytes());
+            logger.info("4. Base Encoded String: " + encodedString);
+        } catch (Throwable e) {
```
ashleyfrieze: Looking at this family of `catch(Throwable ...)` inside each of the functions. It seems like you're not expecting your example to throw at all. In which case, why not have a general catch block in the `main` method, so each of these routines can be revealed without the exception handling clouding the issue?

prejpms/src/main/java/com/baeldung/prejpms/Book.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlRootElement(name = "book")
+public class Book {
+    private long id;
+    private String name;
+
+    public Book() {
+        
+    }
```
ashleyfrieze: Redundant blank line in the constructor

prejpms/src/main/java/com/baeldung/prejpms/Book.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public String getName() {
+        return name;
+    }
+
+    public Long getId() {
+        return id;
+    }
+
```
ashleyfrieze: redundant blank line here too

prejpms/src/main/java/com/baeldung/prejpms/App.java
```diff
@@ -0,0 +1,79 @@
...
+            while (true) {
+                Class<?> caller = Reflection.getCallerClass(i++);
+                if (caller == null) {
+                    break;
+                } else {
+                    sbStack.append(caller.getName())
+                        .append("\n");
+                }
+            }
+            logger.info("2. Call Stack Class Names:\n" + sbStack.toString());
```
ashleyfrieze: Does this logger allow interpolation via `{}` ? If so, it may make the examples easier to read, especially in the version of logging statements pasted into the article.

E.g.

```
// also, try to keep the text shorter, with the objective
// of making it the least to read in the wordpress post - think mobile
logger.info("2. Call Stack:\n {}", sbStack);
```

prejpms/src/main/java/com/baeldung/prejpms/App.java
macroscopic64: I've made each of the methods catch and log its errors and not propagate it so that one error doesn't cause the whole application to exit. This way, it is easy to know which methods failed in a single run. A general catch block in the main method may not help here.

prejpms/src/main/java/com/baeldung/prejpms/App.java
macroscopic64: Extracting a function will make the caller a single line function excluding the try catch block. And in the article, we may have to show both the functions instead of one to connect the dots.
I've explained why I went for individual try-catch blocks in below comment. Please let me know if you feel I should still go ahead.

ashleyfrieze: I'm happy with that.

prejpms/src/main/java/com/baeldung/prejpms/App.java
ashleyfrieze: Agreed. Thanks

ashleyfrieze:
retest this please

PR #6624, opened 2019-03-31
BAEL-2426

alv21:
Java error "Cannot reference “X” before supertype constructor has been called"


core-java-lang-oop-2/src/main/java/com/baeldung/supertypecompilerexception/MyClassSolutionI.java
```diff
@@ -0,0 +1,15 @@
+package com.baeldung.supertypecompilerexception;
+
+public class MyClassSolutionI {
```
ashleyfrieze: I think it may be clearer to use `1`, `2` etc, rather than roman numerals. `I`, `II` etc.

core-java-lang-oop-2/src/main/resources/logback.xml
```diff
@@ -0,0 +1,19 @@
...
+    <logger name="org.springframework" level="WARN" />
+    <logger name="org.springframework.transaction" level="WARN" />
+
+    <!-- in order to debug some marshalling issues, this needs to be TRACE -->
+    <logger name="org.springframework.web.servlet.mvc" level="WARN" />
+
+    <root level="INFO">
+        <appender-ref ref="STDOUT" />
+    </root>
+</configuration>
```
ashleyfrieze: Do we need this file at all?

core-java-lang-oop-2/src/test/resources/.gitignore
```diff
@@ -0,0 +1,13 @@
...
+/target
+/neoDb*
+/data
+/src/main/webapp/WEB-INF/classes
+*/META-INF/*
+
+# Packaged files #
+*.jar
+*.war
+*.ear
```
ashleyfrieze: This is another git ignore - in the resources...?

core-java-lang-oop-2/.gitignore
```diff
@@ -0,0 +1,26 @@
...
+
+# Files generated by integration tests
+*.txt
+backup-pom.xml
+/bin/
+/temp
+
+#IntelliJ specific
+.idea/
+*.iml
```
ashleyfrieze: This has a lot of stuff in that seems outside of the scope of the example - where did it originate?

May be simpler to just have

```
target/
.idea/
bin/
*.iml
```

Perhaps the equivalent eclipse project files too...?

PR #6627, opened 2019-03-31
Bael 617: BIRT reporting with Spring Boot

gmconte:


pom.xml
```diff
@@ -648,7 +648,8 @@
                 <module>spring-boot-jasypt</module>
                 <module>spring-boot-keycloak</module>
                 <module>spring-boot-logging-log4j2</module>
-                <module>spring-boot-mvc</module>
+				<module>spring-boot-mvc</module>
```
KevinGilmore: Can we convert the indents back to spaces here and below before merging?

gmconte: Hi Kevin, all done. I'm sorry I did it in 2 commits instead of one

PR #6629, opened 2019-03-31
Guide to Guava Multiset

sushant57:


guava-collections/src/test/java/org/baeldung/guava/GuavaMultiSetUnitTest.java
```diff
@@ -0,0 +1,67 @@
...
+        bookStore.add("Potter");
+        bookStore.add("Potter");
+
+        assertThat(bookStore.contains("Potter")).isTrue();
+        assertThat(bookStore.count("Potter")).isEqualTo(3);
+
+        bookStore.remove("Potter");
+        assertThat(bookStore.contains("Potter")).isTrue();
+        assertThat(bookStore.count("Potter")).isEqualTo(2);
+    }
```
ashleyfrieze: This is really two different scenarios. One is that adding makes things present. The other is that removing after adding only decrements the count, rather than removes the object completely.

ashleyfrieze: So, I'd recommend splitting into two tests.

guava-collections/src/test/java/org/baeldung/guava/GuavaMultiSetUnitTest.java
```diff
@@ -0,0 +1,67 @@
...
+    public void givenMultiSet_whenSetCount_shouldReturnCorrectCount() {
+        Multiset<String> bookStore = HashMultiset.create();
+        bookStore.setCount("Potter", 50);
+        assertThat(bookStore.count("Potter")).isEqualTo(50);
+    }
+
+    @Test
+    public void givenMultiSet_whenSettingNegativeCount_shouldThrowException() {
+        Multiset<String> bookStore = HashMultiset.create();
+        assertThatThrownBy(() -> bookStore.setCount("Potter", -1)).isInstanceOf(IllegalArgumentException.class);
```
ashleyfrieze: I recommend a line break before `.isInstanceOf`

:+1: for using AssertJ here!

guava-collections/src/test/java/org/baeldung/guava/GuavaMultiSetUnitTest.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Test
+    public void givenMultiSet_whenSettingCountWithOldCount_shouldReturnCorrectValue() {
+        Multiset<String> bookStore = HashMultiset.create();
+        assertThat(bookStore.setCount("Potter", 0, 2)).isTrue();
```
ashleyfrieze: The 0 is an edge case here. I.e. when there are none, you have to provide 0. There's a mid-level case, which is that you know the count of some present items - say _there are 6_ and you set the count on the basis of that assumption.

In fact, there are three cases:

- None present - special case 0
- We know the count correctly - case n
- We do not know the count - so we get a failure

I suggest breaking each into its own test.

guava-collections/src/test/java/org/baeldung/guava/GuavaMultiSetUnitTest.java
sushant57: Done.

guava-collections/src/test/java/org/baeldung/guava/GuavaMultiSetUnitTest.java
sushant57: Done.

guava-collections/src/test/java/org/baeldung/guava/GuavaMultiSetUnitTest.java
sushant57: Done.

PR #6635, opened 2019-03-31
Bael 2434

cscib:


spring-core/src/main/java/com/baeldung/resource/LoadResourceConfig.java
```diff
@@ -0,0 +1,55 @@
...
+    @Bean
+    public String resourceStringUsingStreamUtils() throws IOException {
+        try (InputStream inputStream = resource.getInputStream()) {
+            return StreamUtils.copyToString(inputStream, StandardCharsets.UTF_8);
+        }
+    }
+
+    @Bean
+    public String resourceStringUsingGuava() throws IOException {
+        try (final Reader reader = new InputStreamReader(resource.getInputStream())) {
```
ashleyfrieze: I recommend removing all instances of `final` from this example. It's not adding much and it's just more for the reader to read.

spring-core/src/test/java/com/baeldung/resource/LoadResourceAsStringIntegrationTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+import org.springframework.test.context.support.AnnotationConfigContextLoader;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(loader = AnnotationConfigContextLoader.class, classes = LoadResourceConfig.class)
+public class LoadResourceAsStringIntegrationTest {
+
+    private static final String ORIGINAL_DESCRIPTION = "This is a resource text file. This file will be loaded as a " + "resource and use its contents as a string.";
```
ashleyfrieze: This `final` is expected and what you'd expect.

spring-core/src/main/java/com/baeldung/resource/LoadResourceConfig.java
```diff
@@ -0,0 +1,55 @@
...
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.nio.charset.StandardCharsets;
+
+@Configuration
+public class LoadResourceConfig {
+    @Value("classpath:resource.txt")
+    private Resource resource;
+
+    @Value("#{T(org.apache.commons.io.FileUtils).readFileToString(" + "T(org.springframework.util.ResourceUtils).getFile('classpath:resource.txt')" + ")}")
```
ashleyfrieze: I'm not sure what the aim is here. If it's to show how much you can put into a SPEL statement, then great. If it's more to show that you can wire in via SPEL, then perhaps we could have a static function somewhere with the _meat_ of this code in, and you can just wire in calling that static function.

This SPEL code is broken up with `" + "` but not line breaks, making it harder to read. It could be broken with `" + <linebreak>` before each `.functionCall(`

PR #6646, opened 2019-04-02
BAEL-1959 Classpath contains multiple bindings

priyeshmashelkar:
Updated pom.xml to exclude log4j binding


priyeshmashelkar:
Created PR with only the file related to article BAEL-1959

PR #6695, opened 2019-04-09
BAEL-2841 spring data jpa populators

sumitsg34:


persistence-modules/spring-data-jpa-2/src/main/resources/fruit-data.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
```
KevinGilmore: Let's add the guava one to the XML file as well

PR #6717, opened 2019-04-13
maven properties should be at top

paulvi:


lor6:
Hi @paulvi In the code style for this repository, we keep Maven properties at the end of the file.

paulvi:
Maybe then add comment in README on within pom.xml that

`Maven properties at the end of the file`

eugenp:
Out of curiosity, @paulvi - is there any official guidance about where to keep these? I don't remember ever seeing something like that, but I might just not be aware of it.

paulvi:
From time to time I myself looking for some reference of standard POM content,
I only can give https://maven.apache.org/pom.html#Quick_Overview

But mainly it is because most of projects have the same order of sections.
Now Spring Boot should be the most known example
https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/pom.xml

PR #6727, opened 2019-04-14
BAEL-2661 - Groovy def keyword

SmartyAnsh:


core-groovy-2/pom.xml
```diff
@@ -43,13 +64,40 @@
...
+                            <goal>verify</goal>
+                        </goals>
+                        <configuration>
+                            <includes>
+                                <include>**/*Test5.java</include>
+                            </includes>
+                        </configuration>
+                    </execution>
+                </executions>
+			</plugin>
```
KevinGilmore: Let's use spaces for the indentation

core-groovy-2/pom.xml
```diff
@@ -71,10 +119,16 @@
     </repositories>
 
     <properties>
+        <junit.platform.version>1.0.0</junit.platform.version>
+<!--         <groovy.version>2.4.13</groovy.version> -->
```
KevinGilmore: Let's remove these commented lines.

core-groovy-2/pom.xml
SmartyAnsh: resolved

core-groovy-2/pom.xml
SmartyAnsh: removed commented lines

PR #6742, opened 2019-04-16
#BAEL-2828 - first and top example methods added

dcalap:


dcalap:
Fixed the import and removed methods because 'same erasure' error using Passenger and Optional<Passenger>(seen in https://stackoverflow.com/questions/1998544/method-has-the-same-erasure-as-another-method-in-type)

PR #6764, opened 2019-04-19
Bael 2889: Run Jar application with command line arguments

liesheng:


core-java-8-2/src/main/java/com/baeldung/jarArguments/example_manifest.txt
```diff
@@ -0,0 +1 @@
+Main-Class: com.baeldung.jarArguments.JarExample
```
KevinGilmore: this file should be in src/main/resources

PR #6789, opened 2019-04-21
fix intermediate mvn error

rockoder:


eric-martin:
Thanks @rockoder - I ended up making the same change in a different branch to get around the jenkins timeout above.

PR #6814, opened 2019-04-25
BAEL-2587: create separate module for cognito

adrianprecub:


spring-5-security-cognito/README.md
```diff
@@ -0,0 +1,3 @@
+## Relevant articles:
+
+- [Authenticating with Amazon Cognito Using Spring Security](http://www.baeldung.com/authenticating-with-amazon-cognito-using-spring-security)
```
jzheaux: @adrianprecub thanks for adding this. Will you please make sure the link is `https`? Also, I think this will actually be called: `https://www.baeldung.com/spring-security-oauth-cognito`

adrianprecub: yep, i'll update the pr

PR #6821, opened 2019-04-25
Update BinaryTree.java

sushibossftw:
Fixed wrong code block in traverseLevelOrder method.


maibin:
Hey @sushibossftw, could you please let us know what's wrong with the previous code in here?
Thanks!

sushibossftw:
In BFS you check the both right and left child nodes and add them to queue if they’re not null. In previous code there was one more unnecessary left child checking condition. I simply changed the extra left condition to right check so.

PR #6861, opened 2019-04-29
Adding files for bifunction module

soufiane-cheouati:


core-java-8/src/main/java/com/baeldung/bifunction/ArithmeticOeration.java
```diff
@@ -0,0 +1,13 @@
+package com.baeldung.bifunction;
+
+import java.util.function.BiFunction;
+
+public class ArithmeticOeration {
```
ashleyfrieze: This should be `ArithmeticOperation` - you lost a `p` :)

core-java-8/src/main/java/com/baeldung/bifunction/Calculator.java
```diff
@@ -0,0 +1,28 @@
+package com.baeldung.bifunction;
+
+import java.util.function.BiFunction;
+
+public class Calculator {
+	
+    private Integer a = 0;
+    private Integer b = 0;
```
ashleyfrieze: While I appreciate that the functional operations need `Integer`, I think this calculating pojo would naturally represent its own inner values using `int`. There's no meaning to `null` values as far as the calculator is concerned. This would affect the members and the constructor. The functional part would still use `Integer` and the compiler would do the necessary boxing/unboxing.

core-java-8/src/main/java/com/baeldung/bifunction/Calculator.java
```diff
@@ -0,0 +1,28 @@
...
+    public Calculator(Integer a, Integer b) {
+        this.a = a;
+        this.b = b;
+    }
+
+    public Integer calculate(Integer a, Integer b, BiFunction<Integer, Integer, Integer> obj){
+        return obj.apply(a, b);
+    }
+	
+    public Integer calculate(BiFunction<Integer, Integer, Integer> obj){
```
ashleyfrieze: `obj` is a poor name for the thing that's coming in here. We're asking the calculator to apply a calculation to its own members. Surely the value would called `calculation` or `method` or `operation`

core-java-8/src/main/java/com/baeldung/bifunction/Calculator.java
```diff
@@ -0,0 +1,28 @@
...
+	
+    private Integer a = 0;
+    private Integer b = 0;
+	
+    public Calculator(Integer a, Integer b) {
+        this.a = a;
+        this.b = b;
+    }
+
+    public Integer calculate(Integer a, Integer b, BiFunction<Integer, Integer, Integer> obj){
```
ashleyfrieze: I can't see what this function adds to `Calculator`. It doesn't use the members `a` or `b`. It could, therefore, be static. It's doing nothing more than the `apply` function on the method you pass in, which makes it a pretty redundant thing.

PR #6907, opened 2019-05-07
Java 12 String API Test

rockoder:


core-java-modules/core-java-12/src/test/java/com/baeldung/string/StringAPITest.java
```diff
@@ -0,0 +1,43 @@
...
+public class StringAPITest {
+
+    @Test
+    public void whenPositiveArgument_thenReturnIndentedString() {
+        String multilineStr = "This is\na multiline\nstring.";
+        String outputStr = "   This is\n   a multiline\n   string.\n";
+
+        String postIndent = multilineStr.indent(3);
+
+        assertThat(outputStr, equalTo(postIndent));
```
KevinGilmore:
```suggestion
        assertThat(postIndent, equalTo(outputStr));
```

KevinGilmore: Since we're using the Hamcrest matchers, and not simply assertEquals, let's put the arguments in this order for all the tests.

PR #6966, opened 2019-05-15
BAEL-2378 Non-blocking Spring Boot with Kotlin Coroutines

CROSP:


spring-boot-kotlin-io/src/main/kotlin/com/baeldung/nonblockingcoroutines/config/DatastoreConfig.kt
```diff
@@ -0,0 +1,22 @@
...
+
+
+@Configuration
+@EnableR2dbcRepositories
+class DatabaseConfig : AbstractR2dbcConfiguration() {
+
+    @Bean
+    override fun connectionFactory(): ConnectionFactory {
+        return PostgresqlConnectionFactory(PostgresqlConnectionConfiguration.builder()
+                // Set options here
```
ashleyfrieze: Does this example have any options to set? Do we need this comment?

spring-boot-kotlin-io/src/main/kotlin/com/baeldung/nonblockingcoroutines/config/RouterConfig.kt
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.web.reactive.function.server.coRouter
+
+@Configuration
+class RouterConfiguration {
+
+    @FlowPreview
+    @Bean
+    fun productRoutes(productsHandler: ProductsHandler) = coRouter {
+        GET("/", productsHandler::findAll)
+        GET("/{id}", productsHandler::findOne)
```
ashleyfrieze: Just so I understand, the router cannot operate on `Flux`, `Flow` or `Mono` - it needs the downstream handler to `suspend` in order to await the actual value?

spring-boot-kotlin-io/src/main/kotlin/com/baeldung/nonblockingcoroutines/handlers/ProductsHandler.kt
```diff
@@ -0,0 +1,49 @@
...
+@Component
+class ProductsHandler(
+        @Autowired var webClient: WebClient,
+        @Autowired var productRepository: ProductRepositoryCoroutines) {
+
+    @FlowPreview
+    suspend fun findAll(request: ServerRequest): ServerResponse =
+            ServerResponse.ok().json().bodyAndAwait(productRepository.getAllProducts())
+
+
```
ashleyfrieze: Watch the extra spacing here.

spring-boot-kotlin-io/src/main/kotlin/com/baeldung/nonblockingcoroutines/config/DatastoreConfig.kt
CROSP: We need to provide database configuration options here in order to access a postgresql, in that case, instance. Locally, I used credentials for accessing my instance of PostgreSQL. 
As I was told by other authors, code examples should be compilable, but not always usable. 
What is the right way?

spring-boot-kotlin-io/src/main/kotlin/com/baeldung/nonblockingcoroutines/handlers/ProductsHandler.kt
CROSP: Thank you. My bad, missed this issue.

spring-boot-kotlin-io/src/main/kotlin/com/baeldung/nonblockingcoroutines/config/RouterConfig.kt
CROSP: We can also use Flux/Mono or Flow without any problems. Suspending functions are handled by the framework, Here is an example of using Flux/Mono with the DSL router definition. https://github.com/eugenp/tutorials/tree/master/spring-reactive-kotlin

spring-boot-kotlin-io/src/main/kotlin/com/baeldung/nonblockingcoroutines/handlers/ProductsHandler.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.web.reactive.function.client.awaitExchange
+import org.springframework.web.reactive.function.server.ServerRequest
+import org.springframework.web.reactive.function.server.ServerResponse
+import org.springframework.web.reactive.function.server.bodyAndAwait
+import org.springframework.web.reactive.function.server.json
+
+@Component
+class ProductsHandler(
+        @Autowired var webClient: WebClient,
+        @Autowired var productRepository: ProductRepositoryCoroutines) {
```
ashleyfrieze: Indentation in this file seems out of alignment. Would `@Autowired` be on the same line as the `var` definition?

CROSP: Hi @ashleyfrieze , as far as I know, this should be formatted in that way. This a primary constructor of the class and arguments are separated by a comma. Here are some official formatting conventions https://kotlinlang.org/docs/reference/coding-conventions.html#class-header-formatting

PR #6985, opened 2019-05-19
BAEL-773 - Ratpack with Groovy

SmartyAnsh:


ratpack/pom.xml
```diff
@@ -26,11 +26,21 @@
             <artifactId>ratpack-core</artifactId>
             <version>${ratpack.version}</version>
         </dependency>
+        <dependency>
+    		<groupId>org.codehaus.groovy</groupId>
+		    <artifactId>groovy-sql</artifactId>
+		    <version>2.4.15</version>
```
ashleyfrieze: I think these `version` number need to be in `properties`

ratpack/build.gradle
```diff
@@ -14,6 +14,8 @@ if (!JavaVersion.current().java8Compatible) {
 
 apply plugin: "io.ratpack.ratpack-java"
 apply plugin: 'java'
+apply plugin: 'groovy'
+apply plugin: 'io.ratpack.ratpack-groovy'
```
ashleyfrieze: Why do we need both the gradle file AND the pom?

ratpack/src/main/groovy/com/baeldung/Ratpack.groovy
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+
+    handlers {
+        
+        get {
+            render 'Hello World from Ratpck with Groovy!!'
+        }
+        
+        get("greet/:name"){Context ctx ->
```
ashleyfrieze: Watch the whitespace here between `{}` and `()`

ratpack/src/main/groovy/com/baeldung/Ratpack.groovy
```diff
@@ -0,0 +1,61 @@
...
+            render "Hello " + ctx.getPathTokens().get("name") + " !!!"
```
ashleyfrieze: (being picky, I'd not have the space before `!!!`

ashleyfrieze: Would it be cleaner to use the inferred single parameter here? or is it better to declare `Context`?

ratpack/src/main/groovy/com/baeldung/Ratpack.groovy
```diff
@@ -0,0 +1,61 @@
...
+        }
+        
+        get("data"){
```
ashleyfrieze: `) {` spacing

ratpack/src/main/groovy/com/baeldung/RatpackGroovyApp.groovy
```diff
@@ -0,0 +1,12 @@
+package com.baeldung;
+
+public class RatpackGroovyApp {
+
+    public static void main(String[] args) {
+        File file = new File("src/main/groovy/com/baeldung/Ratpack.groovy");
+        def shell = new GroovyShell()  
+        shell.evaluate(file)
```
ashleyfrieze: Why does this defer to a new groovy shell? Are we not already in a groovy runtime?

ratpack/src/main/resources/User.sql
```diff
@@ -0,0 +1,10 @@
+DROP TABLE IF EXISTS user;
+CREATE TABLE user (
+  id bigint auto_increment primary key,
+  title varchar(255),
+  name varchar(255),
+  country varchar(255)
+);
+
+insert into user values(1,'Mr.','Norman', 'USA');
+insert into user values(2,'Mr.','Lewis', 'USA');
```
ashleyfrieze: It's a small point, but:

- `Mr.` would conventionally be just `Mr`
- People usually have two names - maybe choose movie stars
- Shall we have a mix of genders to be inclusive to our female readers?

ratpack/src/test/groovy/com/baeldung/RatpackGroovySpec.groovy
```diff
@@ -0,0 +1,47 @@
...
+import ratpack.test.http.TestHttpClient;
+import ratpack.test.ServerBackedApplicationUnderTest;
+import org.junit.Test;
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+import ratpack.test.MainClassApplicationUnderTest
+
+class RatpackGroovySpec {
+
+    ServerBackedApplicationUnderTest aut = new MainClassApplicationUnderTest(RatpackGroovyApp.class)
```
ashleyfrieze: let's use a full name for the application under test - like `application` or even `app` - `aut` is very cryptic.

ratpack/src/test/groovy/com/baeldung/RatpackGroovySpec.groovy
```diff
@@ -0,0 +1,47 @@
...
+    @Delegate TestHttpClient client = TestHttpClient.testHttpClient(aut)
+
+    @Test
+    void "test if app is started"() {
+        when:
+        get("")
+
+        then:
+        response.statusCode == 200
+        assert response.body.text == "Hello World from Ratpck with Groovy!!"
```
ashleyfrieze: would we not use an assertions library? rather than just `assert`? I'm not sure what's conventional in Groovy? I'd rather hope there's something that does more than just `true` or `false` :)

ratpack/src/main/groovy/com/baeldung/Ratpack.groovy
SmartyAnsh: ctx is the parameter available in the handlers.
It is better to declare the Context as reader will get to know the parameter available inside closure if of Context type and hence can understand all the available properties and behavior of it.

ratpack/src/main/groovy/com/baeldung/RatpackGroovyApp.groovy
SmartyAnsh: We run this Application class while testing. We run it through jUnit test and Groovy Shell is required to run the Ratpack.groovy script to make it available as mock for testing.

ratpack/pom.xml
SmartyAnsh: will make the change

ratpack/build.gradle
SmartyAnsh: gradle was already there in the Ratpack module to let reader build and run it as gradle project

ratpack/src/main/groovy/com/baeldung/Ratpack.groovy
SmartyAnsh: will correct it

ratpack/src/main/groovy/com/baeldung/Ratpack.groovy
SmartyAnsh: will correct it

ratpack/src/main/resources/User.sql
SmartyAnsh: agreed

ratpack/src/test/groovy/com/baeldung/RatpackGroovySpec.groovy
SmartyAnsh: agreed

ratpack/src/test/groovy/com/baeldung/RatpackGroovySpec.groovy
SmartyAnsh: assert is by default available in Groovy spock framework and is more conventional. Also we are just comparing the text and status code so we don't need to have java like assert statements here.

ratpack/src/main/groovy/com/baeldung/Ratpack.groovy
```diff
@@ -0,0 +1,79 @@
...
+
+    handlers {
+
+        get { render 'Hello World from Ratpck with Groovy!!' }
+
+        get("greet/:name") { Context ctx ->
+            render "Hello " + ctx.getPathTokens().get("name") + "!!!"
+        }
+
+        get("data"){
```
ashleyfrieze: `MINOR` - spacing between `)` and `{`

ratpack/src/main/groovy/com/baeldung/Ratpack.groovy
```diff
@@ -0,0 +1,79 @@
...
+            render Jackson.json([title:"Mr",name:"Norman",country:"USA"])
```
ashleyfrieze: `MINOR` - the spacing is missing between commas

ratpack/src/main/groovy/com/baeldung/Ratpack.groovy
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        post("user") {
+            Promise<User> user = parse(Jackson.fromJson(User))
+            user.then { u -> render u.name }
```
ashleyfrieze: Question - could this also be expressed without the local variable?

e.g.

```
parse(Jackson.fromJson(User))
    .then { user -> render user.name }
```

It might read more like a DSL if it can be made more fluent.

ratpack/src/main/groovy/com/baeldung/Ratpack.groovy
```diff
@@ -0,0 +1,79 @@
...
+        }
+        
+        get('fetchUserName/:id') { Context ctx ->
+            Connection connection = ctx.get(DataSource.class).getConnection()
+            PreparedStatement queryStatement = connection.prepareStatement("select name from user where id=?")
+            queryStatement.setInt(1, Integer.parseInt(ctx.getPathTokens().get("id")))
+            ResultSet resultSet = queryStatement.executeQuery()
+            resultSet.next()
+            render resultSet.getString(1)
+
```
ashleyfrieze: redundant whitespace

ratpack/src/main/groovy/com/baeldung/Ratpack.groovy
SmartyAnsh: Fixed

ratpack/src/main/groovy/com/baeldung/Ratpack.groovy
SmartyAnsh: fixed

ratpack/src/main/groovy/com/baeldung/Ratpack.groovy
SmartyAnsh: Yes it is possible, however to keep the promise example easy I have not changed it.
For variation, I have used similar DSL structure in post('addUser') handler

ratpack/src/main/groovy/com/baeldung/Ratpack.groovy
```diff
@@ -0,0 +1,76 @@
...
+    bindings {
+        module(HikariModule) { config ->
+            config.dataSourceClassName = 'org.h2.jdbcx.JdbcDataSource'
+            config.addDataSourceProperty('URL', "jdbc:h2:mem:devDB;INIT=RUNSCRIPT FROM 'classpath:/User.sql'")
+        }
+    }
+
+    handlers {
+
+        get { render 'Hello World from Ratpck with Groovy!!' }
```
ashleyfrieze: Typo - this needs to be `Ratpack`

SmartyAnsh: typo fixed

PR #7037, opened 2019-05-27
govindpatel - hexagonal-architecture

govindpatel:
MessageService uses MessageDAOInterface(port) which later can be replaced with some other DAO implementation.


eric-martin:
test this please

eric-martin:
(that was a comment to kick off the CI build again)

PR #7042, opened 2019-05-28
[BAEL-2959] A Guide to MapDB code

sjmillington:


libraries/src/test/java/com/baeldung/mapdb/CollectionsUnitTest.java
```diff
@@ -0,0 +1,33 @@
...
+  public void givenSetCreatedInDB_whenMultipleElementsAdded_checkOnlyOneExists() {
+
+    DB db = DBMaker.memoryDB().make();
+
+    NavigableSet<String> set = db.
+            treeSet("mySet")
+            .serializer(Serializer.STRING)
+            .createOrOpen();
+
+    final String MY_STRING = "Baeldung!";
```
ashleyfrieze: Let's take the `final` keyword out altogether. It's really not a good keyword unless you're trying to do something special.

In this case you have a local variable, which you're not changing and you're trying to write it as though it's a class level constant.

`private static final String MY_STRING` would be a fine thing to have for a global constant, but that's not what's happening here.

ashleyfrieze: Let's call it `myString`

libraries/src/test/java/com/baeldung/mapdb/HelloBaeldungUnitTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+public class HelloBaeldungUnitTest {
+
+  @Test
+  public void givenInMemoryDBInstantiateCorrectly_whenDataSavedAndRetrieved_checkRetrievalCorrect() {
+
+    DB db = DBMaker.memoryDB().make();
+
+    final String WELCOME_MESSAGE_KEY = "Welcome Message";
+    final String WELCOME_MESSAGE = "Hello Baeldung!";
```
ashleyfrieze: Similar feedback relating to `final` here too

libraries/src/test/java/com/baeldung/mapdb/HelloBaeldungUnitTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    HTreeMap myMap = db.hashMap("myMap").createOrOpen();
+    myMap.put(WELCOME_MESSAGE_KEY, WELCOME_MESSAGE);
+
+    String welcomeMessage = (String) myMap.get(WELCOME_MESSAGE_KEY);
+
+    db.close();
+
+    assertEquals(WELCOME_MESSAGE, welcomeMessage);
+
```
ashleyfrieze: redundant blank line

libraries/src/test/java/com/baeldung/mapdb/HelloBaeldungUnitTest.java
```diff
@@ -0,0 +1,51 @@
...
+  @Test
+  public void givenInFileDBInstantiateCorrectly_whenDataSavedAndRetrieved_checkRetrievalCorrect() {
+
+    DB db = DBMaker.fileDB("file.db").make();
+
+    final String WELCOME_MESSAGE_KEY = "Welcome Message";
+    final String WELCOME_MESSAGE = "Hello Baeldung!";
+
+    HTreeMap myMap = db.hashMap("myMap").createOrOpen();
```
ashleyfrieze: Why does this not have specialised types? Can it only be done with no type specifier?

libraries/src/test/java/com/baeldung/mapdb/SortedTableMapUnitTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.mapdb.volume.Volume;
+
+import static junit.framework.Assert.assertEquals;
+
+public class SortedTableMapUnitTest {
+
+  @Test
+  public void givenValidSortedTableMapSetup_whenQueried_checkValuesCorrect() {
+
+    final String VOLUME_LOCATION = "sortedTableMapVol.db";
```
ashleyfrieze: this looks like it would rather be `private static final`, which would be good.

libraries/src/test/java/com/baeldung/mapdb/SortedTableMapUnitTest.java
```diff
@@ -0,0 +1,50 @@
...
+    SortedTableMap.Sink<Integer, String> sink =
+        SortedTableMap.create(
+            vol,
+            Serializer.INTEGER,
+            Serializer.STRING
+        ).createFromSink();
+
+    //add content
+    for(int i = 0; i < 100; i++){
+      sink.put(i, "Value " + String.valueOf(i));
```
ashleyfrieze: It might be more correct to use the `Integer` method to create the string - https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#toString-int-

libraries/src/test/java/com/baeldung/mapdb/CollectionsUnitTest.java
sjmillington: Changed MY_STRING to myString and dropped the final

libraries/src/test/java/com/baeldung/mapdb/HelloBaeldungUnitTest.java
sjmillington: Same as above

libraries/src/test/java/com/baeldung/mapdb/HelloBaeldungUnitTest.java
sjmillington: Removed the redundant blank line (assuming you meant the one below the `assertEquals`

libraries/src/test/java/com/baeldung/mapdb/HelloBaeldungUnitTest.java
sjmillington: It only accepts the specialised types when defining it alongside the `.keySerializer...` and `.valueSerializer` - which I didn't want to bring the serializers into the hello world example.

libraries/src/test/java/com/baeldung/mapdb/SortedTableMapUnitTest.java
sjmillington: moved `VOLUME_LOCATION` into a `private static final` field.

libraries/src/test/java/com/baeldung/mapdb/SortedTableMapUnitTest.java
sjmillington: Changed `String.valueOf(i)` to `Integer.toString(i)`

PR #7053, opened 2019-05-30
[BAEL-2935] - added break mechanisms for Stream.forEach

catalin-burcea:


java-streams-2/src/main/java/com/baeldung/breakforeach/CustomSpliterator.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+
+public class CustomSpliterator extends Spliterators.AbstractSpliterator<String> {
+
+    Spliterator<String> splitr;
+    Predicate<String> predicate;
+    boolean isMatched = true;
```
ashleyfrieze: This should really this private.

java-streams-2/src/main/java/com/baeldung/breakforeach/CustomSpliterator.java
```diff
@@ -0,0 +1,31 @@
...
+
+    CustomSpliterator(Spliterator<String> splitr, Predicate<String> predicate) {
+        super(splitr.estimateSize(), 0);
+        this.splitr = splitr;
+        this.predicate = predicate;
+    }
+
+    @Override
+    public boolean tryAdvance(Consumer<? super String> consumer) {
```
ashleyfrieze: Minor question - is this meant to be threadsafe? Streams can be run with `parallel`. Worth checking if this might be called by multiple threads. If that were the case, then it could be marked as `synchronized"... maybe?

java-streams-2/src/main/java/com/baeldung/breakforeach/CustomSpliterator.java
```diff
@@ -0,0 +1,31 @@
...
+        boolean hadNext = splitr.tryAdvance(elem -> {
+            if (predicate.test(elem)) {
```
ashleyfrieze: Should we check `isMatched` here - just in case there were calls to `tryAdvance` after the first time `tryAdvance` returns `false`?

java-streams-2/src/main/java/com/baeldung/breakforeach/CustomSpliterator.java
```diff
@@ -0,0 +1,31 @@
+package com.baeldung.breakforeach;
+
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+
+public class CustomSpliterator extends Spliterators.AbstractSpliterator<String> {
```
ashleyfrieze: Let's make this generic - `CustomSpliterator<T>`

java-streams-2/src/main/java/com/baeldung/breakforeach/CustomSpliterator.java
```diff
@@ -0,0 +1,31 @@
...
+
+    Spliterator<String> splitr;
```
ashleyfrieze: `Spliterator<T>`

java-streams-2/src/main/java/com/baeldung/breakforeach/CustomSpliterator.java
```diff
@@ -0,0 +1,31 @@
...
+    Predicate<String> predicate;
```
ashleyfrieze: `Predicate<T>`

java-streams-2/src/main/java/com/baeldung/breakforeach/CustomSpliterator.java
ashleyfrieze: `Consumer<? extends T>`

java-streams-2/src/main/java/com/baeldung/breakforeach/CustomTakeWhile.java
```diff
@@ -0,0 +1,14 @@
+package com.baeldung.breakforeach;
+
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+public class CustomTakeWhile {
+
+    public static Stream<String> takeWhile(Stream<String> stream, Predicate<String> predicate) {
```
ashleyfrieze: `public static <T> Stream<T> takeWhile(Stream<T>...`

java-streams-2/src/main/java/com/baeldung/breakforeach/CustomForEach.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.Spliterator;
+import java.util.function.BiConsumer;
+import java.util.stream.Stream;
+
+public class CustomForEach {
+
+    static class Breaker {
+        private boolean shouldBreak = false;
+
+        void breakForEach() {
```
ashleyfrieze: Let's rename this `break`

java-streams-2/src/main/java/com/baeldung/breakforeach/CustomForEach.java
```diff
@@ -0,0 +1,32 @@
...
+            shouldBreak = true;
+        }
+
+        boolean get() {
+            return shouldBreak;
+        }
+    }
+
+    static void forEach(Stream<String> stream, BiConsumer<String, Breaker> consumer) {
```
ashleyfrieze: Let's make this generic - I did that in the draft of the article.

java-streams-2/src/main/java/com/baeldung/breakforeach/CustomForEach.java
```diff
@@ -0,0 +1,32 @@
+package com.baeldung.breakforeach;
+
+import java.util.Spliterator;
+import java.util.function.BiConsumer;
+import java.util.stream.Stream;
+
+public class CustomForEach {
+
+    static class Breaker {
```
ashleyfrieze: The functions and types here should be `public` not `package private`

java-streams-2/src/main/java/com/baeldung/breakforeach/CustomForEach.java
catalin-burcea: It's not possible. `break` is a keyword. Do you have another suggestion?

PR #7122, opened 2019-06-13
Array Iterator (BAEL-2993)

jaafarnaddaf:


core-java-modules/core-java-arrays/src/main/java/com/baeldung/array/ArrayIterator.java
```diff
@@ -0,0 +1,62 @@
...
+		}
+
+		int[] row = arr[index];
+		for (int i = 0; i < row.length; i++) {
+			System.out.print(row[i] + "\t");
+		}
+	}
+
+	public static void printWithOffset(int[][] arr, int row, int col, int rowOffset, int colOffset) {
+		for (; row >= 0 && row < arr.length; row += rowOffset) {
```
ashleyfrieze: It's generally a _bad thing_ to use a `for` loop with a missing clause.
It may be wiser to use:

```
while (row < arr.length) {
   // the code...

   row += rowOffset;
}

File: core-java-modules/core-java-arrays/src/main/java/com/baeldung/array/ArrayIterator.java [hunk 2]
```diff
@@ -0,0 +1,62 @@
...
+			if (col >= 0 && col < arr[row].length) {
+				System.out.print(arr[row][col] + "\t");
+			} else if (rowOffset == 0 && (colOffset > 0 || col < 0)) {
+				break;
+			} else {
+				System.out.print("-\t");
+			}
+
+			if (rowOffset == 0 && colOffset == 0) {
+				break;
```
ashleyfrieze: This might be worth commenting - why do we check for 0 offset?

jaafarnaddaf: We check so that we don't get stuck in an infinite loop. If both of the offsets are set to 0, this means that there no next element.

core-java-modules/core-java-arrays/src/main/java/com/baeldung/array/ArrayIterator.java
jaafarnaddaf: True, will edit

PR #7135, opened 2019-06-15
Code snippet for tutorial on folding  technique (BAEL-2949)

veontomo:


rpvilao:
Since we are providing javadoc let's complete it. Javadoc also for the variables, etc. Also let's format the code, there are files with 2 line feeds instead of one and a lot of line feeds in the end as well.

PR #7146, opened 2019-06-20
Bael 3045 validate date

chandra1123:


java-dates-2/src/main/java/com/baeldung/date/validation/DateValidator.java
```diff
@@ -0,0 +1,5 @@
+package com.baeldung.date.validation;
+
+public interface DateValidator {
+   public boolean isValid(String dateStr);
```
KevinGilmore:
```suggestion
   boolean isValid(String dateStr);
```

KevinGilmore: public modifier not needed in interfaces

PR #7174, opened 2019-06-23
[BAEL-14274] - Fixed article code for https://www.baeldung.com/jackso…

amit2103:
…n-annotations


jackson-simple/src/test/java/com/baeldung/jackson/date/Event.java
```diff
@@ -4,17 +4,17 @@
 
 import com.fasterxml.jackson.annotation.JsonFormat;
 
-public class EventWithFormat {
+public class Event {
```
lor6: I think the EventWithFormat is for a different article; instead we should have the EventWithSerializer in the article

jackson-simple/src/test/java/com/baeldung/jackson/date/Event.java
```diff
@@ -4,17 +4,17 @@
...
     public String name;
 
     @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd-MM-yyyy hh:mm:ss")
     public Date eventDate;
 
-    public EventWithFormat() {
+    public Event() {
```
lor6: please check the updated article: https://www.baeldung.com/jackson-annotations 
the Event classes should not be changed here

amit2103: Done, reverted the rename

jackson-simple/src/test/java/com/baeldung/jackson/date/Event.java
amit2103: Done, reverted the rename

PR #7180, opened 2019-06-23
[BAEL-14274] - Fixed article code for https://www.baeldung.com/java-8…

amit2103:
…-lambda-expressions-tips


lor6:
the article mentions using different names for the methods is the recommended approach, so let's keep it that way in the codebase also

PR #7227, opened 2019-06-30
[BAEL-15425] Update Spring Security Database article

rozagerardo:
Jira: http://team.baeldung.com/browse/BAEL-15425
update IP: http://inprogress.baeldung.com/wp-admin/post.php?post=148983&action=edit&classic-editor

### DIFF
--- a/spring-security-mvc-boot/src/main/java/org/baeldung/custom/config/SecurityConfig.java
+++ b/spring-security-mvc-boot/src/main/java/org/baeldung/custom/config/SecurityConfig.java
@@ -1,57 +1,12 @@
 package org.baeldung.custom.config;
 
-import org.baeldung.custom.security.MyUserDetailsService;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
-import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.security.authentication.dao.DaoAuthenticationProvider;
-import org.springframework.security.config.annotation.authentication.builders.AuthenticationManagerBuilder;
-import org.springframework.security.config.annotation.web.builders.HttpSecurity;
-import org.springframework.security.config.annotation.web.builders.WebSecurity;
-import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
-import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
 import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
 import org.springframework.security.crypto.password.PasswordEncoder;
 
 @Configuration
-@EnableWebSecurity
-@ComponentScan("org.baeldung.security")
-public class SecurityConfig extends WebSecurityConfigurerAdapter {
-
-    @Autowired
-    private MyUserDetailsService userDetailsService;
-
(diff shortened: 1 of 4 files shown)


spring-security-mvc-boot/src/main/java/org/baeldung/custom/web/MainController.java
```diff
@@ -52,7 +55,17 @@ public Foo findFooByName(@RequestParam final String name) {
...
     public Organization findOrgById(@PathVariable final long id) {
-        return organizationRepository.findById(id).orElse(null);
+        return organizationRepository.findById(id)
+            .orElse(null);
+    }
+
+    @PreAuthorize("hasPermission(#id, 'Foo', 'read')")
+    @RequestMapping(method = RequestMethod.GET, value = "/user")
+    @ResponseBody
+    public MyUserPrincipal retrieveUserDetails() {
```
jzheaux: You should be able to replace this with

```suggestion
    public MyUserPrincipal retreiveUserDetails(@AuthenticationPrincipal MyUserPrincipal principal)
```

spring-security-mvc-boot/src/main/java/org/baeldung/custom/web/MainController.java
```diff
@@ -52,7 +55,17 @@ public Foo findFooByName(@RequestParam final String name) {
     @RequestMapping(method = RequestMethod.GET, value = "/organizations/{id}")
     @ResponseBody
     public Organization findOrgById(@PathVariable final long id) {
-        return organizationRepository.findById(id).orElse(null);
+        return organizationRepository.findById(id)
+            .orElse(null);
+    }
+
+    @PreAuthorize("hasPermission(#id, 'Foo', 'read')")
+    @RequestMapping(method = RequestMethod.GET, value = "/user")
```
jzheaux: While we are busy simplifying, let's please change these to:

```suggestion
    @GetMapping("/user")
```

spring-security-mvc-boot/src/main/java/org/baeldung/custom/config/SecurityConfig.java
```diff
@@ -1,56 +1,20 @@
...
-
-    @Bean
-    public DaoAuthenticationProvider authenticationProvider() {
-        final DaoAuthenticationProvider authProvider = new DaoAuthenticationProvider();
-        authProvider.setUserDetailsService(userDetailsService);
-        authProvider.setPasswordEncoder(encoder());
-        return authProvider;
+    protected void configure(HttpSecurity http) throws Exception {
+        super.configure(http);
+        http.csrf()
```
jzheaux: A common mistake is when an application disables this, but still exposes their API over HTTP basic. Since the browser will automatically resend HTTP basic creds, it can't remain a public API and have CSRF disabled.

If we leave it disabled, let's make sure the article briefly calls out why we are justified in doing it in the article (and please remember that readers will be tempted to add this for the convenience).

On the other hand, it's pretty easy to leave this enabled; you can use the `SecurityMockMvcRequestPostProcessors#csrf()` post processor in `MockMvc` to add a test one:

```java
this.mvc.perform(get("/").with(csrf()));
```

rozagerardo:
Thanks for the improvement suggestions @jzheaux . I agree on all of the points. I now addressed all the comments :)

PR #7387, opened 2019-07-22
Adding Metrics AspectJ article code changes

norbertoritzmann:
PR for the article: http://jira.baeldung.com/browse/BAEL-3050

### DIFF
--- a/metrics/src/main/java/com/baeldung/metrics/aspectj/MetricsAspectJMain.java
+++ b/metrics/src/main/java/com/baeldung/metrics/aspectj/MetricsAspectJMain.java
@@ -22,9 +22,9 @@ public class MetricsAspectJMain {
     }
 
     private static void startReport() {
-        SharedMetricRegistries.add(ObjectRunner.REGISTRY_NAME, registry);
+        SharedMetricRegistries.add(ObjectRunner.REGISTRY_NAME, REGISTRY);
 
-        ConsoleReporter reporter = ConsoleReporter.forRegistry(registry)
+        ConsoleReporter.forRegistry(REGISTRY)
                 .convertRatesTo(TimeUnit.SECONDS)
                 .convertDurationsTo(TimeUnit.MILLISECONDS)
                 .outputTo(new PrintStream(System.out))
--- a/metrics/src/main/java/com/baeldung/metrics/aspectj/MetricsAspectJMain.java
+++ b/metrics/src/main/java/com/baeldung/metrics/aspectj/MetricsAspectJMain.java
@@ -7,7 +7,7 @@ import java.io.PrintStream;
 import java.util.concurrent.TimeUnit;
 
 public class MetricsAspectJMain {
-    static final MetricRegistry registry = new MetricRegistry();
+    private static final MetricRegistry REGISTRY = new MetricRegistry();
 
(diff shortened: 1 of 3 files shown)


metrics/src/main/java/com/baeldung/metrics/aspectj/MetricsAspectJMain.java
```diff
@@ -0,0 +1,34 @@
+package com.baeldung.metrics.aspectj;
+
+import com.codahale.metrics.ConsoleReporter;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.SharedMetricRegistries;
+import java.io.PrintStream;
+import java.util.concurrent.TimeUnit;
+
+public class MetricsAspectJMain {
+    static final MetricRegistry registry = new MetricRegistry();
```
ashleyfrieze: this should probably be `private`. As a static reference, it should also be called `REGISTRY` or `METER_REGISTRY` as constants should be in uppercase.

metrics/src/main/java/com/baeldung/metrics/aspectj/MetricsAspectJMain.java
```diff
@@ -0,0 +1,34 @@
...
+        ObjectRunner runner = new ObjectRunner();
+
+        for (int i = 0; i < 5; i++) {
+            runner.run();
+        }
+
+        Thread.sleep(3000L);
+    }
+
+    static void startReport() {
```
ashleyfrieze: Why is this package private? It should be `private`

metrics/src/main/java/com/baeldung/metrics/aspectj/MetricsAspectJMain.java
```diff
@@ -0,0 +1,34 @@
...
+        SharedMetricRegistries.add(ObjectRunner.REGISTRY_NAME, registry);
+
+        ConsoleReporter reporter = ConsoleReporter.forRegistry(registry)
+                .convertRatesTo(TimeUnit.SECONDS)
+                .convertDurationsTo(TimeUnit.MILLISECONDS)
+                .outputTo(new PrintStream(System.out))
+                .build();
+        reporter.start(3, TimeUnit.SECONDS);
```
ashleyfrieze: I'm not sure the _explanatory variable_ `reporter` adds much here... how about going from `.build()` to `.report()` fluently, rather than breaking the fluent chain with a temp variable that's only used straight afterwards.

metrics/src/main/java/com/baeldung/metrics/aspectj/ObjectRunner.java
```diff
@@ -0,0 +1,23 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(ObjectRunner.class);
+    public static final String REGISTRY_NAME = "ObjectRunner";
+
+    @Timed(name = "timerName")
+    public void run() {
+        logger.info("run");
+        try {
+            Thread.sleep(1000L);
+        } catch (InterruptedException e)
+        {
```
ashleyfrieze: This `{` belongs on the previous line
It's conventional to comment an empty catch block too

```
} catch (InterruptedException e) {
   // ignore this exception
}
```

PR #7396, opened 2019-07-23
BAEL-3153: Added test cases for String toLowerCase and toUpperCase

albanoj2:
The toLowerCase and toUpperCase methods are simple enough that no new classes are needed under the main/ directory. The included tests exercise the changing of case for an English string using the default locale, a foreign string using the default locale, and a foreign string using a provided locale.


java-strings-2/README.MD
```diff
@@ -1,7 +1,7 @@
-## Relevant Articles
-
-- [Java Localization – Formatting Messages](https://www.baeldung.com/java-localization-messages-formatting)
-- [Check If a String Contains a Substring](https://www.baeldung.com/java-string-contains-substring)
-- [Removing Stopwords from a String in Java](https://www.baeldung.com/java-string-remove-stopwords)
-- [Blank and Empty Strings in Java](https://www.baeldung.com/java-blank-empty-strings)
-- [String Initialization in Java](https://www.baeldung.com/java-string-initialization)
+## Relevant Articles
```
KevinGilmore: Did anything actually change in this file? If it's only whitespace or line break changes, then we need to avoid those.

albanoj2: I think that Eclipse formatted this file. It looks like only whitespace changed (it wasn't my intention to change that file at all, but it got added into the commit on accident).

java-strings-2/src/test/java/com/baeldung/string/changecase/ToUpperCaseUnitTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.Locale;
+
+import org.junit.Test;
+
+public class ToUpperCaseUnitTest {
+
+    private static final Locale TURKISH = new Locale("tr");
+    private String name = "John Doe";
+    private String foreignUppercase = "\u0069";
```
KevinGilmore: Shouldn't this be called foreignLowercase?

albanoj2: Yes, you are right, it should be `foreignLowercase`. I have made the correction and pushed in the updated test cases.

PR #7399, opened 2019-07-23
BAEL-3005 | Validating List objects in Spring

pallavi-priyadarshani:


spring-mvc-simple-2/pom.xml
```diff
@@ -18,6 +18,10 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-web</artifactId>
         </dependency>
+        <dependency>
```
KevinGilmore: I don't think we should be introducing Lombok into the mix for such a trivial model class and where Lombok is not the focus of the article.

pallavi-priyadarshani: yeah, sure. I'll remove it.

PR #7451, opened 2019-07-30
BAEL-3116: Finding the Least Common Multiple in Java

tapankavasthi:
- Add a tutorial to implement algorithms used for computing
  LCM of two numbers.
- List of approaches covered:
  - Simple Method
  - Prime Factorization Method
  - Euclidean Algorithm
  - BigInteger Class
  - Lambda Implementation
- Reference: [BAEL-3116](http://jira.baeldung.com/browse/BAEL-3116)

### DIFF
--- a/java-numbers-2/src/main/java/com/baeldung/lcm/SimpleAlgorithm.java
+++ b/java-numbers-2/src/main/java/com/baeldung/lcm/SimpleAlgorithm.java
@@ -1,0 +1,18 @@
+package com.baeldung.lcm;
+
+public class SimpleAlgorithm {
+    public static int lcm(int number1, int number2) {
+        if (number1 == 0 || number2 == 0) {
+            return 0;
+        }
+        int absNumber1 = Math.abs(number1);
+        int absNumber2 = Math.abs(number2);
+        int absHigherNumber = Math.max(absNumber1, absNumber2);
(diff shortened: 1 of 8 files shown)


java-numbers-2/src/main/java/com/baeldung/lcm/SimpleAlgorithm.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class SimpleAlgorithm {
+    public static int lcm(int number1, int number2) {
+        if (number1 == 0 || number2 == 0) {
+            return 0;
+        }
+        int absNumber1 = Math.abs(number1);
+        int absNumber2 = Math.abs(number2);
+        int lcm = Math.max(absNumber1, absNumber2);
+        while (lcm % Math.min(absNumber1, absNumber2) != 0) {
```
KevinGilmore: Observation: absNumber1 and absNumber2 do not change inside the loop. It would be more efficient to store the Math.min and Math.max results in local variables before entering the loop, rather than invoking these methods in each loop iteration.

tapankavasthi: @KevinGilmore Thanks for pointing this out. I have changed it now.

PR #7465, opened 2019-08-01
[BAEL-16045] - Check Article Code Matches GitHub for https://www.bael…

amit2103:
…dung.com/java-try-with-resources article


lor6:
the code for the article has been moved to core-java-exceptions so we need to add this code there

PR #7476, opened 2019-08-02
Bael 2688

pcoates33:


spring-amqp/pom.xml
```diff
@@ -3,37 +3,26 @@
...
     <name>spring-amqp</name>
-    <packaging>jar</packaging>
     <description>Introduction to Spring-AMQP</description>
 
     <parent>
+        <artifactId>parent-boot-2</artifactId>
         <groupId>com.baeldung</groupId>
-        <artifactId>parent-modules</artifactId>
-        <version>1.0.0-SNAPSHOT</version>
+        <version>0.0.1-SNAPSHOT</version>
```
rpvilao: is that intended?

pcoates33: Yes : the sample code is using Spring Boot 2. If I don't use the parent-boot-2 then I think 'd need to add the dependency management stuff to my pom. The version in the parent-boot-2 is 0.0.1-SNAPSHOT, so I assume I need to match the version here. I'm no maven expert, so more than happy to change this if there's a better way.

PR #7515, opened 2019-08-05
BAEL-3115

vatsalgosar:
GCD implementation using the Brute Force method, Euclid's Algorithm and Stein's Algorithm


java-math/src/main/java/com/baeldung/algorithms/gcd/GCDImplementation.java
```diff
@@ -0,0 +1,51 @@
+package com.baeldung.algorithms.gcd;
+
+public class GCDImplementation {
+
+    public static int gcdByBruteForce(int n1, int n2) {
+        int gcd = 1;
+        for (int i = 1; i <= n1 && i <= n2; i++)
```
rpvilao: let's use brackets on all the branches (even the single op ones) - this goes for all of the envs, if, for, etc

PR #7537, opened 2019-08-08
[BAEL-16683] - Upgrade parent-boot-1 to the latest version of Boot 1.…

amit2103:
…5.22

### DIFF
--- a/parent-boot-1/pom.xml
+++ b/parent-boot-1/pom.xml
@@ -55,7 +55,7 @@
 
 	<properties>
 		<rest-assured.version>3.1.0</rest-assured.version>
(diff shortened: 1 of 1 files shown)


parent-boot-1/pom.xml
```diff
@@ -55,7 +55,7 @@
 
 	<properties>
 		<rest-assured.version>3.1.0</rest-assured.version>
-		<spring-boot.version>1.5.19.RELEASE</spring-boot.version>
+		<spring-boot.version>1.5.22.RELEASE</spring-boot.version>
```
jeffzhu2010: test webhook

amit2103: @jeffzhu2010 Can you please elaborate on what to do?

PR #7579, opened 2019-08-14
BAEL-3120 alfred.samanga@gmail.com 

Alfred-Samanga:
Counting words in a String.

### DIFF
--- a/java-strings-3/src/main/java/com/baeldung/string/wordcount/WordCounter.java
+++ b/java-strings-3/src/main/java/com/baeldung/string/wordcount/WordCounter.java
@@ -3,11 +3,14 @@ package com.baeldung.string.wordcount;
 import java.util.StringTokenizer;
 
 public class WordCounter {
+    static final int WORD = 0;
+    static final int SEPARATOR = 1;
+
     public static int countWordsUsingRegex(String arg) {
         if (arg == null) {
             return 0;
         }
-        final String[] words = arg.split("\\pP|\\s+");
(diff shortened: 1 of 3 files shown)


java-strings-3/src/main/java/com/baeldung/string/wordcount/WordCounter.java
```diff
@@ -0,0 +1,48 @@
+package com.baeldung.string.wordcount;
+
+import java.util.StringTokenizer;
+
+public class WordCounter {
+    public static void main(String[] args) {
```
rpvilao: why? maybe we can remove this one.

java-strings-3/src/test/java/com/baeldung/string/wordcount/WordCountUnitTest.java
```diff
@@ -0,0 +1,24 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Test;
+public class WordCountUnitTest {
+    private String string1 = "This is a test sentence with eight words";
+    private String string2 = "This#is%a test sentence with eight       words";
+
+    @Test
+    public void givenStringWith8Words_whenUsingRegexCount_ThenResultEqual8() {
+        assertThat(WordCounter.countWordsUsingRegex(string2) == 8).isTrue();
```
rpvilao: not the right way to assert integers, use the method that compares integers

PR #7622, opened 2019-08-21
[BAEL-3149] Java 'private' access modifier

NickTononi:


core-java-modules/core-java-lang-oop-2/src/main/java/com/baeldung/core/modifiers/Employee.java
```diff
@@ -0,0 +1,50 @@
...
+    public void changeId(String customId) {
+        if (customId.endsWith("_ID")) {
+            this.privateId = customId;
+        } else {
+            this.privateId = customId + "_ID";
+        }
+    }
+
+    public String getId() {
+        return this.privateId;
```
KevinGilmore:
```suggestion
        return privateId;
```

core-java-modules/core-java-lang-oop-2/src/main/java/com/baeldung/core/modifiers/Employee.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public boolean isManager() {
+        return this.manager;
```
KevinGilmore:
```suggestion
        return manager;
```

PR #7628, opened 2019-08-22
Drop last 4 bytes of computed SHA-1 hash

### LINKED ISSUE #7598: UUIDGenerator has wrong length check for UUID type5
The `type5UUIDFromBytes` method uses SHA-1 to hash information which means the result should be 20 bytes. However, the `constructType5UUID` method it called checks whether the length of the byte array to be equal to 16 which should be the result when using MD5.

Philippus:
Computation of a SHA-1 hash will generate 20 bytes. Applying this PR will drop the last 4 (they are not used for generating the UUID). This will make the assertion in the `constructType5UUID`-method pass. Fixes #7598. Also fixes a wrong message for an exception.


eric-martin:
@Philippus - Thanks for the PR. Looks like you are missing an import statement causing a build problem. I'll see if I can fix it.

Philippus:
ah, thanks! I missed that bit.

PR #7639, opened 2019-08-23
BAEL-3220 Parsing an XML File Using StAX

pazis:
http://inprogress.baeldung.com/wp-admin/post.php?post=158264&action=edit&classic-editor


pazis:
Fixed.

PR #7672, opened 2019-08-29
[BAEL-3221] unmarshalling dates using JAXB - examples

catalin-burcea:


jaxb/src/main/java/com/baeldung/jaxb/dateunmarshalling/Book.java
```diff
@@ -0,0 +1,31 @@
...
+public class Book {
+
+    @XmlElement(name = "title", required = true)
+    private String title;
+
+    @XmlElement(name = "published", required = true)
+    private XMLGregorianCalendar published;
+
+    public Book() {
+    }
```
ashleyfrieze: Do we need this constructor? There are no other constructors, so you get the default one _for free_.

jaxb/src/main/java/com/baeldung/jaxb/dateunmarshalling/Book2.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @XmlElement(name = "title", required = true)
+    private String title;
+
+    @XmlElement(name = "published", required = true)
+    @XmlJavaTypeAdapter(DateAdapter.class)
+    private Date published;
+
+    public Book2() {
+    }
```
ashleyfrieze: Same point as above

jaxb/src/main/java/com/baeldung/jaxb/dateunmarshalling/Book3.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @XmlElement(name = "title", required = true)
+    private String title;
+
+    @XmlElement(name = "published", required = true)
+    @XmlJavaTypeAdapter(LocalDateTimeAdapter.class)
+    private LocalDateTime published;
+
+    public Book3() {
+    }
```
ashleyfrieze: and again :)

jaxb/src/main/java/com/baeldung/jaxb/dateunmarshalling/DateAdapter.java
```diff
@@ -0,0 +1,25 @@
+package com.baeldung.jaxb.dateunmarshalling;
+
+import javax.xml.bind.annotation.adapters.XmlAdapter;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public class DateAdapter extends XmlAdapter<String, Date> {
+
+    private final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
```
ashleyfrieze: This should probably be `static` and have a BLOCK CAPITAL name.

jaxb/src/main/java/com/baeldung/jaxb/dateunmarshalling/DateAdapter.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Override
+    public String marshal(Date v) {
+        synchronized (dateFormat) {
```
ashleyfrieze: I appreciate that this is not threadsafe... would it be safer/simpler to make the `yyyy-MM-dd...` the `private static final` thing and create a new `SimpleDateFormat` object every time?

jaxb/src/main/java/com/baeldung/jaxb/dateunmarshalling/JaxbDateUnmarshalling.java
```diff
@@ -0,0 +1,52 @@
...
+    public static void main(String[] args) {
+        try {
+            Book book = unmarshalDates(getInputStream(DEFAULT_DATE_UNMARSHALLING_FILE));
+            Book2 book2 = unmarshalDatesUsingCustomXmlAdapter(getInputStream(CUSTOM_DATE_UNMARSHALLING_FILE));
+            Book3 book3 = unmarshalDatesUsingJava8(getInputStream(CUSTOM_DATE_UNMARSHALLING_FILE));
+            System.out.println(book);
+            System.out.println(book2);
+            System.out.println(book3);
+        } catch (JAXBException e) {
+            e.printStackTrace();
```
ashleyfrieze: This is IDE generated code... could we avoid the try/catch here and just let `main` throw an Exception?

Or would be better illustrating all this using unit tests, rather than a main method?

PR #7701, opened 2019-09-01
BAEL-3190 - Working with Web Services in Groovy

SmartyAnsh:


core-groovy-2/pom.xml
```diff
@@ -48,6 +48,11 @@
             <version>${spock-core.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+        	<groupId>com.github.groovy-wslite</groupId>
```
rpvilao: fix formatting

SmartyAnsh: resolved

PR #7709, opened 2019-09-01
BAEL-3142 Removing an element from an ArrayList

dcalap:


core-java-modules/core-java-collections-array-list/src/main/java/com/baeldung/java/list/RemoveFromList.java
```diff
@@ -0,0 +1,40 @@
+package com.baeldung.java.list;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+public class RemoveFromList {
```
rpvilao: fix indentation in the class

core-java-modules/core-java-collections-array-list/src/main/java/com/baeldung/java/list/RemoveFromList.java
```diff
@@ -0,0 +1,41 @@
...
+
+                // Remove with an element
+                sports.remove("Baseball");
+
+                // Iterator remove method
+                Iterator<String> iterator = sports.iterator();
+                while (iterator.hasNext()) {
+                        if (iterator.next().equals("Boxing")) {
+                                iterator.remove();
+                        }
```
manojsharma27: Add a `break;` after line no 29. No need to iterate further once 'Boxing' is removed.

core-java-modules/core-java-collections-array-list/src/main/java/com/baeldung/java/list/RemoveFromList.java
```diff
@@ -0,0 +1,41 @@
...
+                }
+
+                // ArrayList removeIf method (Java 8)
+                for (int i = 0; i < sports.size(); i++) {
```
manojsharma27: This loop is not required at all. As per comment at line 33, only the statement `sports.removeIf(p -> p.equals("Cycling"));` is enough.

dcalap:
Totally agree with the changes. Updated code.

PR #7714, opened 2019-09-02
BAEL-2974 Logger vs System.out

dhruba619:


logging-modules/log4j/src/main/java/com/baeldung/log4j/Log4jComparisonSysout.java
```diff
@@ -0,0 +1,18 @@
+package com.baeldung.log4j;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.PrintStream;
+
+public class Log4jComparisonSysout {
+
+	public static void main(String[] args) throws FileNotFoundException {
```
KevinGilmore: please change indentation to use spaces, not tabs

dhruba619: Updated. Changed to spaces.

PR #7733, opened 2019-09-06
PR-BAEL-2816

yavuztas:
Add source code for article BAEL-2816


patterns/design-patterns-2/src/main/java/com/baeldung/hexagonal/AccountApplication.java
```diff
@@ -0,0 +1,21 @@
+package com.baeldung.hexagonal;
```
ashleyfrieze: This stuff isn't part of the article. Have you got your PR coming from the wrong branch?

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/wherejointable/Group.java
```diff
@@ -0,0 +1,85 @@
...
+    @GeneratedValue
+    private Long id;
+
+    private String name;
+
+    @ManyToMany(mappedBy = "groups")
+    private List<User> users = new ArrayList<>();
+
+    public Group() {
+
```
ashleyfrieze: Let's remove the redundant blank line, or put in a comment about why there's a default constructor as well as a parameterized one.

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/wherejointable/Group.java
```diff
@@ -0,0 +1,85 @@
...
+        return "Group [name=" + name + "]";
+    }
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + ((name == null) ? 0 : name.hashCode());
+        return result;
+    }
```
ashleyfrieze: This is a rather messy hashcode... Do we need hashcode and equals for this implementation? If so, please generate using the more modern Java 7+ template.

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/wherejointable/Group.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Override
+    public boolean equals(Object obj) {
```
ashleyfrieze: Although this is generated code, it probably should use `{}` with if statements. Again, this looks like a pre-Java 7 template for creating an equals method.

patterns/design-patterns-2/src/main/java/com/baeldung/hexagonal/AccountApplication.java
yavuztas: Sorry, it seems I accidentally picked also my evaluation article codes into my branch which never merged but stayed in my fork. I will remove them.

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/wherejointable/Group.java
yavuztas: You are right, we do not need to define default constructor here so I will remove them for both Group and User classes.

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/wherejointable/Group.java
yavuztas: Although removing equals and hashCode doesn't break the tests here, it is a good practice to implement them, to define the equality of entities independent from synthetic identifiers. It might cause unexpected behaviors for example when we deal with collections like `.contains()` or adding an element to `Set` structures, etc.

So, I propose refactoring them to Java 7+ way and do not remove. What do you think?

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/wherejointable/Group.java
yavuztas: Same for here, refactoring them to Java 7+ way.

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/wherejointable/Group.java
ashleyfrieze: I think the issue here is if they are not needed by anything in our example code, then they're just noise. There's also a decision to justify which is why the hashcode does not hash all properties.

My general rule is that if it's hard to explain, try removing it ;)

persistence-modules/hibernate-mapping/src/test/java/com/baeldung/hibernate/wherejointable/HibernateWhereJoinTableIntegrationTest.java
```diff
@@ -0,0 +1,118 @@
...
+    @After
+    public void tearDown() {
+        session.getTransaction().commit();
+        session.close();
+    }
+
+    @AfterClass
+    public static void afterTests() {
+        sessionFactory.close();
+    }
```
ashleyfrieze: I think this test class has its methods in an unconventional order. It's most common for the order to be something like:

- befores
- afters
- tests
- private helper methods

This class seems to be

- a helper method
- befores
- tests
- afters

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/wherejointable/Group.java
yavuztas: Sure, I removed them because we don't explain their effects here. I am more of a developer rather than an educator so, you have the point. But, I am getting closer I hope :)

persistence-modules/hibernate-mapping/src/test/java/com/baeldung/hibernate/wherejointable/HibernateWhereJoinTableIntegrationTest.java
yavuztas: I changed them to the conventional order as you said.

PR #7743, opened 2019-09-07
BAEL-3237 AutoValue Defensive Copies

gilday:


code-generation/src/test/java/com/baeldung/autovalue/PersonUnitTest.java
```diff
@@ -0,0 +1,37 @@
...
+            .favoriteMovies(originalFavoriteMoviesList)
+            .build();
+
+        // WHEN modify original list
+        originalFavoriteMoviesList.add("Friday");
+
+        // THEN Person remains unaffected
+        assertFalse(person.favoriteMovies()
+            .contains("Friday"));
+        assertEquals(person.favoriteMovies()
```
KevinGilmore: Let's swap the order of arguments to the assertEquals call.

gilday: Done!

PR #7770, opened 2019-09-10
dupirefr/dupire.francois+pro@gmail.com [BAEL-3128] Breadth-First Search Algorithm in Java

dupirefr:
* Added Tree and algorithm tests for trees

* Implemented algorithm for trees

* Added algorithm for graphs

* Bidirectional graph

* Minor updates

* Readibility changes

### DIFF
--- a/algorithms-miscellaneous-3/src/main/java/com/baeldung/algorithms/breadthfirstsearch/BreadthFirstSearchAlgorithm.java
+++ b/algorithms-miscellaneous-3/src/main/java/com/baeldung/algorithms/breadthfirstsearch/BreadthFirstSearchAlgorithm.java
@@ -1,0 +1,61 @@
+package com.baeldung.algorithms.breadthfirstsearch;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
+
+public class BreadthFirstSearchAlgorithm {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(BreadthFirstSearchAlgorithm.class);
+
+    public static <T> Optional<Tree<T>> search(T value, Tree<T> root) {
+        Queue<Tree<T>> queue = new ArrayDeque<>();
+        queue.add(root);
+
+        return searchTreeQueue(value, queue);
+    }
+
+    private static <T> Optional<Tree<T>> searchTreeQueue(T value, Queue<Tree<T>> queue) {
+        Tree<T> currentNode;
+        while (!queue.isEmpty()) {
+            currentNode = queue.remove();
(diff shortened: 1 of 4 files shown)


algorithms-miscellaneous-3/src/test/java/com/baeldung/algorithms/breadthfirstsearch/BreadthFirstSearchAlgorithmUnitTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private Node<Integer> start;
+    private Node<Integer> firstNeighbour;
+    private Node<Integer> firstNeighbourNeighbour;
+    private Node<Integer> secondNeighbour;
+
+    @BeforeEach
+    void beforeEach() {
+        initTree();
+        initNode();
```
KevinGilmore: Since we're only using one or the other in each test, let's just initialize the appropriate one when we need it instead of always initializing both.

algorithms-miscellaneous-3/src/main/java/com/baeldung/algorithms/breadthfirstsearch/BreadthFirstSearchAlgorithm.java
```diff
@@ -0,0 +1,61 @@
...
+
+            if (currentNode.getValue().equals(value)) {
+                return Optional.of(currentNode);
+            } else {
+                queue.addAll(currentNode.getChildren());
+            }
+        }
+
+        return Optional.empty();
+    }
```
KevinGilmore: Why not just put all this in the search method?

algorithms-miscellaneous-3/src/main/java/com/baeldung/algorithms/breadthfirstsearch/BreadthFirstSearchAlgorithm.java
```diff
@@ -0,0 +1,61 @@
...
+                return Optional.of(currentNode);
+            } else {
+                alreadyVisited.add(currentNode);
+                queue.addAll(currentNode.getNeighbours());
+                queue.removeAll(alreadyVisited);
+            }
+        }
+
+        return Optional.empty();
+    }
```
KevinGilmore: Again, we could do this in a single method.

algorithms-miscellaneous-3/src/main/java/com/baeldung/algorithms/breadthfirstsearch/BreadthFirstSearchAlgorithm.java
dupirefr: Yeah, that doesn't make much sense now that I see it again :-)

PR #7782, opened 2019-09-12
BAEL-3212 - Examples of POJO, Bean and BeanUtils reflection

JonathanTurnock:
@ashleyfrieze - Pull Request for the associated draft article.


spring-core-2/src/main/java/com/baeldung/pojo/EmployeePojo.java
```diff
@@ -0,0 +1,27 @@
...
+        this.startDate = startDate;
+    }
+
+    String name() {
+        return this.firstName + " " + this.lastName;
+    }
+
+    LocalDate getStart() {
+        return this.startDate;
+    }
```
ashleyfrieze: These methods should be `public` rather than `package-private`

ashleyfrieze: Or is the point that they're not `public` so it's not a bean?

JonathanTurnock: @ashleyfrieze I have corrected these to be public now.

core-java-modules/core-java-lang-2/src/main/java/com/baeldung/pojo/EmployeeBean.java
```diff
@@ -0,0 +1,47 @@
...
+public class EmployeeBean {
+
+    private String firstName;
+
+    private String lastName;
+
+    private LocalDate startDate;
+
+    public EmployeeBean() {
+
```
ashleyfrieze: `MINOR` - we don't need the blank line here

core-java-modules/core-java-lang-2/src/main/java/com/baeldung/pojo/EmployeeBean.java
```diff
@@ -0,0 +1,47 @@
+package com.baeldung.pojo;
+
+import java.time.LocalDate;
+
+public class EmployeeBean {
```
ashleyfrieze: In the article, you say that beans must be Serializable... this isn't - or at least it doesn't implement `Serializable`

PR #7809, opened 2019-09-16
[BAEL-3076] ignoring unmapped properties with Mapstruct

catalin-burcea:


mapstruct/src/test/java/com/baeldung/mapper/unmappedproperties/DocumentMapperUnitTest.java
```diff
@@ -0,0 +1,45 @@
...
+        entity.setId(1);
+        entity.setTitle("Price 13-42");
+        entity.setText("List of positions.......");
+        entity.setModificationTime(new Date());
+
+        DocumentDTO dto = DocumentMapper.INSTANCE.documentToDocumentDTO(entity);
+
+        assertEquals(dto.getId(), entity.getId());
+        assertEquals(dto.getTitle(), entity.getTitle());
+        assertEquals(dto.getText(), entity.getText());
```
ashleyfrieze: The assertion should be `assertEquals(expected, actual)` - I think this is reversed.

Perhaps switch to `assertJ` as it's clearer `assertThat(dto.getId()).isEqualTo(entity.getId())`

Though equals == equals :) the reader who understands that JUnit assert equals goes expected first will stumble on examples like this.

PR #7815, opened 2019-09-17
BAEL-2988 Using Enums as Request Parameters in Spring

wugangca:


spring-rest-simple/src/main/java/com/baeldung/Application.java
```diff
@@ -8,7 +8,7 @@
 
 @EnableAutoConfiguration
 @SpringBootApplication
-@ComponentScan("com.baeldung.cors")
+//@ComponentScan("com.baeldung.cors")
```
wugangca: Need to comment it out so that Spring can load the MvcConfig in the com.baeldung.config package

KevinGilmore: Instead of commenting out the line, could we simply remove it? I'm guessing that there's no longer a need for it, since we already have the EnableAutoConfiguration and SpringBootApplication annotations.

wugangca: Removed.

PR #7823, opened 2019-09-18
BAEL-3085 Using a slash character in Spring URLs

asturcon:
IP: http://inprogress.baeldung.com/wp-admin/post.php?post=161454&action=edit&classic-editor


spring-boot-mvc-2/pom.xml
```diff
@@ -1,68 +1,73 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
KevinGilmore: We generally want to avoid changing only whitespaces like changing the indent from tabs to spaces or vice versa on an existing file. Please restore this to its original indentation, and then reapply any other changes you have made, if any.

asturcon: done

PR #7872, opened 2019-09-24
BAEL-3321 Flogger Fluent Logging

Alfred-Samanga:
Added support for Slf4j and Log4j backends


logging-modules/flogger/pom.xml
```diff
@@ -17,11 +17,50 @@
...
             <scope>runtime</scope>
         </dependency>
+
+        <dependency>
+            <groupId>com.google.flogger</groupId>
+            <artifactId>flogger-slf4j-backend</artifactId>
+            <version>0.4</version>
+        </dependency>
+
+<!--        <dependency>-->
```
maibin: Could you please let me know why this part is left in code? If it is not necessary, let's remove it. Make sure also to run integration tests and check if they all pass.

logging-modules/flogger/src/main/resources/log4j.xml
```diff
@@ -0,0 +1,99 @@
+<!DOCTYPE log4j:configuration SYSTEM "log4j.dtd" >
```
maibin: Why we need this file?

logging-modules/flogger/src/test/java/com/baeldung/flogger/FloggerIntegrationTest.java
```diff
@@ -3,6 +3,7 @@
 import com.google.common.flogger.FluentLogger;
```
maibin: let's remove unused tests

logging-modules/flogger/pom.xml
Alfred-Samanga: We added a section in the article about using other backends such as Slf4j and Log4j and using their existing configuration files. The dependencies that are commended out are for Log4j. This is so that the reader of the article can easily reference these dependencies. We could uncomment them, but still, I believe they are necessary for the reader.

logging-modules/flogger/src/main/resources/log4j.xml
Alfred-Samanga: This file we probably don't need, so I have removed it. I was using it to test that existing configurations actually work, but it's not necessary for the flow of the article

logging-modules/flogger/src/test/java/com/baeldung/flogger/FloggerIntegrationTest.java
Alfred-Samanga: If it's because of the import _import org.junit.Ignore_, I have removed this import. Currently, there are no unused tests.

PR #7883, opened 2019-09-25
(martinvw@gmail.com) hexagonal architecture

martinvw:
A short introduction to the hexagonal architecture

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -463,6 +463,7 @@
 
                 <module>hazelcast</module>
                 <module>helidon</module>
+				<module>hexagonal-architecture</module>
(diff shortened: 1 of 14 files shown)


pom.xml
```diff
@@ -463,6 +463,7 @@
 
                 <module>hazelcast</module>
                 <module>helidon</module>
+				<module>hexagonal-architecture</module>
```
martinvw: I see that the main pom.xml already contains a mix of tabs and spaces, around this part of the code it's all spaces. But around line 1204 it's all tabs. What would you prefer me to do here?

PR #7898, opened 2019-09-28
BAEL-1363 Fallback for Zuul Routes

kamleshkr:
BAEL-1363 Fallback for Zuul Routes

### DIFF
--- a/spring-cloud/pom.xml
+++ b/spring-cloud/pom.xml
@@ -25,7 +25,7 @@
         <module>spring-cloud-zookeeper</module>
         <module>spring-cloud-gateway</module>
         <module>spring-cloud-stream</module>
-		<module>spring-cloud-stream-starters/twitterhdfs</module>
+        <module>spring-cloud-stream-starters/twitterhdfs</module>
         <module>spring-cloud-connectors-heroku</module>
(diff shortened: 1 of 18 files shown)


spring-cloud/pom.xml
```diff
@@ -35,9 +35,10 @@
         <module>spring-cloud-archaius</module>
         <module>spring-cloud-functions</module>        
         <module>spring-cloud-vault</module>
-		<!-- <module>spring-cloud-security</module> --> <!-- Fixing in BAEL-10887 -->
-		<module>spring-cloud-task</module>
+        <!-- <module>spring-cloud-security</module> --> <!-- Fixing in BAEL-10887 -->
+        <module>spring-cloud-task</module>
         <module>spring-cloud-zuul</module>
+        <module>spring-cloud-zuul-fallback</module>
```
rpvilao: instead of creating a new module, please check first if we can put everything inside spring-cloud-zuul, and isolate with a java package.

kamleshkr: Hi @rpvilao, as discussed and agreed over JIRA, a new module is required. However, I have updated the ports of the application as suggested by Josh.

PR #7919, opened 2019-10-01
[BAEL-3319] - Documenting a Spring REST API using OpenAPI

harshavs:


spring-boot-springdoc/src/main/java/com/baeldung/springdoc/controller/BookController.java
```diff
@@ -0,0 +1,76 @@
...
+    private BookRepository repository;
+
+    @GetMapping("/{id}")
+    public Book findById(@PathVariable long id) {
+        return repository.findById(id)
+            .orElseThrow(() -> new BookNotFoundException());
+    }
+
+    @RequestMapping(method = RequestMethod.GET, value = "/")
+    @ResponseBody
```
KevinGilmore: Also when using RestController annotation, we don't need ResponseBody annotation.

spring-boot-springdoc/src/main/java/com/baeldung/springdoc/repository/BookRepository.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Optional;
+
+import org.springframework.stereotype.Repository;
+
+import com.baeldung.springdoc.model.Book;
+
+@Repository
+public class BookRepository {
+
+    private List<Book> books = new ArrayList<>();
```
KevinGilmore: Obviously this is not the main point of the article, but I would probably use a Map instead of a List here.

spring-webflux-amqp/pom.xml
```diff
@@ -62,6 +62,12 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-integration</artifactId>
 		</dependency>
+
```
KevinGilmore: And yes, we should revert this change.

spring-boot-springdoc/pom.xml
```diff
@@ -0,0 +1,111 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<groupId>com.baeldung</groupId>
+	<artifactId>spring-boot-springdoc</artifactId>
+	<version>0.0.1-SNAPSHOT</version>
+	<name>spring-boot-springdoc</name>
+	<description>Project for Springdoc integration</description>
+	<packaging>war</packaging>
```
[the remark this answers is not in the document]
harshavs: It was jar originally. I forgot to change it back after my testing.

spring-boot-springdoc/src/main/java/com/baeldung/springdoc/controller/BookController.java
```diff
@@ -0,0 +1,72 @@
...
+        return book;
+    }
+
+    @RequestMapping(method = RequestMethod.HEAD, value = "/")
+    @ResponseStatus(HttpStatus.OK)
+    public Book headBook() {
+        return new Book();
+    }
+
+    @RequestMapping(method = RequestMethod.DELETE, value = "/{id}")
```
KevinGilmore: You should be able to use DeleteMapping here.

spring-boot-springdoc/pom.xml
```diff
@@ -0,0 +1,111 @@
...
+        <relativePath>../parent-boot-2</relativePath>
+    </parent>
+
+    <properties>
+        <java.version>1.8</java.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
```
KevinGilmore: If you're already using a specific starter, like spring-boot-starter-web, I don't think you have to explicitly include just plain spring-boot-starter.

PR #7948, opened 2019-10-04
Feature/pre cleanup

lor6:


javaxval/src/main/resources/META-INF/services/javax.validation.valueextraction.ValueExtractor
```diff
@@ -1 +1 @@
-org.baeldung.valueextractors.ProfileValueExtractor
\ No newline at end of file
+org.baeldung.javaxval.container.validation.valueextractors.ProfileValueExtractor
```
martinvw: @lor6 This exact fragment is used in an article, do we want to update it?

lor6: yep, I see the package is specifically mentioned in the article; I'll add a note in the JIRA to update it after the merge

martinvw: Article: https://www.baeldung.com/bean-validation-container-elements

Search for `Finally, we need to register the class`

PR #7971, opened 2019-10-06
BAEL 3320 JCommander

priyank-sriv:


libraries-2/src/main/java/com/baeldung/jcommander/usagebilling/cli/UsageBasedBilling.java
```diff
@@ -0,0 +1,86 @@
...
+            main.parse(args);
+            parsedCmdStr = main.getParsedCommand();
+            
+            switch (parsedCmdStr) {
+                case Constants.SUBMIT_CMD:
+                    if (this.submitUsageCmd.isHelp()) {
+                        getSubCommandHandle(Constants.SUBMIT_CMD).usage();
+                    }
+                    System.out.println("Parsing usage request...");
+                    this.submitUsageCmd.submit();
```
ashleyfrieze: We don't use `this.` unless it's necessary to disambiguate from a parameter. It's just noise otherwise.

libraries-2/src/main/java/com/baeldung/jcommander/usagebilling/cli/UsageBasedBilling.java
```diff
@@ -0,0 +1,86 @@
+package com.baeldung.jcommander.usagebilling.cli;
+
+import com.beust.jcommander.JCommander;
+import com.beust.jcommander.ParameterException;
+import com.beust.jcommander.UnixStyleUsageFormatter;
+
+public class UsageBasedBilling {
+
+    private JCommander main;
```
ashleyfrieze: `main` is a non-descript name for it. You could call it `jcommander` or `commandParser`?

libraries-2/src/main/java/com/baeldung/jcommander/usagebilling/cli/UsageBasedBilling.java
```diff
@@ -0,0 +1,86 @@
...
+    private void setUsageFormatter(String subCommand) {
+        JCommander cmd = getSubCommandHandle(subCommand);
+        cmd.setUsageFormatter(new UnixStyleUsageFormatter(cmd));
+    }
+
+    static class Constants {
+
+        static final String SUBMIT_CMD = "submit";
+        static final String FETCH_CMD = "fetch";
+    }
```
ashleyfrieze: This isn't _wrong_ but it's another thing for the reader to parse. I'd suggest having `private static final` constants at the top of this class, rather than put them inside a package-private inner static class... you can see from the description how much more processing you're expecting the reader to do.

libraries-2/src/main/java/com/baeldung/jcommander/usagebilling/service/DefaultFetchCurrentChargesService.java
```diff
@@ -0,0 +1,74 @@
...
+import static java.util.Arrays.asList;
+import static java.util.Arrays.fill;
+import static java.util.Collections.emptyList;
+import static java.util.concurrent.ThreadLocalRandom.current;
+import static java.util.stream.Collectors.toList;
+
+class DefaultFetchCurrentChargesService implements FetchCurrentChargesService {
+
+    DefaultFetchCurrentChargesService() {
+    }
```
ashleyfrieze: Why do we need a constructor? Won't we get one by default?

libraries-2/src/main/java/com/baeldung/jcommander/usagebilling/service/DefaultSubmitUsageService.java
```diff
@@ -0,0 +1,21 @@
+package com.baeldung.jcommander.usagebilling.service;
+
+import com.baeldung.jcommander.usagebilling.model.UsageRequest;
+
+import java.util.UUID;
+
+class DefaultSubmitUsageService implements SubmitUsageService {
+
+    DefaultSubmitUsageService() {
+    }
```
ashleyfrieze: Do we need this constructor?

libraries-2/src/test/java/com/baeldung/jcommander/usagebilling/cli/FetchCurrentChargesCommandUnitTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private JCommander jc;
+
+    @Before
+    public void setup() {
+        jc = JCommander
+          .newBuilder()
+          .addObject(new FetchCurrentChargesCommand())
+          .build();
+    }
```
ashleyfrieze: This `@Before` method would be better turned into the initializer - e.g.

```
private JCommander jc = JCommander.newBuilder()
          .addObject(new FetchCurrentChargesCommand())
          .build();
```

Note, it's also relatively unnatural to have just a type name at the end of a line, with the static method call on the next. I know it IS convention to break lines at a `.` in fluent calls. It generally looks more like my example than yours.

libraries-2/src/test/java/com/baeldung/jcommander/usagebilling/cli/FetchCurrentChargesCommandUnitTest.java
```diff
@@ -0,0 +1,74 @@
...
+          "-C", "cb898e7a-f2a0-46d2-9a09-531f1cee1839",
+          "-S", "subscriptionA001",
+          "-S", "subscriptionA002",
+          "-S", "subscriptionA003",
+        });
+        
+        assertEquals(
+          asList("subscriptionA001", "subscriptionA002", "subscriptionA003"),
+          cmd.getSubscriptionIds()
+        );
```
ashleyfrieze: You're using the wrong sort of assertion here, I think. If you use `AssertJ` assertThat, or JUnit's `assertThat`, you'll be able to use something like `assertThat(cmd.getSubscriptionIds()).containsExactly("subscriptionA001", "subscriptionA002", "subscriptionA003")`, which is a lot less boilerplate and a lot richer an assertion.

libraries-2/src/test/java/com/baeldung/jcommander/usagebilling/cli/UsageBasedBillingIntegrationTest.java
```diff
@@ -0,0 +1,31 @@
...
+        String[] argv = {
+          "submit",
+          "--customer", "cb898e7a-f2a0-46d2-9a09-531f1cee1839",
+          "--subscription", "subscriptionPQRMN001",
+          "--pricing-type", "PRE_RATED",
+          "--timestamp", "2019-10-03T10:58:00",
+          "--quantity", "7",
+          "--price", "24.56"
+        };
+        new UsageBasedBilling().run(argv);
```
ashleyfrieze: These tests have no assertions in them... I don't know if that's important.

libraries-2/src/test/java/com/baeldung/jcommander/usagebilling/cli/SubmitUsageCommandUnitTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class SubmitUsageCommandUnitTest {
+
+    private static JCommander jc;
+
+    @BeforeClass
+    public static void setup() {
```
ashleyfrieze: I would recommend moving this to instance level with an initializer. As much as you can argue that the JCommander object is not tiny and is immutable, `static` test setup of anything other than constant Strings is a special case and it's not needed here.

libraries-2/src/main/java/com/baeldung/jcommander/usagebilling/cli/UsageBasedBilling.java
priyank-sriv: sure, fixed.

libraries-2/src/main/java/com/baeldung/jcommander/usagebilling/cli/UsageBasedBilling.java
priyank-sriv: addressed

libraries-2/src/main/java/com/baeldung/jcommander/usagebilling/cli/UsageBasedBilling.java
priyank-sriv: Moved them out to top of the class; still kept the access as package-private since they're used as the `Parameters#commandNames` in their respective sub-command classes.

libraries-2/src/main/java/com/baeldung/jcommander/usagebilling/service/DefaultFetchCurrentChargesService.java
priyank-sriv: Yes, missed this while cleaning up... was using a parameterized constructor in an earlier version of this class. Removed.

libraries-2/src/main/java/com/baeldung/jcommander/usagebilling/service/DefaultSubmitUsageService.java
priyank-sriv: Yes, missed this while cleaning up... was using a parameterized constructor in an earlier version of this class. Removed.

libraries-2/src/test/java/com/baeldung/jcommander/usagebilling/cli/FetchCurrentChargesCommandUnitTest.java
priyank-sriv: Moved it out into initializer.
About the formatting - 
I'm using the formatter for IntelliJ from the repo, and this is what I get when I apply it (which I thought was odd as well). Right now, I've fixed it manually.

The formatter behaves a bit oddly for some of the scenarios such as builders (as above), static method calls (`UUID.randomUUID().toString()` all goes in separate lines), etc.

libraries-2/src/test/java/com/baeldung/jcommander/usagebilling/cli/FetchCurrentChargesCommandUnitTest.java
priyank-sriv: fixed

libraries-2/src/test/java/com/baeldung/jcommander/usagebilling/cli/UsageBasedBillingIntegrationTest.java
priyank-sriv: This was more about invoking the commands with the full set of inputs. I've removed the test and instead, added instructions in the main class.

libraries-2/src/test/java/com/baeldung/jcommander/usagebilling/cli/SubmitUsageCommandUnitTest.java
priyank-sriv: makes sense, fixed.

PR #7972, opened 2019-10-06
Feature/bael 3322 jimfs

lukaszrys:


testing-modules/mocks/pom.xml
```diff
@@ -1,55 +1,62 @@
...
-        <javafaker.version>0.15</javafaker.version>
-        <jukito.version>1.5</jukito.version>
-<mockito.version>2.21.0</mockito.version>
-        <easymock.version>3.5.1</easymock.version>
-        <jmockit.version>1.41</jmockit.version>
-    </properties>
-
-</project>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
JonCook: Something is not quite right with the formatting here. Make sure we are using the eclipse formatter from:
https://github.com/eugenp/tutorials/blob/master/eclipse/formatter.xml

testing-modules/mocks/src/main/java/com/baeldung/jimfs/FileRepository.java
```diff
@@ -0,0 +1,43 @@
...
+    String update(final Path path, final String newContent) {
+        try {
+            Files.write(path, newContent.getBytes());
+            return newContent;
+        } catch (final IOException ex) {
+            throw new UncheckedIOException(ex);
+        }
+    }
+
+    void delete (final Path path){
```
JonCook: Formatting is also off here for the delete method.

testing-modules/mocks/src/test/java/com/baeldung/jimfs/FileRepositoryUnitTest.java
```diff
@@ -0,0 +1,74 @@
...
+import java.nio.file.Path;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class FileRepositoryUnitTest implements FileTestProvider {
+
+    private final FileRepository fileRepository = new FileRepository();
+
+    private static Stream<Arguments> provideFileSystem() {
```
JonCook: Can we format this the same as FileManipulationUnitTest? Actually rather than duplicate this provider in the two tests, could we maybe move it into the interface FileTestProvider and then reference it with a fully qualified name in our tests. e.g) @MethodSource("com.baeldung.jimfs.FileTestProvider#provideFileSystem")

testing-modules/mocks/pom.xml
lukaszrys: I actually have been using this one https://github.com/eugenp/tutorials/tree/master/intelliJ -> is it different?

testing-modules/mocks/src/test/java/com/baeldung/jimfs/FilePathReaderTest.java
```diff
@@ -0,0 +1,48 @@
...
+import com.google.common.jimfs.Jimfs;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import java.nio.file.FileSystem;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
```
JonCook: Hey @lukaszrys - This class causes the build to break I think:
`
[INFO] PMD Failure: com/baeldung/jimfs/FilePathReaderTest.java:14 Rule:UnitTestMustFollowNamingConventionRule Priority:3 Unit test class names need to end in UnitTest, integration tests with IntegrationTest, etc.
`

lukaszrys: @JonCook thanks!

PR #7985, opened 2019-10-09
BAEL-3236 Determine the execution time of JUnit Tests

dev-chirag:


testing-modules/junit-5/src/test/java/com/baeldung/execution/time/RandomNumberUnitTest.java
```diff
@@ -0,0 +1,13 @@
+package com.baeldung.execution.time;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class RandomNumberUnitTest {
+
+    @Test
+    public void whenTwoRandomNumberThenAssertUniqueness() {
```
ashleyfrieze: As these are JUnit5 tests, by convention we remove the `public` keyword from both the class and the method.

There's no guarantee that two random numbers are different. This will result in a flickering test.

Given this is an article about timing tests, how about we, instead, put a `Thread.sleep` in the example. You CAN also guarantee uniqueness when generating two UUIDs.

testing-modules/junit-5/src/test/java/com/baeldung/execution/time/RandomNumberUnitTest.java
```diff
@@ -0,0 +1,13 @@
...
+
+        Assertions.assertNotEquals(Math.random(), Math.random());
```
ashleyfrieze: We usually would statically import `Assertions` rather than prefix the assertion method with it.

testing-modules/junit-5/src/test/java/com/baeldung/execution/time/SampleExecutionTimeUnitTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class SampleExecutionTimeUnitTest {
+
+    @Test
+    public void someUnitTest() {
+
+        assertTrue(this::doSomething);
```
ashleyfrieze: This seems needlessly complex. You're using a boolean supplier to `assertTrue`. Why not just call `doSomething()` directly?

testing-modules/junit-5/src/test/java/com/baeldung/execution/time/SampleExecutionTimeUnitTest.java
```diff
@@ -0,0 +1,46 @@
+package com.baeldung.execution.time;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class SampleExecutionTimeUnitTest {
```
ashleyfrieze: If this is a JUnit 5 test, then the class and the methods should not be marked `public`

testing-modules/junit-5/src/test/java/com/baeldung/execution/time/SampleExecutionTimeUnitTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Test
+    public void someEndToEndTest() {
+
+        try {
+            //simulate an operation that may take 10 seconds
+            Thread.sleep(10000);
+        } catch (InterruptedException e) {
+            fail();
+        }
```
ashleyfrieze: This is distracting - the try/catch, which are just caused by the use of `Thread.sleep` Just make your test methods `throws Exception`

PR #7991, opened 2019-10-10
Fix for "address" field and update to Wicket 8.x

bitstorm:


bitstorm:
@maibin 
Hi,
is there a reason why my PR has been closed without been merged?
Thanks.

PR #7993, opened 2019-10-11
BAEL-3336: Programmatic Transaction Management

alimate:
<del>I chose this module since it already configured the PostgreSQL test container. Also, I had to convert my code to JUnit 4. I guess we can migrate to JUnit 5 later.</del>

### DIFF
--- a/persistence-modules/spring-data-jpa-3/README.md
+++ b/persistence-modules/spring-data-jpa-3/README.md
@@ -1,0 +1,6 @@
+=========
+
+## Spring Data JPA Example Project
+
+### Relevant Articles: 
+- [Programmatic Transaction Management in Spring](http://www.baeldung.com/spring-programmatic-transaction-management)
--- a/persistence-modules/spring-data-jpa-3/README.md
+++ b/persistence-modules/spring-data-jpa-3/README.md
@@ -1,6 +1,0 @@
-=========
-
-## Spring Data JPA Example Project
-
-### Relevant Articles: 
-- [Programmatic Transaction Management in Spring](http://www.baeldung.com/spring-programmatic-transaction-management)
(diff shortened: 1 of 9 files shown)


persistence-modules/spring-data-jpa-3/README.md
```diff
@@ -0,0 +1,6 @@
+=========
```
KevinGilmore: You can omit this file for now. I will add it after the article goes live.

persistence-modules/spring-data-jpa-3/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+        <dependency>
+            <groupId>org.junit.platform</groupId>
+            <artifactId>junit-platform-launcher</artifactId>
+            <version>${junit-platform.version}</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <properties>
+        <testcontainers.version>1.10.6</testcontainers.version>
```
KevinGilmore: The latest is 1.12.2

persistence-modules/spring-data-jpa-3/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+        <postgresql.version>42.2.5</postgresql.version>
```
KevinGilmore: latest is 42.2.8

persistence-modules/spring-data-jpa-3/README.md
alimate: Resolved.

persistence-modules/spring-data-jpa-3/pom.xml
alimate: Resolved.

persistence-modules/spring-data-jpa-3/pom.xml
alimate: Resolved.

PR #7996, opened 2019-10-11
BAEL-2350 - DataJpaTest Configuration not found

martinvw:
Add a simple example to show the correct (package) structure for using @DataJpaTest

### DIFF
--- a/spring-boot-configuration/README.md
+++ b/spring-boot-configuration/README.md
@@ -1,0 +1,7 @@
+# Spring Boot Configuration
+
+This module contains articles about Spring Boot Configuration.
+
+### Relevant Articles: 
+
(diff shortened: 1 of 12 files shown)


spring-boot-configuration/README.md
```diff
@@ -0,0 +1,7 @@
+# Spring Boot Configuration
+
+This module contains articles about Spring Boot Configuration.
+
+### Relevant Articles: 
+
+- [DataJpaTest, Unable to Find Spring Boot Configuration](https://www.baeldung.com/data-jpa-test-unable-to-find/)
```
maibin: let's remove it - editors add that after the article is published, as we don't know the link yet.

martinvw: Okay, learned something new :-)

Updated the PR :-)

PR #8012, opened 2019-10-13
BAEL-3089 - created examples for @SecondaryTable article

fanatixan:


persistence-modules/java-jpa-2/src/main/resources/META-INF/persistence.xml
```diff
@@ -119,4 +119,34 @@
...
+            <property name="javax.persistence.jdbc.user" value="sa" />
+            <property name="javax.persistence.jdbc.password" value="" />
+            <property name="hibernate.dialect"
+                      value="org.hibernate.dialect.H2Dialect" />
+            <property name="hibernate.hbm2ddl.auto" value="create" />
+            <property name="hibernate.hbm2ddl.import_files" value="multipletables.sql" />
+            <property name="show_sql" value="true" />
+            <property name="hibernate.temp.use_jdbc_metadata_defaults"
+                      value="false" />
+<!--             <property name="javax.persistence.sql-load-script-source" -->
```
JonCook: Could we remove this unused comment. I didn't see the sql it refers to in the project?

persistence-modules/java-jpa-2/src/test/java/com/baeldung/jpa/multipletables/MultipleTablesIntegrationTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private static EntityManagerFactory emf;
+    private static EntityManager em;
+
+    @BeforeClass
+    public static void setup() {
+        emf = Persistence.createEntityManagerFactory("jpa-h2-multipltables");
+        em = emf.createEntityManager();
+    }
+
```
JonCook: Could we use test names as per the author documentation like given_xxx_. Thanks.

persistence-modules/java-jpa-2/src/main/resources/META-INF/persistence.xml
fanatixan: Sorry, I forgot to clean it up, removed it.

The sql is in src/test/resources/multipletables.sql

persistence-modules/java-jpa-2/src/test/java/com/baeldung/jpa/multipletables/MultipleTablesIntegrationTest.java
fanatixan: I am aware of the naming convention. However, we don't have anything given in my opinion, that's why I got rid of that part. Also, the documentation allows it: "given part is optional, but the other two are not"

Should I still rename them?

PR #8035, opened 2019-10-17
BAEL-3200 Error handling with Spring AMQP

CROSP:


CROSP:
Hi @JonCook thank you for the review.
Sorry for the code formatting, I didn't notice the incorrect formatting after reinstalling and resetting my IDE. As a result, the formatting and imports were messed up. Sorry about that.

`DLXCustomAmqpConfiguration` seems really to be referenced in the article, I created it for the default configuration, however then change the article. Thank you for pointing this out. 

As for `spring.main.allow-bean-definition-overriding=true` you are absolutely right, this will allow to have multiple beans of the same type. However, in that case, we need to manually set `Qualifier`,`Primary` and other priority annotations for the beans that are primary. But which beans should be primary? Beans for the latest example in the article? 
Also, we have multiple `RabbitListener`s that corresponds to different configurations. 
We may apply, for example, `ConditionalOnProperty`,`ConditionalOnBean`, or similar annotations to select a proper listener conditionally.  Furthermore, we can also use `ConditionalOnProperty` for the `Configuration` classes.
What do you think about this?

PR #8067, opened 2019-10-23
Feature/hexagonal architecture roger yates

rojyates:
Review: Temporary creation. To be deleted.


rojyates:
Removing as per instructions!

PR #8070, opened 2019-10-24
BAEL-3284: Guide to Flash Attributes in a Spring Web Application

tapankavasthi:
### References:
- [BAEL-3284](http://jira.baeldung.com/browse/BAEL-3284)

### DIFF
--- a/spring-mvc-simple-2/src/main/java/com/baeldung/flash_attributes/controllers/PoemSubmission.java
+++ b/spring-mvc-simple-2/src/main/java/com/baeldung/flash_attributes/controllers/PoemSubmission.java
@@ -1,0 +1,49 @@
+package com.baeldung.flash_attributes.controllers;
+
+import com.baeldung.flash_attributes.model.Poem;
+import org.springframework.stereotype.Controller;
+import org.springframework.ui.Model;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.ModelAttribute;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.servlet.mvc.support.RedirectAttributes;
+import org.springframework.web.servlet.support.RequestContextUtils;
+import org.springframework.web.servlet.view.RedirectView;
+import java.util.Map;
+
+import javax.servlet.http.HttpServletRequest;
+
+@Controller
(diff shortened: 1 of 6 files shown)


spring-mvc-simple-2/src/main/java/com/baeldung/flash_attributes/controllers/PoemSubmission.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.web.servlet.view.RedirectView;
+
+import javax.servlet.http.HttpServletRequest;
+
+@Controller
+public class PoemSubmission {
+
+    @GetMapping("/poem/success")
+    public String getSuccess(HttpServletRequest request) {
+        java.util.Map<String, ?> inputFlashMap = RequestContextUtils.getInputFlashMap(request);
```
KevinGilmore: Couldn't we just import java.util.Map so that we don't have to fully qualify it here?

spring-mvc-simple-2/src/main/java/com/baeldung/flash_attributes/model/Poem.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.logging.log4j.util.Strings;
+
+public class Poem {
+    private String title;
+    private String author;
+    private String body;
+
+    public static boolean isValidPoem(Poem poem) {
+        return poem != null && Strings.isNotBlank(poem.getAuthor()) && Strings.isNotBlank(poem.getBody())
+                && Strings.isNotBlank(poem.getTitle());
```
KevinGilmore: Only need to indent two spaces from where the statement begins.

spring-mvc-simple-2/src/main/java/com/baeldung/flash_attributes/controllers/PoemSubmission.java
tapankavasthi: Addressed

spring-mvc-simple-2/src/main/java/com/baeldung/flash_attributes/model/Poem.java
tapankavasthi: Addressed

PR #8094, opened 2019-10-28
Users list populated in index page when exists users

davidmoros:
The users list in index page i spopulated when exists users, before always did appears the message "No users yet!"


maibin:
@davidmoros could you please clarify what happened in here and why you raised this PR? Thanks!

PR #8139, opened 2019-11-04
BAEL-3418 Check If Two Strings Are Anagrams in Java

wugangca:


java-strings-ops-2/src/test/java/com/baeldung/string/AnagramUnitTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+    @Test
+    public void whenNotAnagram_thenReturnFalse() {
+       Anagram anagram = new Anagram();
+       String str1 = "abcaba";
+       String str2 = "cabbac";
+       assertFalse(anagram.isAnagramSort(str1, str2));
+       assertFalse(anagram.isAnagramCouting(str1, str2));
+       assertFalse(anagram.isAnagramMultiset(str1, str2));
+    }
```
ashleyfrieze: Let's break these up into 6 test cases. Let's also look at the naming. I don't think `thenReturnTrue` or `thenReturnFalse` is a helpful `then` in a test name. How about:

- givenAnagram_whenUsingSort
- givenAnagram_whenUsingCounting
- givenAnagram_whenUsingMultiset
- givenNonAnagram_whenUsingSort
... etc

java-strings-ops-2/src/main/java/com/baeldung/string/Anagram.java
```diff
@@ -0,0 +1,50 @@
...
+            return false;
+        }
+        char[] a1 = str1.toCharArray();
+        char[] a2 = str2.toCharArray();
+        Arrays.sort(a1);
+        Arrays.sort(a2);
+        return Arrays.equals(a1, a2);
+    }
+
+    public boolean isAnagramCouting(String str1, String str2) {
```
ashleyfrieze: This should be `counting` not `couting`

java-strings-ops-2/src/main/java/com/baeldung/string/Anagram.java
```diff
@@ -0,0 +1,50 @@
+package com.baeldung.string;
+
+import java.util.Arrays;
+import com.google.common.collect.HashMultiset;
+import com.google.common.collect.Multiset;
+
+public class Anagram {
+    public boolean isAnagramSort(String str1, String str2) {
```
ashleyfrieze: Rather than `str1` and `str2` I think we can afford `string1` and `string2` - it's less cryptic and more friendly. Let's do that throughout the file.

java-strings-ops-2/src/main/java/com/baeldung/string/Anagram.java
```diff
@@ -0,0 +1,50 @@
...
+        Arrays.sort(a1);
+        Arrays.sort(a2);
+        return Arrays.equals(a1, a2);
+    }
+
+    public boolean isAnagramCouting(String str1, String str2) {
+        if (str1.length() != str2.length()) {
+            return false;
+        }
+        final int MAX_CHARACTER = 256;
```
ashleyfrieze: Let's not declare the `final` here - this should be a `private static final` constant for the whole class.

It's also worth pointing out somewhere, that this algorithm will fail with `UTF-8` data. I'm reviewing the code before the article... but perhaps a comment in the code will also help an unwary reader.

java-strings-ops-2/src/main/java/com/baeldung/string/Anagram.java
```diff
@@ -0,0 +1,50 @@
...
+        int count[] = new int[MAX_CHARACTER];
+        for (int i = 0; i < str1.length(); i++) {
+            count[str1.charAt(i)]++;
+            count[str2.charAt(i)]--;
```
ashleyfrieze: This is a clever algorithm!

java-strings-ops-2/src/main/java/com/baeldung/string/Anagram.java
```diff
@@ -0,0 +1,54 @@
+package com.baeldung.string;
+
+import java.util.Arrays;
+import com.google.common.collect.HashMultiset;
+import com.google.common.collect.Multiset;
+
+public class Anagram {
+    //this definition only works for single byte encoding character set.
+    //For multibyte encoding, such as UTF-8, 16, 32 etc., 
+    //we need to increase this number so that it can contain all possible characters.
```
ashleyfrieze: `Minor` - it's conventional to put a space after `//` in comments

java-strings-ops-2/src/main/java/com/baeldung/string/Anagram.java
```diff
@@ -0,0 +1,54 @@
...
+    private static int MAX_CHARACTER = 256;
```
ashleyfrieze: `Minor` - this isn't the max character, it's the number of characters. MAX=255, MIN=0. `CHARACTER_RANGE` might be more correct. Don't change it if you disagree.

PR #8165, opened 2019-11-10
Follow up of [BAEL-2350], renamed module

martinvw:


martinvw:
Note to self, don't forget to rename in parent pom and childs.

PR #8177, opened 2019-11-12
BAEL-3414

rodrigolgraciano:
BAEL-3414 JDK vs Eclipse Collections Benchmark
Fix unit test on another article (HttpRequestUnitTest)

### DIFF
--- a/core-java-modules/core-java-11/pom.xml
+++ b/core-java-modules/core-java-11/pom.xml
@@ -100,46 +100,6 @@
                 </executions>
             </plugin>
         </plugins>
-        <pluginManagement>
-        <plugins>
-            <plugin>
-                <artifactId>maven-clean-plugin</artifactId>
-                <version>2.5</version>
-            </plugin>
-            <plugin>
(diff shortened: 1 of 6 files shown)


core-java-modules/core-java-11/pom.xml
```diff
@@ -42,14 +63,92 @@
...
+            </plugin>
+            <plugin>
+                <artifactId>maven-site-plugin</artifactId>
+                <version>3.3</version>
+            </plugin>
+            <plugin>
+                <artifactId>maven-source-plugin</artifactId>
+                <version>2.2.1</version>
+            </plugin>
+            <plugin>
```
maibin: Why we need all of those plugins?

core-java-modules/core-java-11/src/test/resources/sample.txt
```diff
@@ -0,0 +1 @@
+Sample file content
```
maibin: Remove it

PR #8187, opened 2019-11-13
 BAEL-3224 - Checking if Two Java Dates are On the Same Day

AlNiyas:
Review comments fixed. New branch as suggested.


JonCook:
retest

PR #8194, opened 2019-11-15
BAEL-3385: Hibernate @NotNull vs @Column(nullable = false)

kwoyke:


persistence-modules/spring-boot-persistence-h2/src/test/java/com/baeldung/ItemIntegrationTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Autowired
+    private ItemRepository itemRepository;
+
+    @Test
+    public void shouldNotAllowToPersistNullItemsPrice() {
+        Item saved = null;
+        try {
+            saved = itemRepository.save(new Item());
+        } catch (Exception e) {
```
pivovarit: This test will go green if no exception is thrown, check this method:
https://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/api/Assertions.html#assertThatThrownBy-org.assertj.core.api.ThrowableAssert.ThrowingCallable-

PR #8231, opened 2019-11-23
BAEL-3453 - Circular linked list Java implementation

radhe-sravan:


data-structures/src/main/java/com/baeldung/list/CircularLinkedList.java
```diff
@@ -0,0 +1,102 @@
+package com.baeldung.list;
+
+public class CircularLinkedList {
+
+    Node head = null;
```
JonCook: Could we make these members private? I don't think they need to be available outside the class right?

data-structures/src/main/java/com/baeldung/list/CircularLinkedList.java
```diff
@@ -0,0 +1,102 @@
...
+    Node tail = null;
+
+    public void addNode(int value) {
+
+        Node newNode = new Node(value);
+
+        // If no elements are present, make the newly addNodeed node as head
```
JonCook: A general comment, do we really need all these comments in the code? Most of them are pretty obvious.

data-structures/src/test/java/com/baeldung/list/CircularLinkedListUnitTest.java
```diff
@@ -0,0 +1,60 @@
+package com.baeldung.list;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
```
JonCook: The formatting is off in this class. Can we fix that, please.

data-structures/src/test/java/com/baeldung/list/CircularLinkedListUnitTest.java
```diff
@@ -0,0 +1,60 @@
...
+public class CircularLinkedListUnitTest {
+
+    @Test
+    public void givenACircularLinkedList_WhenAddingElements_ThenListContainsThoseElements() {
+
```
JonCook: Could we remove these empty lines from the start of the tests?

data-structures/src/main/java/com/baeldung/list/CircularLinkedList.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public void traverseList() {
+
+        // Start traversing from the head
+        Node currentNode = head;
+
+        if (head != null) {
+            do {
+                System.out.print(currentNode.value + " ");
```
JonCook: Avoid calls to System.out... and use a proper logger.

PR #8237, opened 2019-11-23
Changes for "Guide to finally in Java" (BAEL-3526)

rahulgul8:


guest/core-java/src/main/java/com/baeldung/finallykeyword/FinallyExecutedCases.java
```diff
@@ -0,0 +1,44 @@
...
+        try {
+            System.out.println("Inside try");
+            return "from try";
+        } catch (Exception e) {
+            System.out.println("Inside catch");
+            return "from catch";
+        } finally {
+            System.out.println("Inside finally");
+        }
+    }
```
ashleyfrieze: Are we missing the case where the try block throws, so we need to return from the exception handler. It's the same as `returnFromTryOrCatch` but with a throw in the `try`

guest/core-java/src/main/java/com/baeldung/finallykeyword/FinallyNotExecutedCases.java
```diff
@@ -0,0 +1,54 @@
...
+        } finally {
+            System.out.println("Inside finally");
+        }
+    }
+
+    public void infiniteLoop() {
+        try {
+            System.out.println("Inside try");
+            while (true)
+                ;
```
ashleyfrieze: Let's just use empty `{}` block, rather than the harder to read dangling `;`

PR #8263, opened 2019-11-28
[BAEL-17468] spring-data-mongodb | fix Live Test

rozagerardo:
fixed test, the 'error' was caused is caused by an improvement in theSpring Data Mongodb library, it now handles the count() operation gracefully by using the mongo-java-driver countDocument() operation (which in turn relies in the count aggregation stage), as suggested by the mongodb docs'


jzheaux:
Closing in favor of https://github.com/eugenp/tutorials/pull/9169

PR #8283, opened 2019-12-01
Double to Integer Casting (BAEL-3517)

sreekanthsnair:


java-numbers-3/README.md
```diff
@@ -0,0 +1,7 @@
+## Java Number Cookbooks and Examples
```
eric-martin: We don't need this file. It will be added by another process after publishing.

java-numbers-3/src/main/java/com/baeldung/doubletolong/DoubleToLongUnitTest.java
```diff
@@ -0,0 +1,36 @@
...
+    public void using_Math_Floor() {
+        Assert.assertEquals(9999L, Math.floor(VALUE), 0);
+    }
+
+    @Test
+    public void using_Type_Cast() {
+        Assert.assertEquals(9999L, (long) VALUE);
+    }
+
+}
```
eric-martin: We don't need the gitkeep file

java-numbers-3/src/main/java/com/baeldung/doubletolong/DoubleToLongUnitTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class DoubleToLongUnitTest {
+
+    final static double VALUE = 9999.999;
+
+    @Test
+    public void using_longValue() {
```
eric-martin: Follow our documented unit test method naming conventions

java-numbers-3/pom.xml
```diff
@@ -0,0 +1,28 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>java-numbers-3</artifactId>
```
KevinGilmore: Need to reference this module in the root pom.xml so that it gets built automatically. Look for occurrences of the other java-numbers modules and add this one in the same places.

PR #8292, opened 2019-12-02
Update ThreadLocalRandom benchmark to show issues with contention

karl82:
Previous benchmark was measuring initialization of thread pools and
their shutdown, not really measuring Random.nextInt() with shared Random
instance and ThreadLocalRandom.current()

Example when run with 8 threads

```
Benchmark                                                         Mode  Cnt     Score     Error   Units
ThreadLocalRandomBenchMarker.randomValuesUsingRandom             thrpt   20     9.597 ±   0.475  ops/us
ThreadLocalRandomBenchMarker.randomValuesUsingThreadLocalRandom  thrpt   20  2195.178 ± 109.579  ops/us
```

This will require to change https://www.baeldung.com/java-thread-local-random#comparing-threadlocalrandom-and-random-using-jmh


maibin:
@fanatixan could you look into this one?

fanatixan:
@maibin, looks good

PR #8334, opened 2019-12-08
BAEL-3491: adding example classes and unit tests

wapgui:
Complete code, link to the article will be added if approved.


JonCook:
A new PR will be created.

PR #8340, opened 2019-12-09
Fix test code for success

jneuhaus:
Two files needed updates so that tests
would complete successfully.


eugenp:
Hey @jneuhaus - what JIRA is this part of?

jneuhaus:
@eugenp I did not submit a bug in Jira for this. I made two modifications to make the tests pass, and submitted those changes as a pull request. Did I need to submit a bug report first?

lor6:
Thanks @jneuhaus  We'll check the fix and merge it if everything looks ok.

amit2103:
Hi @jneuhaus ,
I run UserRepositoryIntegrationTest test case with the existing code and its running successfully in my local. Can you give some description about the issue ?

jneuhaus:
If you view diffs you should see a bad reference; however, it was minor and
possibly already fixed?

lor6:
Hi @jneuhaus  We checked the tests and the syntax of the first query and everything is working ok. Since the first query is a native query, you may run into issues depending on the database you're using. Our example uses H2 database and the syntax is valid for it. 

Thanks.

PR #8343, opened 2019-12-09
BAEL-3370 Guide to read command

srzamfir:


JonCook:
retest

linux-bash/read/src/main/bash/read_inputs.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
```
JonCook: Could we maybe add a comment at the top here with some example usage? I tried for example but didn't see any output:
- `./read_inputs.sh default_read`

linux-bash/read/src/main/bash/read_inputs.sh
```diff
@@ -0,0 +1,71 @@
...
+# section 2.1
+default_read() {
+    read input1 input2 input3
+    echo "[$input1] [$input2] [$input3]"
+}
+
+# section 2.2
+custom_ifs_no_array(){
```
JonCook: A minor point, let's be consistent with the spaces and parenthesis like you have been with some of the other functions. ;-)

linux-bash/read/src/main/bash/read_inputs.sh
srzamfir: i did a simple menu to clarify things out ... just run ./read_inputs.sh

linux-bash/read/src/main/bash/read_inputs.sh
srzamfir: done

PR #8348, opened 2019-12-10
Bael 3510-method multiple return values

M-Abdelbaset:


core-java-modules/core-java-lang-2/src/main/java/com/baeldung/methodmultiplereturnvalues/Coordinates.java
```diff
@@ -0,0 +1,62 @@
+package com.baeldung.methodmultiplereturnvalues;
+
+class Coordinates {
+
+    private final double longitude;
+    private final double latitude;
+    private final String directionsNote;
```
ashleyfrieze: Let's not make these `final`. It's unnecessary code for this example.

core-java-modules/core-java-lang-2/src/main/java/com/baeldung/methodmultiplereturnvalues/Coordinates.java
```diff
@@ -0,0 +1,62 @@
...
+            if (other.directionsNote != null)
+                return false;
+        } else if (!directionsNote.equals(other.directionsNote))
+            return false;
+        if (Double.doubleToLongBits(latitude) != Double.doubleToLongBits(other.latitude))
+            return false;
+        if (Double.doubleToLongBits(longitude) != Double.doubleToLongBits(other.longitude))
+            return false;
+        return true;
+    }
```
ashleyfrieze: Let's remove the hashcode/equals as we're not using them and they're just boilerplate.

core-java-modules/core-java-lang-2/src/main/java/com/baeldung/methodmultiplereturnvalues/Coordinates.java
```diff
@@ -0,0 +1,62 @@
+package com.baeldung.methodmultiplereturnvalues;
+
+class Coordinates {
```
ashleyfrieze: This class should surely be public, rather than package private?

core-java-modules/core-java-lang-2/src/main/java/com/baeldung/methodmultiplereturnvalues/Tuple3.java
```diff
@@ -0,0 +1,64 @@
...
+                return false;
+        } else if (!third.equals(other.third))
+            return false;
+        if (second == null) {
+            if (other.second != null)
+                return false;
+        } else if (!second.equals(other.second))
+            return false;
+        return true;
+    }
```
ashleyfrieze: Again, let's remove `equals` and `hashcode` methods

core-java-modules/core-java-lang-2/src/test/java/com/baeldung/methodmultiplereturnvalues/MultipleReturnValuesUsingArraysUnitTest.java
```diff
@@ -0,0 +1,47 @@
+package com.baeldung.methodmultiplereturnvalues;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
ashleyfrieze: You're using JUnit5 assertions with JUnit4 tests - this is confusing. Please choose which you're using and use consistently.

core-java-modules/core-java-lang-2/src/test/java/com/baeldung/methodmultiplereturnvalues/MultipleReturnValuesUsingTuplesUnitTest.java
```diff
@@ -0,0 +1,28 @@
...
+import com.baeldung.methodmultiplereturnvalues.Tuple3;
+
+public class MultipleReturnValuesUsingTuplesUnitTest {
+
+    private MultipleReturnValuesUsingTuples multipleReturnValuesUsingTuples;
+    
+    @Before
+    public void setUp() {
+        this.multipleReturnValuesUsingTuples = new MultipleReturnValuesUsingTuples();
+    }
```
ashleyfrieze: I think the functions we're testing can be made `static`, so we can avoid these clunky test setups.

core-java-modules/core-java-lang-2/src/main/java/com/baeldung/methodmultiplereturnvalues/Coordinates.java
M-Abdelbaset: OK

core-java-modules/core-java-lang-2/src/test/java/com/baeldung/methodmultiplereturnvalues/MultipleReturnValuesUsingArraysUnitTest.java
M-Abdelbaset: done

core-java-modules/core-java-lang-2/src/test/java/com/baeldung/methodmultiplereturnvalues/MultipleReturnValuesUsingTuplesUnitTest.java
M-Abdelbaset: ok

core-java-modules/core-java-lang-2/src/main/java/com/baeldung/methodmultiplereturnvalues/Coordinates.java
M-Abdelbaset: I just wanted to make the class immutable. Anyway, I've removed the final modifier from all classes.

core-java-modules/core-java-lang-2/src/test/java/com/baeldung/methodmultiplereturnvalues/MultipleReturnValuesUsingCollectionsUnitTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+public class MultipleReturnValuesUsingCollectionsUnitTest {
+
+    @Test
+    public void whenReturningList_thenMultipleFieldsAreReturned() {
```
ashleyfrieze: This is now a `JUnit5` test, which means its class and methods do not need to be `public`.

core-java-modules/core-java-lang-2/src/main/java/com/baeldung/methodmultiplereturnvalues/MultipleReturnValuesUsingTuples.java
```diff
@@ -0,0 +1,18 @@
+package com.baeldung.methodmultiplereturnvalues;
+
+import java.util.List;
+
+class MultipleReturnValuesUsingTuples {
+
+    static Tuple2<Integer, Coordinates> getFirstNullNearbyLocation(List<Coordinates> coordinatesList) {
+
+        int id = 55;
```
ashleyfrieze: This example is a little harder to follow as the `Coordinates` class has a `NearbyLocation` field that doesn't serve much of a purpose, and the `id` here is a meaningless value.

What if you made this example something like `getMostDistantPoint` which had to return a distance and the coordinate from the list which was furthest from some coordinate also supplied. You could implement a distance function in `Coordinates` which uses the pythagoras distance between two coordinates... It might be a good example...?

PR #8351, opened 2019-12-11
BAEL-3253: Use a random port number for Tomcat

kwoyke:


libraries-server/src/main/java/com/baeldung/tomcat/ProgrammaticTomcat.java
```diff
@@ -15,14 +16,29 @@
 
     private Tomcat tomcat = null;
 
+    private int randomPort;
+
+    public ProgrammaticTomcat() {
+        // Get a random port number in range 6000 (inclusive) - 9000 (exclusive)
+        this.randomPort = new Random()
```
pivovarit: We have no guarantees here that the chosen port will be actually free. It'd make sense to keep to validate that. You can use the trick with a `ServerSocket` for that:

```
public static boolean isFree(int port) {
        try {
            new ServerSocket(port).close();
            return true;
        } catch (IOException e) {
            return false;
        } 
    }
```

and then:
```
...
.ints(6000, 9000)
.filter(port -> isFree(port))
.findFirst()
.orElse(8080);
```

PR #8354, opened 2019-12-11
[BAEL-3495] Added code for new article 

martinvw:
Moved class so that the scans do not collide
Make sure the LiveTest actually stops when done
Also removed some useless .gitignore / README parts.

And add actual code for new article


spring-security-core/.gitignore
```diff
@@ -1,13 +1,2 @@
-*.class
-
 #folders#
-/target
-/neoDb*
-/data
-/src/main/webapp/WEB-INF/classes
-*/META-INF/*
-
```
rpvilao: intended? just making sure.

martinvw: Yes, a lot of things that were ignored could never occur.

martinvw:
I had to redo my changes because the directory was moved around

PR #8358, opened 2019-12-12
BAEL-3589 : Improvement: Java Generate Random String

sampada07:
@glmartin : For your review.

### DIFF
--- a/core-java-modules/core-java-strings/src/test/java/com/baeldung/randomstrings/RandomStringsUnitTest.java
+++ b/core-java-modules/core-java-strings/src/test/java/com/baeldung/randomstrings/RandomStringsUnitTest.java
@@ -39,6 +39,35 @@ public class RandomStringsUnitTest {
     }
 
     @Test
+    public void givenUsingJava8_whenGeneratingRandomAlphabeticString_thenCorrect() {
+        final int leftLimit = 97; // letter 'a'
+        final int rightLimit = 122; // letter 'z'
+        final int targetStringLength = 10;
+        final Random random = new Random();
+
+        final String generatedString = random.ints(targetStringLength, leftLimit, rightLimit + 1)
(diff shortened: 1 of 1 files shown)


core-java-modules/core-java-strings/src/test/java/com/baeldung/randomstrings/RandomStringsUnitTest.java
```diff
@@ -1,72 +1,101 @@
...
+
+    @Test
+    public void givenUsingJava8_whenGeneratingRandomAlphanumericString_thenCorrect() {
+        final int leftLimit = 48; // numeral '0'
+        final int rightLimit = 122; // letter 'z'
+        final int targetStringLength = 10;
+        final Random random = new Random();
+
+        final String generatedString = random.ints(targetStringLength, leftLimit, rightLimit + 1)
+            .filter(i -> (i < 57 || i > 65) && (i < 90 || i > 97))
```
martinvw: Sorry to bother you, but the length of the result string varies because of this filter, I doubt that that is desirable.

Maybe add assertions (about the length) to both tests.

And add a limit after the filter instead of before.

PS: I'm also surprised to see `Random `being used here instead of `SecureRandom`

martinvw: And one more it seems your filter also excludes some wrong values, maybe you better extract it as a separate method and do some boundary tests on it.

The following values are filtered out: 9:;<=>?@AZ[\]^_`a

Of which 9, A, Z and a don't look like values you would like to filter out.

sampada07: Thanks @martinvw for the suggested improvements, i have incorporated them. For the Random vs SecureRandom thing, please refer to the JIRA ticket's description! :)

PR #8370, opened 2019-12-14
BAEL-3500 Emails with Logback

shubhi22:


logging-modules/logback/pom.xml
```diff
@@ -50,6 +50,17 @@
...
         </dependency>
+        <dependency>
+            <groupId>javax.mail</groupId>
+            <artifactId>mail</artifactId>
+            <version>1.4.7</version>
+        </dependency>
+        <dependency>
+            <groupId>javax.activation</groupId>
+            <artifactId>activation</artifactId>
+            <version>1.1.1</version>
```
KevinGilmore: Let's define the version numbers in properties rather than inline.

shubhi22: I had done that initially but later removed those as part of copying dependencies to the post (forgot to redo them).
Done now, please check,

PR #8376, opened 2019-12-14
BAEL-3158

YassinHajaj:
Introduction to Convolutional Neural Networks (CNN)


machine-learning/src/main/java/com/baeldung/cnn/MNISTDataSet.java
```diff
@@ -0,0 +1,63 @@
...
+    public static final String TEST_IMAGES = "src/main/resources/mnist/t10k-images.idx3-ubyte";
+    private static int OFFSET_SIZE = 4;
+    private static int NUM_ITEMS_OFFSET = 4;
+    private static int ITEMS_SIZE = 4;
+    private static int ROWS = 28;
+    private static int COLUMNS = 28;
+    private static int IMAGE_OFFSET = 16;
+    private static int IMAGE_SIZE = ROWS * COLUMNS;
+
+    public static RecordReaderDataSetIterator load(boolean test) throws IOException, InterruptedException {
```
ashleyfrieze: In general, it can be cleaner to provide three functions:

- `loadTest()`
- `loadTraining()`
- `load(File, File) // called by the other two`

This avoids the function that switches its meaning by an incoming boolean...

YassinHajaj: Indeed, this is a very good catch. Solved it and pushed.

ashleyfrieze:
This article was closed.

PR #8390, opened 2019-12-16
Make sure that native sqlite4 dependencies are copied to expected path

bhattacharyyasom:
We need to add this plugin to get native dependencies to specific folder.
In this example we set the `sqlite4java.library.path` to show native libraries.
So we make sure that the native dependencies are copied over to the `native-libs` folder.
Without this added I am seeing unreliable behavior on IDEs.

Reference : https://stackoverflow.com/questions/26901613/easier-dynamodb-local-testing


lor6:
The plugin was added to the project in the meantime, so I'll close this PR.

PR #8392, opened 2019-12-17
[BAEL-3485] - Java Range lookup problem

macroscopic64:


algorithms-searching/src/test/java/com/baeldung/algorithms/quadtree/QuadTreeSearchTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.Assert;
+
+import java.util.List;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class QuadTreeSearchTest {
```
KevinGilmore: Unit test class names should end in "UnitTest", otherwise the build will fail

macroscopic64: Yes, just realized it now after going through the Jenkins log. Thanks for confirming. Have fixed it now.

PR #8395, opened 2019-12-17
BAEL-3242: Use random port for test execution in spring-vertx module

kwoyke:


spring-vertx/src/main/java/com/baeldung/vertxspring/config/PortConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.context.annotation.Profile;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+
+@Configuration
+public class PortConfiguration {
+
+    private static final int DEFAULT_PORT = 8069;
+
```
JonCook: Hey @kwoyke - One very minor thing, using the eclipse formatter, we lose one of the empty lines. Sorry but let's fix that.

PR #8410, opened 2019-12-20
Whats New in Gradle 6.0

gilday:
Introduces a multi-project Gradle 6.0 build. The build uses Gradle wrapper to ensure that Gradle 6.0 is used.

While the What's New in Gradle 6.0 article includes a section on JDK 13 support, I did not include this code, because to my knowledge, the Baeldung build server uses JDK 1.8.

Consistent with the `gradle-5` project, I did not include this project in top level Maven build.

Descriptions of the projects in this multi-project build are as follows

* **configuration-avoidance** - demonstrates building custom tasks using Gradle Configuration Avoidance APIs
* **dependency-constraints** - demonstrates adding rich dependency information to a Gradle build
* **fibonacci-spi** and **fibonacci-recursive** - demonstrates sharing test fixtures between projects
* **httpclient-platform**, **person-rest-client**, and **widget-rest-client** - demonstrates using a `java-platform` to control dependency versions across projects
* **module-metadata-publishing** - demonstrates configuring a Gradle build to publish Gradle Module Metadata JSON to a Maven repository

Article preview http://inprogress.baeldung.com/?p=170448&preview=true


gradle-6/dependency-constraints/src/main/java/com/baeldung/gradle/RxHelloWorld.java
```diff
@@ -0,0 +1,20 @@
+package com.baeldung.gradle;
+
+import com.google.common.collect.ImmutableList;
+import io.reactivex.Observable;
+
+import java.util.List;
+
+/** Demonstrates a library type that returns an RxJava type. */
```
glmartin: Let's use traditional javadoc here. Here is some information: https://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#format

gilday: I think I've addressed this feedback, but I struggled to find a way to automatically check this. The Oracle documentation you've linked references a DocCheck doclet for verifying formatting, but I cannot find this jar.

The Baeldung style guide does not specify that JavaDoc should be formatted.

```
<setting id="org.eclipse.jdt.core.formatter.comment.format_javadoc_comments" value="false"/>
```

 I changed this to `true` and formatted the project. Do the changes address your concern?

PR #8418, opened 2019-12-22
[BAEL-3390] - Overview of generating barcodes and QR codes in Java

catalin-burcea:


libraries-3/pom.xml
```diff
@@ -8,8 +8,9 @@
 
     <parent>
         <groupId>com.baeldung</groupId>
-        <artifactId>parent-modules</artifactId>
-        <version>1.0.0-SNAPSHOT</version>
+        <artifactId>parent-boot-2</artifactId>
+        <version>0.0.1-SNAPSHOT</version>
+        <relativePath>../parent-boot-2</relativePath>
```
ashleyfrieze: Why have you changed the parent?

libraries-3/src/main/java/com/baeldung/barcodes/BarcodesController.java
```diff
@@ -0,0 +1,116 @@
...
+import java.io.IOException;
+
+@RestController
+@RequestMapping("/barcodes")
+public class BarcodesController {
+
+    //Barbecue library
+
+    @GetMapping(value = "/barbecue/upca", produces = MediaType.IMAGE_PNG_VALUE)
+    public ResponseEntity<BufferedImage> barbecueUPCABarcode(@RequestParam("barcodeText") String barcodeText) throws BarcodeException, OutputException {
```
ashleyfrieze: Multiple `throws` here. Spring doesn't care what you throw, so maybe just `throws Exception`?

ashleyfrieze: Can we not use `pathparam` rather than request param. This would be a cleaner REST interface, then.

libraries-3/src/main/java/com/baeldung/barcodes/BarcodesController.java
```diff
@@ -0,0 +1,116 @@
...
+    @GetMapping(value = "/barcode4j/code128", produces = MediaType.IMAGE_PNG_VALUE)
+    public ResponseEntity<BufferedImage> barcode4jCode128Barcode(@RequestParam("barcodeText") String barcodeText) {
+        BufferedImage image = Barcode4jBarcodeGenerator.generateCode128BarcodeImage(barcodeText);
+        return new ResponseEntity<>(image, HttpStatus.OK);
+    }
+
+    @GetMapping(value = "/barcode4j/pdf417", produces = MediaType.IMAGE_PNG_VALUE)
+    public ResponseEntity<BufferedImage> barcode4jPDF417Barcode(@RequestParam("barcodeText") String barcodeText) {
+        BufferedImage image = Barcode4jBarcodeGenerator.generatePDF417BarcodeImage(barcodeText);
+        return new ResponseEntity<>(image, HttpStatus.OK);
```
ashleyfrieze: It seems like you might be able to create a function to build the response entity

```java
private static ResponseEntity<BufferedImage> okResponse(BufferedImage image) {
    return new ResponseEntity<>(image, HttpStatus.OK);
}
```

This would allow callsites to be a little cleaner:

```java
   return okResponse(generatePDF417BarcodeImage(barcodeText));
```

libraries-3/pom.xml
catalin-burcea: Since we have a web-server, I decided to add the Spring Boot parent which also has the previous parent as its parent.

libraries-3/src/main/java/com/baeldung/barcodes/BarcodesController.java
catalin-burcea: done

catalin-burcea: done

libraries-3/src/main/java/com/baeldung/barcodes/BarcodesController.java
catalin-burcea: done

PR #8429, opened 2019-12-24
BAEL-3394 : Using headless mode in Java

sampada07:


core-java-modules/core-java-lang-2/src/test/java/com/baeldung/headlessmode/HeadlessModeUnitTest.java
```diff
@@ -0,0 +1,83 @@
...
+import javax.imageio.ImageIO;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class HeadlessModeUnitTest {
+
+    private static final String IN_FILE = "src/test/resources/product.png";
```
rpvilao: it sounds to me this is a good candidate to get the stream as a resource rather than via filesystem, what do you think?

core-java-modules/core-java-lang-2/src/test/java/com/baeldung/headlessmode/HeadlessModeUnitTest.java
```diff
@@ -0,0 +1,83 @@
...
+    public void whenHeadlessMode_thenImagesWork() {
+        boolean result = false;
+        try (FileInputStream inStream = new FileInputStream(IN_FILE); FileOutputStream outStream = new FileOutputStream(OUT_FILE)) {
+
+            BufferedImage inputImage = ImageIO.read(inStream);
+
+            result = ImageIO.write(inputImage, FORMAT, outStream);
+
+        } catch (IOException e) {
+            LOGGER.error(e.getMessage());
```
rpvilao: if that fails this should fail completely right? In this case better to declare that the method can throw exception, if we cannot handle the exception we should always throw it to the stack

core-java-modules/core-java-lang-2/src/test/java/com/baeldung/headlessmode/HeadlessModeUnitTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+        assertThat(fm.getHeight()).isGreaterThan(0);
+        assertThat(fm.getAscent()).isGreaterThan(0);
+        assertThat(fm.getDescent()).isGreaterThan(0);
+    }
+
+    @Test
+    public void whenHeadlessMode_thenImagesWork() {
+        boolean result = false;
+        try (FileInputStream inStream = new FileInputStream(IN_FILE); FileOutputStream outStream = new FileOutputStream(OUT_FILE)) {
```
rpvilao: get as resource is a good candidate (as discussed above)

core-java-modules/core-java-lang-2/src/test/java/com/baeldung/headlessmode/HeadlessModeUnitTest.java
sampada07: I can get the stream as resource, but for the write operation i still need the whole path. So to keep in and out file code consistent, i used this way. If you still insist, i can change the in-file code though :)

core-java-modules/core-java-lang-2/src/test/java/com/baeldung/headlessmode/HeadlessModeUnitTest.java
sampada07: Sounds good, i will make this change. Let me know what you think on the previous comment for file i/o and i'll push all changes at once.

core-java-modules/core-java-lang-2/src/test/java/com/baeldung/headlessmode/HeadlessModeUnitTest.java
rpvilao: I see! the best option is to read as resource and write to a temp folder, in this way we are not dependent from the way the source code is structured nor the building system we are using.

core-java-modules/core-java-lang-2/src/test/java/com/baeldung/headlessmode/HeadlessModeUnitTest.java
rpvilao: cool :)

core-java-modules/core-java-lang-2/src/test/java/com/baeldung/headlessmode/HeadlessModeUnitTest.java
sampada07: Ok, that makes sense. I've updated the code.

sampada07:
Closing this one, will check in from another branch in another PR :)

PR #8441, opened 2019-12-27
[BAEL-2948] Using cookies with Selenium WebDriver

chris9408:
[BAEL-2948] Using cookies with Selenium WebDriver
Article: http://inprogress.baeldung.com/?p=173129&preview=true


testing-modules/selenium-junit-testng/src/test/java/com/baeldung/selenium/junit/SeleniumCookiesJUnitLiveTest.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Test
+    public void whenOverridingCookie_thenItIsUpdated() {
+        driver.navigate().to(navUrl);
+        Cookie lpCookie = driver.manage().getCookieNamed("lp_120073");
+        driver.manage().deleteCookie(lpCookie);
+
+        lpCookie = new Cookie("lp_120073", "foo");
+        driver.manage().addCookie(lpCookie);
```
KevinGilmore: Let's use a new variable here when re-adding the cookie.

testing-modules/selenium-junit-testng/src/test/java/com/baeldung/selenium/junit/SeleniumCookiesJUnitLiveTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+        assertThat(lpCookie.getValue(), equalTo("foo"));
```
KevinGilmore: Should retrieve the newly added cookie from the driver, then assert on the value of the retrieved cookie. This line only asserts on the value that you set a couple of lines earlier.

PR #8469, opened 2020-01-03
[BAEL-3339] - Creates @CurrentSecurityContext example

jesperancinha:


spring-boot-security/src/test/java/com/baeldung/springbootsecurity/oauth2server/CustomConfigAuthorizationServerIntegrationTest.java
```diff
@@ -1,33 +1,68 @@
...
 
 @RunWith(SpringRunner.class)
 @SpringBootTest(webEnvironment = RANDOM_PORT, classes = SpringBootAuthorizationServerApplication.class)
 @ActiveProfiles("authz")
 public class CustomConfigAuthorizationServerIntegrationTest extends OAuth2IntegrationTestSupport {
 
+    URL base;
+
+    @LocalServerPort
+    int port;
```
ashleyfrieze: These fields would normally be `private` in a unit test.

spring-boot-security/src/test/java/com/baeldung/springbootsecurity/oauth2server/CustomConfigAuthorizationServerIntegrationTest.java
```diff
@@ -1,33 +1,68 @@
...
+
+    @Before
+    public void setUp() throws MalformedURLException {
```
ashleyfrieze: Generally, it's a good idea to have `throws Exception` rather than a more specific one in the signature of test methods.

spring-boot-security/src/test/java/com/baeldung/springbootsecurity/oauth2server/CustomConfigAuthorizationServerIntegrationTest.java
```diff
@@ -1,33 +1,68 @@
...
         ClientCredentialsResourceDetails resourceDetails = getClientCredentialsResourceDetails("baeldung", singletonList("read"));
         OAuth2RestTemplate restTemplate = getOAuth2RestTemplate(resourceDetails);
 
         OAuth2AccessToken accessToken = restTemplate.getAccessToken();
 
-        assertNotNull(accessToken);
+        final String authentication  = restTemplate.execute(base.toString() + "/authentication", HttpMethod.GET, clientHttpRequest -> {
+        }, clientHttpResponse -> IOUtils.toString(clientHttpResponse.getBody(), Charset.defaultCharset()));
 
+        final String principal  = restTemplate.execute(base.toString() + "/principal", HttpMethod.GET, clientHttpRequest -> {
```
ashleyfrieze: Let's omit the use of `final` here. Let's also break out these additions as separate unit tests to illustrate them individually.

spring-boot-security/src/test/java/com/baeldung/springbootsecurity/oauth2server/CustomConfigAuthorizationServerIntegrationTest.java
```diff
@@ -1,33 +1,68 @@
...
+        }, clientHttpResponse -> IOUtils.toString(clientHttpResponse.getBody(), Charset.defaultCharset()));
+
+        assertNotNull(accessToken);
+        assertEquals(authentication, "baeldung");
+        assertEquals(principal, "baeldung");
```
ashleyfrieze: The assertions are in the wrong order. It's `assertEquals(expected, actual)`

spring-boot-security/src/main/java/com/baeldung/springbootsecurity/oauth2server/config/AuthorizationServerConfig.java
```diff
@@ -25,20 +25,20 @@ public void configure(AuthorizationServerEndpointsConfigurer endpoints) throws E
...
+                .secret(passwordEncoder().encode("baeldung"))
+                .authorizedGrantTypes("client_credentials", "password", "authorization_code")
+                .scopes("openid", "read")
+                .autoApprove(true)
+                .and()
+                .withClient("baeldung-admin")
+                .secret(passwordEncoder().encode("baeldung"))
+                .authorizedGrantTypes("authorization_code", "client_credentials", "refresh_token")
+                .scopes("read", "write")
+                .autoApprove(true);
```
ashleyfrieze: This change should probably be reverted.

spring-boot-security/src/test/java/com/baeldung/springbootsecurity/oauth2server/OAuth2IntegrationTestSupport.java
```diff
@@ -31,4 +39,18 @@ protected OAuth2RestTemplate getOAuth2RestTemplate(final ClientCredentialsResour
...
     }
+
+    protected String executeGetRequest(OAuth2RestTemplate restTemplate, String path) {
+        return restTemplate.execute(base.toString() + path, GET, getRequestCallback(), getStringResponseExtractor());
+    }
+
+    protected RequestCallback getRequestCallback() {
+        return clientHttpRequest -> {
+        };
+    }
```
ashleyfrieze: This is a do nothing callback... I would suggest that this way of expressing it is a bit harder to parse. If you did something like this:

```java
private static final RequestCallback DO_NOTHING_CALLBACK = request -> {};
```

And then used it at the call site as `DO_NOTHING_CALLBACK` - I think it might be more self-documenting.

spring-boot-security/src/test/java/com/baeldung/springbootsecurity/oauth2server/OAuth2IntegrationTestSupport.java
```diff
@@ -31,4 +39,18 @@ protected OAuth2RestTemplate getOAuth2RestTemplate(final ClientCredentialsResour
...
+
+    protected ResponseExtractor<String> getStringResponseExtractor() {
+        return clientHttpResponse -> IOUtils.toString(clientHttpResponse.getBody(), Charset.defaultCharset());
+    }
```
ashleyfrieze: Similarly, you don't need to make a new lambda every time for this. You just want something like:

```java
protected static final ResponseExtractor<String> EXTRACT_BODY_AS_STRING =  clientHttpResponse -> IOUtils.toString(clientHttpResponse.getBody(), Charset.defaultCharset());
```

PR #8499, opened 2020-01-09
vijay-sassy: Creating Date in java 8

vijay-sassy:
Added Create Date class & Unit Tests for all the cases


core-java-modules/core-java-datetime-java8/src/test/java/com/baeldung/datebasics/CreateDateUnitTest.java
```diff
@@ -0,0 +1,56 @@
...
+    
+    @Test
+    public void whenUsingClock_thenLocalDate() {
+        assertEquals("2020-01-08", date.getTodaysDateFromClock());
+    }
+    
+    @Test
+    public void givenValues_whenUsingZone_thenLocalDate() {
+        assertEquals("2020-01-08", date.getTodaysDateFromZone("Asia/Kolkata"));
+    }
```
KevinGilmore: Note that these tests will fail on any other date.

PR #8530, opened 2020-01-14
DRY the functional web handler

briantopping:
This separation of handlers is a great idea for testing, but unfortunately the results of those tests depend on the functionality being duplicated to the functions that are actually called. This patch makes the code more DRY.


lor6:
Thanks @briantopping  We'll check the fix and merge it if everything look ok

sampada07:
The code that this PR aims to remove is actually used in the article to present an alternate and concise way to define routes. In that sense, even if this piece of code is unused, it should be kept as is. If readers want to go for composed routes, they can refer to this fully-functional code.

dkapil:
Closing this as per above comment from @sampada07

PR #8536, opened 2020-01-15
BAEL-20602: Use regular expression to match strings with new lines

kwoyke:


kwoyke:
We'll keep it simple and use assertThat.contains().

PR #8537, opened 2020-01-15
Redundant Variable 'config'

the-magic-max:
Instead of initializing a local variable first and then return it, you could directly return a new Builder instance.

-> `return new Builder().uri(URL).build();`


lor6:
Thanks @the-magic-max  Merged.

PR #8538, opened 2020-01-15
Bael 20764 1

amit2103:


amit2103:
move version in pom properties

java-numbers-2/pom.xml
```diff
@@ -21,6 +21,11 @@
             <artifactId>jmh-generator-annprocess</artifactId>
             <version>${jmh-generator.version}</version>
         </dependency>
+        <dependency>
```
lor6: is this a new dependency added here?

amit2103: This dependency was present. I have moved the version into properties tag.

PR #8546, opened 2020-01-18
move versions in pom properties tag

amit2103:


amit2103:
There is a build failure for spring-roo module. Following is the error
Access denied to: http://repo.spring.io/spring-roo

lor6:
retest this please

PR #8550, opened 2020-01-19
Bael 3460

mguarnaccia:


libraries-3/src/main/java/com/baeldung/airline/CommandLine.java
```diff
@@ -0,0 +1,17 @@
...
+
+import com.github.rvesse.airline.annotations.Cli;
+import com.github.rvesse.airline.help.Help;
+
+@Cli(name = "baeldung-cli", 
+description = "Baeldung Airline Tutorial",
+defaultCommand = Help.class, 
+commands = { DatabaseSetupCommand.class, LoggingCommand.class, Help.class })
+public class CommandLine {
+
```
KevinGilmore: Please use the standard formatter on all the files and correct the indentation. You're using a combination of tabs and spaces.

libraries-3/pom.xml
```diff
@@ -23,6 +23,11 @@
             <artifactId>lombok</artifactId>
             <version>${lombok.version}</version>
         </dependency>
+        <dependency>
+		        <groupId>com.github.rvesse</groupId>
```
KevinGilmore: indent only with spaces

libraries-3/src/main/java/com/baeldung/airline/CommandLine.java
mguarnaccia: Fixed

libraries-3/pom.xml
mguarnaccia: Fixed. Also moved airline version to a specific parameter

sjmillington:
test this please

PR #8556, opened 2020-01-20
BAEL-3454: Guide to AVL trees in Java

maryarm:


algorithms-miscellaneous-5/src/main/java/com/baeldung/algorithms/balancedbinarytree/AVLTree.java
```diff
@@ -0,0 +1,140 @@
+package com.baeldung.algorithms.balancedbinarytree;
+
+public class AVLTree {
+
+    public static class Node {
+        int key;
+        int height;
+        Node left, right;
```
eric-martin: sep lines for the two variables

maryarm: Fixed.

PR #8587, opened 2020-01-25
Locking files in Java

psoares:


JonCook:
retest this please

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
```diff
@@ -0,0 +1,239 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Charsets;
+
+
+public class FileLocks {
+	
+	private static Logger log = LoggerFactory.getLogger(FileLocks.class);
```
JonCook: Let's make the logger final and prefer uppercase for the variable name

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
```diff
@@ -0,0 +1,239 @@
...
+	/**
+	 * Trying to get an exclusive lock on a read-only FileChannel won't work.
+	 */
+	static void getExclusiveLockFromInputStream() throws IOException, NonWritableChannelException {
+		Path path = Files.createTempFile("foo", "txt");
+		try (FileInputStream fis = new FileInputStream(path.toFile()); FileLock lock = fis.getChannel().lock()) {
+			log.debug("This won't happen");
+		} catch (NonWritableChannelException e) {
+			log.error(
+					"The channel obtained through a FileInputStream isn't writable. "
```
JonCook: Why do we need to concat the string here?

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
```diff
@@ -0,0 +1,239 @@
...
+		Path path = Files.createTempFile("foo", "txt");
+		try (RandomAccessFile file = new RandomAccessFile(path.toFile(), "rw");
+				FileLock lock = file.getChannel().lock(from, size, false)) {
+			if (lock.isValid()) {
+				log.debug("This is a valid exclusive lock");
+				return lock;
+			}
+			return null;
+		} catch (Exception e) {
+			System.out.println(e.getMessage());
```
JonCook: Here we have a mixture of log statement and system.out.println.

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
```diff
@@ -0,0 +1,239 @@
...
+			throw e;
+		}
+	}
+
+	
+	/**
+	 * Getting an exclusive lock from a RandomAccessFile works if the file is in write mode.
+	 * @param from beginning of the locked region
+	 * @param size how many bytes to lock
+	 * @return
```
JonCook: If we are going to use JavaDoc at all, could we fill in the return types.

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
```diff
@@ -0,0 +1,239 @@
...
+	 */
+	static FileLock getReadLockFromRandomAccessFile(long from, long size) throws IOException {
+		Path path = Files.createTempFile("foo", "txt");
+		try (RandomAccessFile file = new RandomAccessFile(path.toFile(), "r"); // could also be "rw", but "r" is sufficient for reading
+				FileLock lock = file.getChannel().lock(from, size, true)) {
+			if (lock.isValid()) {
+				log.debug("This is a valid shared lock");
+				return lock;
+			}
+			return null;
```
JonCook: This return null, seems to be redundant as repeated again later.

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
```diff
@@ -0,0 +1,239 @@
...
+							channel.write(buffer, channel.size());
+						}
+					}
+				} catch (OverlappingFileLockException e) {
+					// Failed to lock. Try again later.
+					Thread.sleep(50);
+					lockAndWrite();
+				}
+			} catch (IOException e) {
+				// TODO Auto-generated catch block
```
JonCook: Can we remote the TODO type statements please.

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
```diff
@@ -0,0 +1,239 @@
...
+		Stream.generate(() -> new Thread(new Writer(path, "foo " + count.incrementAndGet(), countDownLatch)))
+				.limit(concurrentWriters).forEach(Thread::start);
+
+		countDownLatch.await();
+		AtomicInteger lineCount = new AtomicInteger(0);
+		Files.lines(path).forEach((line) -> {
+			lineCount.incrementAndGet();
+			System.out.println(line);
+		});
+		log.info("Total lines written = " + lineCount.get());
```
JonCook: With slf4j we don't need to use the + to concat we can use for example:
logger.debug("Printing variable value: {}", variable);

Which is a bit more elegant.

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
```diff
@@ -0,0 +1,239 @@
...
+	/**
+	 * Getting a write lock on a file region
+	 */
+	static FileLock getExclusiveLockFromFileChannelOpen(long from, long size) throws IOException {
+		Path path = Files.createTempFile("foo", "txt");
+		try (FileChannel channel = FileChannel.open(path, StandardOpenOption.APPEND);
+				FileLock lock = channel.lock(from, size, false)) {
+			String text = "Hello, world.";
+			ByteBuffer buffer = ByteBuffer.allocate(text.length() + System.lineSeparator().length());
+			buffer.put((text + System.lineSeparator()).getBytes(Charsets.UTF_8));
```
JonCook: Charsets is a guava class, was that deliberate? as this article is about NIO I think it would be better to use..

- StandardCharsets.UTF_8

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
```diff
@@ -0,0 +1,239 @@
...
+					}
+				} catch (OverlappingFileLockException e) {
+					// Failed to lock. Try again later.
+					Thread.sleep(50);
+					lockAndWrite();
+				}
+			} catch (IOException e) {
+				// TODO Auto-generated catch block
+				e.printStackTrace();
+			}
```
JonCook: Could we also avoid calls to printStackTrace and use a logger?

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
```diff
@@ -0,0 +1,239 @@
...
+
+		}
+
+	}
+
+	public static void main(String[] args) throws InterruptedException, IOException {
+		Path path = Paths.get("/tmp/foo");
+		Files.deleteIfExists(path);
+		Files.createFile(path);
```
JonCook: Is this main method needed and the Writer static class? They aren't called from the tests or mentioned in the article at all?

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
psoares: No, this was because I was not using the right formatting. I used concatenation to break the statement into 2 lines because, usually, I limit the width of lines in my style. I'll fix this.

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
psoares: I removed the code that wasn't used in unit tests altogether since that was a remainder from previous versions of the article and wasn't being referenced in the article.

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
psoares: This code was deleted too, for the same reason.

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
psoares: This code was deleted.

core-java-modules/core-java-nio-2/src/main/java/com/baeldung/lock/FileLocks.java
psoares: No. I deleted both the main method and the Writer static class.

JonCook:
Replaced by https://github.com/eugenp/tutorials/pull/8688

PR #8603, opened 2020-01-26
BAEL-20858 move spring boot camel

Maiklins:


spring-boot-modules/spring-boot-camel/pom.xml
```diff
@@ -11,6 +11,7 @@
         <groupId>com.baeldung</groupId>
         <artifactId>parent-modules</artifactId>
         <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../../parent-modules</relativePath>
```
dkapil: if the parent is parent-modules, then we can easily switch it to parent-modules as well, let's do that.

Maiklins: do you mean to change it to parent-boot-2?

dkapil: @Maiklins sorry, I meant we can change the parent to the spring-boot-submodules i.e. the actual parent

Maiklins: Done

PR #8620, opened 2020-01-28
BAEL-21545 POM Properties Cleanup

dkapil:
- Cleaning pom.xml for project names starting with p,q and r


quarkus-extension/quarkus-app/pom.xml
```diff
@@ -23,12 +28,11 @@
         <dependency>
             <groupId>com.baeldung.quarkus.liquibase</groupId>
             <artifactId>quarkus-liquibase-runtime</artifactId>
-            <version>1.0-SNAPSHOT</version>
+            <version>${project.version}</version>
```
lor6: is this property defined already?

dkapil: yes, it will automatically redirect to its parent extension `quarkus-extension` and will fetch version from there

PR #8624, opened 2020-01-28
BAEL-20864 move spring boot customstarter

Maiklins:


spring-boot-modules/spring-boot-custom-starter/greeter-spring-boot-autoconfigure/pom.xml
```diff
@@ -7,10 +7,10 @@
     <name>greeter-spring-boot-autoconfigure</name>
 
     <parent>
-        <groupId>com.baeldung</groupId>
+        <groupId>com.baeldung.spring-boot-modules</groupId>
         <artifactId>spring-boot-custom-starter</artifactId>
         <version>0.0.1-SNAPSHOT</version>
-        <relativePath>../../spring-boot-custom-starter</relativePath>
+        <relativePath>..</relativePath>
```
dkapil: do we even need this if our parent is the immediate directory above? let's remove it from here and other poms in this PR as well where it is not needed.

PR #8724, opened 2020-02-14
Bael 3850: EnabledForJRERange and DisabledForJRERange JUnit Annotations

mona-mohamadinia:
Added two tests and removed a few deprecated stuff.


testing-modules/junit5-annotations/src/test/java/com/baeldung/junit5/conditional/ConditionalAnnotationsUnitTest.java
```diff
@@ -104,16 +82,4 @@ public void notValidForIntelliJ() {
...
-    @Retention(RetentionPolicy.RUNTIME)
-    @DisabledIf("Math.random() >= 0.5")
-    @interface CoinToss {
-    }
-
-    @RepeatedTest(2)
-    @CoinToss
-    public void gamble() {
-        System.out.println("This tests run status is a gamble with %50 rate");
-    }
```
ashleyfrieze: Rather than deleting these, please can you comment them out with a comment that they'll only work if you're running JUnit Jupiter 5.5 or lower.

PR #8733, opened 2020-02-16
Hexagonal Architecture in Java - Code Example

omerfinger:
a code example for my evaluation article - A Quick and Practical Example of Hexagonal Architecture in Java


omerfinger:
need to update BDD convention in tests

PR #8758, opened 2020-02-19
BAEL-3855

mguarnaccia:
Lock Striping


lock-striping/src/com/baeldung/concurrent/lock/BenchMark.java
```diff
@@ -0,0 +1,66 @@
...
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+
+@State(Scope.Thread)
+@Warmup(iterations = 2, time = 1, timeUnit = TimeUnit.MILLISECONDS)
+@Measurement(iterations = 3, time = 1, timeUnit = TimeUnit.MILLISECONDS)
+public class BenchMark {
+    ConcurrentAccessMap accessMyMap;
```
rpvilao: fix indentation throughout the class

lock-striping/src/com/baeldung/concurrent/lock/ConcurrentAccessMap.java
```diff
@@ -0,0 +1,33 @@
+package com.baeldung.concurrent.lock;
+
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+import com.google.common.base.Supplier;
+
+public abstract class ConcurrentAccessMap {
```
rpvilao: indentation

lock-striping/src/com/baeldung/concurrent/lock/ConcurrentAccessMap.java
```diff
@@ -0,0 +1,33 @@
...
+		
+	public ConcurrentAccessMap(Map<String, String> map) {
+		this.map = map;
+	}
+	
+	public final void doWork(String type) {
+		requests = new CompletableFuture<?>[THREADS * SLOTS];
+
+		for (int i = 0; i < THREADS; i++) {
+        	requests[SLOTS * i + 0] = CompletableFuture.supplyAsync(putSupplier(i));
```
rpvilao: indentation

lock-striping/src/com/baeldung/concurrent/lock/ConcurrentAccessMap.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public final void doWork(String type) {
+        requests = new CompletableFuture<?>[THREADS * SLOTS];
+
+        for (int i = 0; i < THREADS; i++) {
+            requests[SLOTS * i + 0] = CompletableFuture.supplyAsync(putSupplier(i));
+            requests[SLOTS * i + 1] = CompletableFuture.supplyAsync(getSupplier(i));
+            requests[SLOTS * i + 2] = CompletableFuture.supplyAsync(getSupplier(i));
+            requests[SLOTS * i + 3] = CompletableFuture.supplyAsync(getSupplier(i)); 	
+		}
```
rpvilao: fix indentation

mguarnaccia: Updated now

rpvilao:
@mguarnaccia maybe I was not clear enough before. you named the module, but the folder is still called lock-stripping and it's still in the root of the project. please take a look again to my previous comments. thanks!

mguarnaccia:
> @mguarnaccia maybe I was not clear enough before. you named the module, but the folder is still called lock-stripping and it's still in the root of the project. please take a look again to my previous comments. thanks!

@rpvilao Take a look, there's a new folder on the root "core-java-concurrency-collections-2" and code is inside java/main folder. hope I caught your request now

rpvilao:
thanks! almost there :p https://github.com/eugenp/tutorials/tree/master/core-java-modules
it needs to be inside this module

mguarnaccia:
Fixed

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/ConcurrentAccessBenchmark.java
```diff
@@ -0,0 +1,49 @@
...
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+
+@State(Scope.Thread)
+@Fork(value = 1)
+@Warmup(iterations = 0)
+public class ConcurrentAccessBenchmark {
+    ConcurrentAccessExperiment accessMyMap;
```
jzheaux: I think this would be simplified by having both instances of your test instantiated from the beginning:

```suggestion
    SingleLock single = new SingleLock();
    StripedLock striped = new StripedLock(BUCKETS);
```

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/ConcurrentAccessBenchmark.java
```diff
@@ -0,0 +1,49 @@
...
+            case "Striped Lock":
+                accessMyMap = new StripedLock(BUCKETS);
+                break;
+        }
+    }
+
+    @Benchmark
+    @BenchmarkMode(Mode.Throughput)
+    @OutputTimeUnit(TimeUnit.MILLISECONDS)
+    public void test() throws InterruptedException {
```
jzheaux: The JMH documentation states that benchmark methods should return a value in order to protect against dead code elimination. Maybe `doWork` could return the resulting `Map`.

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/ConcurrentAccessBenchmark.java
```diff
@@ -0,0 +1,49 @@
...
+            case "Single Lock":
+                accessMyMap = new SingleLock();
+                break;
+            case "Striped Lock":
+                accessMyMap = new StripedLock(BUCKETS);
+                break;
+        }
+    }
+
+    @Benchmark
```
jzheaux: I believe that having more than one benchmark method would simplify both `ConcurrentAccessBenchmark` and `ConcurrentAccessExperiment`. For example:

```java
@Benchmark
public Map<String, String> singleLockHashMap() {
    return this.single.doWork(new HashMap<>(), THREADS, SLOTS);
}

@Benchmark
public Map<String, String> stripedLockHashMap() {
    return this.striped.doWork(new HashMap<>(), THREADS, SLOTS);
}

@Benchmark
public Map<String, String> singleLockConcurrentHashMap() {
    return this.single.doWork(new ConcurrentHashMap<>(), THREADS, SLOTS);
}

@Benchmark
public Map<String, String> stripedLockHashMap() {
    return this.striped.doWork(new ConcurrentHashMap<>(), THREADS, SLOTS);
}
```

Then, you wouldn't need the `setup` method in this class nor in the experiment class. You could also then remove the `lockType` parameter and wouldn't need a direct reference to `ConcurrentAccessExperiment`.

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/ConcurrentAccessExperiment.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+
+import com.google.common.base.Supplier;
+
+public abstract class ConcurrentAccessExperiment {
+
+    public ConcurrentAccessExperiment() {
```
jzheaux: Let's remove this since it's redundant

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/SingleLock.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public SingleLock() {
+        lock = new ReentrantLock();
+    }
+
+    protected synchronized Supplier<?> putSupplier(Map<String,String> map, int key) {
+        return (()-> {
+            boolean done = false;
+            try {
+                while(!done) {
```
jzheaux: Why not simply use `lock.lock()`? That seems to be the default suggestion from [the JavaDoc](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantLock.html).

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/SingleLock.java
```diff
@@ -0,0 +1,44 @@
...
+import com.google.common.base.Supplier;
+
+public class SingleLock extends ConcurrentAccessExperiment {
+    ReentrantLock lock;
+
+    public SingleLock() {
+        lock = new ReentrantLock();
+    }
+
+    protected synchronized Supplier<?> putSupplier(Map<String,String> map, int key) {
```
jzheaux: I'm not sure I understand the point of these methods being `synchronized` if they are using programmatic locks. I've always understood that these are alternatives - you can use locks instead of `synchronized` as an optimization.

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/ConcurrentAccessBenchmark.java
mguarnaccia: Now both are inited on test call

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/ConcurrentAccessBenchmark.java
mguarnaccia: Done!

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/ConcurrentAccessExperiment.java
mguarnaccia: Done

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/SingleLock.java
mguarnaccia: I was trying to keep both *Lock classes the more alike. Also, without this, we should stop and resume tasks that couldn't acquire the lock: that would be out of scope for this tutorial.

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/ConcurrentAccessBenchmark.java
mguarnaccia: Done!

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/SingleLock.java
mguarnaccia: ok, it was added from a previous thread we had. removed now.

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/SingleLock.java
mguarnaccia: done, but I had to replicate the behavior in both *Lock classes

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/SingleLock.java
```diff
@@ -0,0 +1,46 @@
...
+        lock = new ReentrantLock();
+    }
+
+    protected Supplier<?> putSupplier(Map<String,String> map, int key) {
+        return (()-> {
+            try {
+                lock.lock();
+                map.put("key" + key, "value" + key);
+            } catch (Exception e) {
+                this.putSupplier(map, key);
```
jzheaux: Wouldn't this be an infinite loop? Seems risky to just infinitely try again. I'd recommend doing:

```java
lock.lock();
try {
    return map.put("key" + key, "value" + key);
} finally {
    lock.unlock();
}
```
This is what the JavaDoc recommends.

Also, if you return the result of `map.put`, then you don't need the extra `return null;` at the bottom.

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/StripedLock.java
```diff
@@ -0,0 +1,51 @@
...
+public class StripedLock extends ConcurrentAccessExperiment {
+    Striped<Lock> stripedLock;
+
+    public StripedLock(int buckets) {
+        stripedLock = Striped.lock(buckets);
+    }
+
+    protected Supplier<?> putSupplier(Map<String,String> map, int key) {
+        return (()-> {
+        	int bucket = key % stripedLock.size();
```
jzheaux: Please check the indentation here.

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/ConcurrentAccessBenchmark.java
```diff
@@ -0,0 +1,52 @@
...
+    ConcurrentAccessExperiment accessMyMap;
+    static final int SLOTS = 4;
+    static final int THREADS = 10000;
+    static final int BUCKETS = Runtime.getRuntime().availableProcessors() * SLOTS;
+
+    @Benchmark
+    @BenchmarkMode(Mode.Throughput)
+    @OutputTimeUnit(TimeUnit.MILLISECONDS)
+    public void singleLockHashMap() throws InterruptedException {
+        (new SingleLock()).doWork(new HashMap<String,String>(), THREADS, SLOTS);
```
jzheaux: I don't understand why you are re-instantiated the object each time. Won't that play into your results?

It seems like what you want to do is just reuse one already instantiated at the field level. That way, your benchmark is only about the work being done on the map and not about that plus the cost of instantiating the worker object.

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/ConcurrentAccessBenchmark.java
```diff
@@ -0,0 +1,52 @@
...
+public class ConcurrentAccessBenchmark {
+    ConcurrentAccessExperiment accessMyMap;
+    static final int SLOTS = 4;
+    static final int THREADS = 10000;
+    static final int BUCKETS = Runtime.getRuntime().availableProcessors() * SLOTS;
+
+    @Benchmark
+    @BenchmarkMode(Mode.Throughput)
+    @OutputTimeUnit(TimeUnit.MILLISECONDS)
+    public void singleLockHashMap() throws InterruptedException {
```
jzheaux: These need to return values, in accordance with JMH's guidelines.

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/SingleLock.java
mguarnaccia: Well, what if the lock is not acquired? the thread won't execute and we lose a put/get

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/ConcurrentAccessBenchmark.java
mguarnaccia: I'm sorry but I don't get the point. Instantiating the object is not such a big cost, also it appears in every test. If we want to instantiate the maps before, we should also remove the "put" phase inside the test.

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/ConcurrentAccessBenchmark.java
mguarnaccia: Sorry, I only changed the doWork method. I'll fix it.

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/StripedLock.java
mguarnaccia: Done

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/SingleLock.java
jzheaux: `lock` will block until the lock is acquired. From the JavaDoc:

> If the lock is held by another thread then the current thread becomes disabled for thread scheduling purposes and lies dormant until the lock has been acquired, at which time the lock hold count is set to one.

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/ConcurrentAccessBenchmark.java
jzheaux: To be clear, I'm only referring to `new SingleLock()` and by association `new StripedLock()`.

The reason we should remove the instantiation is that it is unnecessary. What are you gaining from instantiating it here?

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/SingleLock.java
jzheaux: The thread will block until the lock is acquired. From the JavaDoc:

> If the lock is held by another thread then the current thread becomes disabled for thread scheduling purposes and lies dormant until the lock has been acquired, at which time the lock hold count is set to one.

I guess the question I have is, what is so different about this use case that we'd deviate from the JavaDoc?

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/ConcurrentAccessBenchmark.java
mguarnaccia: Ok

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/SingleLock.java
mguarnaccia: Ok

core-java-modules/core-java-concurrency-collections-2/src/main/java/com/baeldung/concurrent/lock/SingleLock.java
```diff
@@ -0,0 +1,36 @@
...
+    ReentrantLock lock;
+
+    public SingleLock() {
+        lock = new ReentrantLock();
+    }
+
+    protected Supplier<?> putSupplier(Map<String,String> map, int key) {
+        return (()-> {
+            try {
+                lock.lock();
```
jzheaux: We don't want to lock inside the try (see the [JavaDoc](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantLock.html)). The reason is that if lock.lock() did in fact throw a RuntimeException, then the unlock could cause some unexpected results (since it's re-entrant, it could unlock one more level than intended, for example).

mguarnaccia: Ok!

PR #8765, opened 2020-02-22
Java-125 move spring boot parent

Maiklins:


spring-boot-modules/spring-boot-parent/pom.xml
```diff
@@ -13,7 +13,7 @@
         <groupId>com.baeldung</groupId>
         <artifactId>parent-modules</artifactId>
         <version>1.0.0-SNAPSHOT</version>
-        <relativePath>..</relativePath>
+        <relativePath>../../parent-boot-2</relativePath>
```
dkapil: shouldn't parent be spring-boot-modules and no relative path will be needed in that case?

PR #8775, opened 2020-02-25
BAEL-3497 IntelliJ: Doesn't Resolve Custom Spring Boot Properties

oskar9247:


spring-boot-properties/src/main/java/com/baeldung/configuration/processor/CustomProperties.java
```diff
@@ -0,0 +1,40 @@
+package com.baeldung.configuration.processor;
+
+import org.springframework.boot.context.properties.*;
+import org.springframework.context.annotation.*;
+
+@Configuration
+@ConfigurationProperties(prefix = "com.baeldung")
+public class CustomProperties
+{
```
ashleyfrieze: This class has non standard use of `{}` - please follow the formatting guides

spring-boot-properties/src/main/java/com/baeldung/configuration/processor/PropertyBeanInjection.java
```diff
@@ -0,0 +1,25 @@
...
+
+    String getProperty1Value()
+    {
+        return customProperties.getProperty1();
+    }
+
+    String getProperty2Value()
+    {
+        return customProperties.getProperty2();
+    }
```
ashleyfrieze: `{}` positioning again

spring-boot-properties/src/test/resources/configuration-processor.properties
```diff
@@ -0,0 +1,2 @@
+com.baeldung.property1 = test1
+com.baeldung.property2 = test2
```
ashleyfrieze: I think it would be better if we could use more realistic looking properties with good descriptions. For example, `url` and `timeout`

spring-boot-modules/spring-boot-properties/pom.xml
```diff
@@ -43,6 +43,11 @@
             <artifactId>httpcore</artifactId>
             <version>${httpcore.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-configuration-processor</artifactId>
+            <version>${configuration-processor.version}</version>
+        </dependency>
```
KevinGilmore: need to add the optional element
`<optional>true></optional>`

PR #8793, opened 2020-02-29
How to compare json objects using gson article

omerfinger:
Compare Two JSON Objects with Gson article

### DIFF
--- a/gson/src/test/java/com/baeldung/gson/jsoncompare/JsonCompareUnitTest.java
+++ b/gson/src/test/java/com/baeldung/gson/jsoncompare/JsonCompareUnitTest.java
@@ -1,0 +1,93 @@
+package com.baeldung.gson.jsoncompare;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonParser;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class JsonCompareUnitTest {
+    @Test
+    public void givenIdenticalSimpleObjects_whenCompared_thenEqual() {
+        JsonParser parser = new JsonParser();
+        String string1 = "{\"customer\": {\"id\": \"44521\",\"fullName\": \"Emily Jenkins\", \"age\": 27 }}";
+        String string2 = "{\"customer\": {\"id\": \"44521\", \"fullName\": \"Emily Jenkins\",\"age\": 27}}";
+
+        Assert.assertTrue(parser.parse(string1)
+            .isJsonObject());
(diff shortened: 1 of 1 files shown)


gson/src/test/java/com/baeldung/gson/jsoncompare/JsonCompareUnitTest.java
```diff
@@ -0,0 +1,93 @@
...
+public class JsonCompareUnitTest {
+    @Test
+    public void givenIdenticalSimpleObjects_whenCompared_thenEqual() {
+        JsonParser parser = new JsonParser();
+        String string1 = "{\"customer\": {\"id\": \"44521\",\"fullName\": \"Emily Jenkins\", \"age\": 27 }}";
+        String string2 = "{\"customer\": {\"id\": \"44521\", \"fullName\": \"Emily Jenkins\",\"age\": 27}}";
+
+        Assert.assertTrue(parser.parse(string1)
+            .isJsonObject());
+        Assert.assertEquals(parser.parse(string1), parser.parse(string2));
```
ashleyfrieze: Let's static import the assertions inside `org.junit.Assert` so we don't need to prefix them with `Assert.`

gson/src/test/java/com/baeldung/gson/jsoncompare/JsonCompareUnitTest.java
```diff
@@ -0,0 +1,93 @@
...
+    @Test
+    public void givenNestedObjectsDifferentArrayOrder_whenCompared_thenNotEqual() {
+        JsonParser parser = new JsonParser();
+        String string1 = "{\"customer\": {\"id\": \"44521\",\"fullName\": \"Emily Jenkins\", \"age\": 27, \"consumption_info\" : {\"last_buy\": \"2012-04-23\", \"prouducts\": [\"banana\", \"eggs\"]}}}";
+        String string2 = "{\"customer\": {\"id\": \"44521\",\"fullName\": \"Emily Jenkins\", \"age\": 27, \"consumption_info\" : {\"last_buy\": \"2012-04-23\", \"prouducts\": [\"eggs\", \"banana\"]}}}";
+
+        JsonElement json1 = parser.parse(string1);
+        JsonElement json2 = parser.parse(string2);
+
+        Assert.assertTrue(json1.isJsonObject());
```
ashleyfrieze: I don't think we need to repeatedly assert `isJsonObject` in all the tests. After the first one, we've proved the point. Let's only have it for the first time it's demonstrated.

PR #8797, opened 2020-03-01
Upgrade springdoc-openapi version to v1.2.32

bnasslahsen:
The objective is to make the configuration of springdoc-openapi simple for spring-boot projects.

Upgrading to 1.2.32, helps projects to use all the new springdoc-openapi properties:
 - https://springdoc.github.io/springdoc-openapi-demos/springdoc-properties.html


lor6:
Thanks @bnasslahsen  We've upgraded the version in the meantime.

PR #8810, opened 2020-03-02
BAEL-3900 Fixed ApplicationLiveTest.

mathieufortin01:
CSRF was messing the test, and security config for that component was missing. Added csrf().disable() (as this was set for other components in the same module). I assume there is nothing to change in the related article.

### DIFF
--- a/spring-security-modules/spring-security-mvc-boot-1/src/main/java/org/baeldung/custom/config/SecurityConfig.java
+++ b/spring-security-modules/spring-security-mvc-boot-1/src/main/java/org/baeldung/custom/config/SecurityConfig.java
@@ -2,12 +2,25 @@ package org.baeldung.custom.config;
 
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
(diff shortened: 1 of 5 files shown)


spring-security-modules/spring-security-mvc-boot-1/src/main/java/org/baeldung/custom/config/SecurityConfig.java
```diff
@@ -2,12 +2,25 @@
...
 import org.springframework.security.crypto.password.PasswordEncoder;
 
 @Configuration
-public class SecurityConfig {
+@EnableWebSecurity
+public class SecurityConfig extends WebSecurityConfigurerAdapter {
 
+    @Override
+    protected void configure(final HttpSecurity http) throws Exception {
+        http.csrf()
```
JonCook: Can we format this new block properly using the appropriate formatter e.g eclipse:

- https://github.com/eugenp/tutorials/blob/master/eclipse/formatter.xml

mathieufortin01: Done. Sorry about this omission.

PR #8814, opened 2020-03-05
BAEL-2552 - spatialguru.net@gmail.com

markathomas:
- Add controller and unit test showing how to set timeouts

### DIFF
--- a/spring-5-webflux/src/test/java/com/baeldung/spring/serverconfig/TimeoutLiveTest.java
+++ b/spring-5-webflux/src/test/java/com/baeldung/spring/serverconfig/TimeoutLiveTest.java
@@ -6,7 +6,9 @@ import javax.net.ssl.SSLException;
 
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.test.context.SpringBootTest.WebEnvironment;
@@ -35,6 +37,9 @@ public class TimeoutLiveTest {
 
     private WebTestClient webTestClient;
 
+    @Rule
+    public ExpectedException exception = ExpectedException.none();
+
     @Before
     public void setup() throws SSLException {
         webTestClient = WebTestClient.bindToServer(getConnector())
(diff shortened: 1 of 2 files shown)


spring-5-webflux/src/test/java/com/baeldung/spring/serverconfig/TimeoutLiveTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Before
+    public void setup() throws SSLException {
+        webTestClient = WebTestClient.bindToServer(getConnector())
+                                     .baseUrl(BASE_URL)
+                                     .build();
+    }
+
+    @Test
+    public void shouldTimeout() {
```
rpvilao: this sounds a bit hacky, the best would be to create a junit rule and before the call just record rule.expect(ReadTimeoutException.class)

spring-5-webflux/src/test/java/com/baeldung/spring/serverconfig/TimeoutLiveTest.java
```diff
@@ -0,0 +1,88 @@
...
+    public void shouldNotTimeout() {
+        try {
+            WebTestClient.ResponseSpec response = webTestClient.get()
+              .uri("/timeout/{timeout}", 1)
+              .exchange();
+            response.expectStatus()
+              .isOk()
+              .expectBody(String.class)
+              .isEqualTo("OK");
+        } catch (ReadTimeoutException e) {
```
rpvilao: I would remove this since the exception fails the test, and it's better to have that reported then to hide it

markathomas:
Thanks @rpvilao - I wasn't aware of the @Rule annotation.  Always used TestNG personally.  It has been changed per your request.

spring-5-webflux/src/test/java/com/baeldung/spring/serverconfig/TimeoutLiveTest.java
markathomas: Changes have been made.

spring-5-webflux/src/test/java/com/baeldung/spring/serverconfig/TimeoutLiveTest.java
markathomas: Changes have been made.

spring-5-webflux/src/test/java/com/baeldung/spring/serverconfig/TimeoutLiveTest.java
```diff
@@ -44,28 +49,22 @@ public void setup() throws SSLException {
...
-              .uri("/timeout/{timeout}", 3)
-              .exchange();
-            Assert.fail();
-        } catch (ReadTimeoutException ignore) {
-        }
+        exception.expect(ReadTimeoutException.class);
+        webTestClient.get()
+          .uri("/timeout/{timeout}", 3)
+          .exchange();
+        Assert.fail();
```
rpvilao: this can now be removed because if the exception is not thrown, JUnit it will fail it since it was expecting it

markathomas: Good point.  I forgot to remove that.  It is done.

PR #8818, opened 2020-03-06
Fix Spring Spring security Voter sample code bug in MvcConfig path

btesf:
There is a slight bug in the source code for this tutorial: https://www.baeldung.com/spring-security-acl


lor6:
Thanks @btesf  The packages have changed so there's a conflict now with the PR. I'll close it, but we'll look into the issue and add an internal fix if needed.

PR #8846, opened 2020-03-10
Guava MapMaker

ramkumarvenkat:


ramkumarvenkat:
Thanks @JonCook 

1) I did format it yesterday, but the formatting output looked weird with `@Test` in the same line as the test name. So I was not sure if it is the correct formatter. Anyways, I have formatted and pushed it now. Please let me know if this looks incorrect
2) and 3) - Done, thanks!
4) It was a bit off for me as well, as I wrote the tests. There is only a size() method, which reflects the number of values added in the map (and not the capacity). So it will always be 0. The other option I was thinking was to insert some values in the map and then asserting based off of it. But then, the scope of the article is about `MapMaker` and not about concurrrent hash maps. So I just kept it simple to test if the map is instantiated properly. Please let me know if you have a better idea!

ramkumarvenkat:
Hi @JonCook - I have fixed the formatting.

Please note that the IntelliJ formatter is still broken, tabs with 8 spacing is created as default and line wraps are not proper (doesn't seem to consider annotations): https://github.com/eugenp/tutorials/blob/master/intelliJ/intelliJ-formatter.xml#L37

Weirdly, the formatting doesn't work with the eclipse formatter as well (I am using the latest IntelliJ). 

But anyways, the way I fixed was: to look at the formatting options from a few file sthat were committed recently, kept changing the default values and applied the formatting till the committed files don't have any git diffs. Then I applied this new formatting to my file. I am not super happy with what I did, but atleast I have a version of the formatter that works consistently with other files in the repo.

PR #8848, opened 2020-03-10
BAEL-3881

pazis:
Gradient Descent in Java


algorithms-miscellaneous-6/pom.xml
```diff
@@ -0,0 +1,28 @@
...
+    </parent>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>${org.assertj.core.version}</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
```
KevinGilmore: Since we aren't using assertj in the unit test, we could remove it. Or, we could change the assertions to use assertj.

PR #8883, opened 2020-03-15
BAEL-3304 - spatialguru.net@gmail.com

markathomas:


spring-swagger-codegen/spring-openapi-generator-api-client/.gitignore
```diff
@@ -0,0 +1,21 @@
+*.class
```
KevinGilmore: Question: does the OpenAPI generator also generate the .gitignore file? If so, keep it, otherwise we don't need it because the one at the root of the repo will take care of things.

markathomas: Yes it does.  I'll remove it.

markathomas: Removed

PR #8893, opened 2020-03-16
Add embedded app server

AlexOreshkevich:
Added TomEE application server using `tomee-maven-plugin` that could quickly run application using `mvn tomee:run`.


lor6:
Thanks @AlexOreshkevich

PR #8894, opened 2020-03-16
BAEL-3912: Changes done

nachoverdu:


nachoverdu:
I have redone the examples according to the comments

PR #8900, opened 2020-03-17
BAEL-3918 - Adding BigDecimal numbers using the Stream API

fabripautasso:


core-java-modules/core-java-streams-3/src/main/java/com/baeldung/streams/bigdecimals/StreamAddNumbers.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.math.BigDecimal;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Random;
+import java.util.stream.IntStream;
+
+public class StreamAddNumbers {
+
+    public int addInts() {
```
ashleyfrieze: Illustrating these functions with unit tests would be preferable. This means you probably don't want to use `Random`. But you can add up the numbers 0, 1 & 2 from `IntStream.range`

core-java-modules/core-java-streams-3/src/main/java/com/baeldung/streams/bigdecimals/StreamAddNumbers.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public double addDoubleOnList() {
+        List<Double> doubleNumbers = Arrays.asList(23.48, 52.26, 13.5);
+        return doubleNumbers.stream()
+            .mapToDouble(Double::doubleValue)
+            .sum();
+    }
+
+    public int addNumbersWithReduce() {
+        List<Integer> intNumbers = Arrays.asList(5, 1, 100);
```
ashleyfrieze: Use `Stream.of` rather than `asList().stream()`

core-java-modules/core-java-streams-3/src/main/java/com/baeldung/streams/bigdecimals/StreamAddNumbers.java
fabripautasso: I moved the code to a unit test class

core-java-modules/core-java-streams-3/src/main/java/com/baeldung/streams/bigdecimals/StreamAddNumbers.java
fabripautasso: Done

core-java-modules/core-java-streams-3/src/test/java/com/baeldung/streams/bigdecimals/AddNumbersUnitTest.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.List;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+import org.junit.Test;
+
+public class AddNumbersUnitTest {
+
+    @Test
+    public void givenIntStream_whenSum_thenResultIsCorrect() {
```
ashleyfrieze, line 16: Note: I've renamed this method - the then doesn't describe what the test does, it describes the required outcome.

fabripautasso: Sounds good, thanks

PR #8909, opened 2020-03-18
fix launch run command for windows

piuserre:
In order to launch the application, is necessary to remove "./"


lor6:
Thanks @piuserre  Merged.

PR #8921, opened 2020-03-19
[JAVA-961] standardize packages in spring security modules - Part 2

dupirefr:


spring-security-modules/spring-security-mvc-custom/src/main/resources/webSecurityConfig.xml
```diff
@@ -31,10 +31,5 @@
                 <user name="admin1" password="{noop}admin1Pass" authorities="ROLE_ADMIN"/>
             </user-service>
         </authentication-provider>
-    </authentication-manager>
-
-<!--     <mvc:interceptors>
-        <bean id="loggerInterceptor" class="org.baeldung.web.interceptor.LoggerInterceptor" />
```
jzheaux: Should the package be updated here, too?

PR #8928, opened 2020-03-21
Bael 2398 couroutines in java

w0wka91:


coroutines-java/src/main/java/com/baeldung/App.java
```diff
@@ -0,0 +1,11 @@
+package com.baeldung;
```
glmartin: Let's move this code to an "introduction" package.  So the package should be `com.baeldung.introduction`.

coroutines-java/src/main/java/com/baeldung/introduction/App.java
```diff
@@ -0,0 +1,11 @@
+package com.baeldung;
```
glmartin: Lets repackage this as `com.baeldung.introduction`

PR #8969, opened 2020-03-26
BAEL-3926 - Java Map with case-insensitive keys

omerfinger:


omerfinger:
There is another pending PR who opened a new module with the exact name

eric-martin:
Retest this please

java-collections-maps-3/pom.xml
```diff
@@ -2,25 +2,39 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-    <modelVersion>4.0.0</modelVersion>
-    <artifactId>java-collections-maps-3</artifactId>
-    <version>0.1.0-SNAPSHOT</version>
-    <name>java-collections-maps-3</name>
-    <packaging>jar</packaging>
```
eric-martin, line 9: Can you undo your change here? The only change to the pom file should be the new deps and props.

omerfinger: Hi, yes I will, at the time this pom wasn't existing, so I created this file. I will delete it.
As for the name element, it said in the authors' documentation it should be deleted.

omerfinger: Also, packaging default value is jar.

PR #9017, opened 2020-04-02
Mockint objectmapper with mockito

pegeroad:


spring-mockito/src/test/java/com/baeldung/reader/FlowerReaderServiceUnitTest.java
```diff
@@ -0,0 +1,51 @@
...
+    @Test
+    public void whenCallingWithJsonAsString_thenReturnsExpectedFlowerObject() throws JsonProcessingException {
+
+        String flowerAsJson = "{\"name\": \"rose\", \"petals\": 100}";
+        String roseName = "rose";
+        Integer rosePetals = 100;
+        Flower rose = new Flower(roseName, rosePetals);
+
+        Mockito.when(objectMapper.readValue(flowerAsJson, Flower.class)).thenReturn(rose);
+        Flower expectedRose = flowerReader.readFlower(flowerAsJson);
```
KevinGilmore, line 34: It is not necessary to provide a specific string to specify the Mock behavior. You can simply use any(String.class) as the parameter to readValue instead of an actual string. Then you can specify any string you want when you actually make the call to your service's readFlower method.

spring-mockito/src/test/java/com/baeldung/reader/FlowerReaderServiceUnitTest.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Test
+    public void whenCallingWithIncorrectJson_thenThrowsException() throws JsonProcessingException {
+
+        String incorrectJson = "this is not a flower in JSON";
+
+        Mockito.when(objectMapper.readValue(incorrectJson, Flower.class)).thenThrow(JsonProcessingException.class);
+
+        Assertions.assertThrows(JsonProcessingException.class, () -> flowerReader.readFlower(incorrectJson));
```
KevinGilmore, line 48: Similarly, you can use any(String.class) here. But there is no reason to actually specify an invalid JSON string when calling your service. If you do, then what is the point of mocking the behavior? That is the actual exception that would be thrown when reading an invalid JSON string if you did not mock the readValue method.

spring-mockito/src/test/java/com/baeldung/reader/FlowerReaderServiceWithoutSpringUnitTest.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Test
+    public void whenCallingWithIncorrectJson_thenThrowsException() throws JsonProcessingException {
+
+        String incorrectJson = "this is not a flower in JSON";
+
+        Mockito.when(objectMapper.readValue(incorrectJson, Flower.class)).thenThrow(JsonProcessingException.class);
+
+        Assertions.assertThrows(JsonProcessingException.class, () -> flowerReader.readFlower(incorrectJson));
```
KevinGilmore, line 49: Same applies here as in the other test class. By the way, this reinforces my earlier thought that we don't need to involve Spring at all and should probably move this code to a different module.

spring-mockito/src/test/java/com/baeldung/reader/FlowerReaderServiceUnitTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Test
+    public void whenCallingWithJsonAsString_thenReturnsExpectedFlowerObject() throws JsonProcessingException {
+
+        String flowerAsJson = "{\"name\": \"rose\", \"petals\": 100}";
+        String roseName = "rose";
+        Integer rosePetals = 100;
+        Flower rose = new Flower(roseName, rosePetals);
+
+        Mockito.when(objectMapper.readValue(flowerAsJson, Flower.class)).thenReturn(rose);
```
ashleyfrieze, line 33: Note: we generally static import all mockito methods.

spring-mockito/src/test/java/com/baeldung/reader/FlowerReaderServiceUnitTest.java
```diff
@@ -0,0 +1,51 @@
...
+        Flower expectedRose = flowerReader.readFlower(flowerAsJson);
+
+        Assertions.assertEquals(roseName, expectedRose.getName());
+        Assertions.assertEquals(rosePetals, expectedRose.getPetals());
```
ashleyfrieze, line 37: and we generally static import assertions too

testing-modules/mockito-2/src/main/java/com/baeldung/mockito/objectmapper/FlowerJsonStringValidator.java
```diff
@@ -0,0 +1,21 @@
...
+        this.objectMapper = objectMapper;
+    }
+
+    public boolean flowerHasPetals(String jsonFlowerAsString) throws JsonProcessingException {
+        Flower flower = readFlower(jsonFlowerAsString);
+        return flower.getPetals() > 0;
+    }
+
+    private Flower readFlower(String flowerAsJson) throws JsonProcessingException {
+        return objectMapper.readValue(flowerAsJson, Flower.class);
```
KevinGilmore, line 19: Since we only use this once and it's a one-liner, let's just move it to the flowerHasPetals method to simplify the call stack.

PR #9071, opened 2020-04-10
[JAVA-616] Split core-java-arrays* modules

dupirefr:


core-java-modules/core-java-arrays-operations-advanced/README.md
```diff
@@ -0,0 +1,16 @@
...
+- Sum, average of elements
+- Intersections
+- ...
+
+### Relevant Articles: 
+- [How to Copy an Array in Java](https://www.baeldung.com/java-array-copy)
+- [Arrays.deepEquals](https://www.baeldung.com/java-arrays-deepequals)
+- [Find Sum and Average in a Java Array](https://www.baeldung.com/java-array-sum-average)
+- [Intersection Between two Integer Arrays](https://www.baeldung.com/java-array-intersection)
+- [[More basic operations-->]](/core-java-modules/core-java-operations-basic)
```
jzheaux, line 16: Does this wording make sense? I think I'd say "more basic operations" if I was on a "Basic operations" page. Maybe you mean "Basic Operations". Either way, I'd probably recommend that we leave links to just be within the same module set for now, so no link here.

core-java-modules/core-java-arrays-operations-basic/README.md
```diff
@@ -0,0 +1,19 @@
+## Core Java Arrays - Basic Operations
+
+This module contains articles about basic operations on arrays in Java
```
jzheaux, line 3: Let's try this:

```
This module contains articles about Java array fundamentals. They assume no previous background knowledge on working with arrays in Java.
```

I think listing specifics like iteration or retrieval is problematic since there are advanced ways to iterate, etc. I think the difference between basic and advanced is probably what the article expects you to already know before reading.

PR #9077, opened 2020-04-11
BAEL-3962 Downloading attachments using Java Mail API

chanchal-detani:
@ashleyfrieze : Your review is most valuable and important for me before the merge happens.

Hence kindly requesting to review and provide your valuable feedback.


mail-attachment-downloader/README.md
```diff
@@ -0,0 +1,7 @@
+## Mail-Attachment-Downloader
+
+This module contains articles about downloading attachment from email using pop3 protocol.
+
+### Relevant articles:
+
+- [A quick guide to downloading email attachments using Java Mail API](https://www.baeldung.com/)
```
ashleyfrieze, line 7: We don't add this to the README - it's done after publication.

You've also created a root-level project here. I think we should probably put it somewhere inside an existing module.

https://github.com/eugenp/tutorials/tree/master/libraries-3 seems a good place for me.

mail-attachment-downloader/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+	<name>mail-attachment-downloader</name>
+	<description>Utility to download attachments from email using java mail API</description>
+
+	<properties>
+		<maven.compiler.target>1.8</maven.compiler.target>
+		<maven.compiler.source>1.8</maven.compiler.source>
+	</properties>
+
+	<dependencies>
+		<!-- https://mvnrepository.com/artifact/com.sun.mail/javax.mail -->
```
ashleyfrieze, line 17: We can lose this comment.

mail-attachment-downloader/src/main/java/com/baeldung/mail/attachment/utils/EmailAttachmentReceiver.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public String getUserName() {
+        return userName;
+    }
+
+    public String getPassword() {
+        return password;
+    }
+
+    public void downloadEmailAttachmentsTo(String directory) throws NoSuchProviderException, MessagingException, IOException {
```
ashleyfrieze, line 51: Let's not throw three exceptions from a method. That's generally bad Java practice.

mail-attachment-downloader/src/main/java/com/baeldung/mail/attachment/utils/EmailAttachmentReceiver.java
```diff
@@ -0,0 +1,92 @@
...
+        for (Message message : messages) {
+            // filtering message with content type 'multipart'
+            if (message.getContentType()
+                .equalsIgnoreCase(MULTIPART)) {
+                Multipart multipart = (Multipart) message.getContent();
+                for (int partCount = 0; partCount < multipart.getCount(); partCount++) {
+                    MimeBodyPart mimeBodyPart = (MimeBodyPart) multipart.getBodyPart(partCount);
+                    if (Part.ATTACHMENT.equalsIgnoreCase(mimeBodyPart.getDisposition())) {
+                        // this is the attachment.
+                        mimeBodyPart.saveFile(directory + File.separator + mimeBodyPart.getFileName());
```
ashleyfrieze, line 83: We are three indents deep. This seems to be a good case for extracting a subroutine that does one things.

mail-attachment-downloader/src/main/java/com/baeldung/mail/attachment/utils/EmailAttachmentReceiver.java
```diff
@@ -0,0 +1,92 @@
...
+                    }
+                }
+            }
+        }
+        // closing open connections
+        inbox.close();
+        store.close();
```
ashleyfrieze, line 90: You didn't close the session.

Are all these classes Autocloseable? If so, shouldn't we be demonstrating their use with `try-with-resources`?

mail-attachment-downloader/src/test/java/com/baeldung/mail/attachment/ApplicationTests.java
```diff
@@ -0,0 +1,6 @@
+package com.baeldung.mail.attachment;
+
+public class ApplicationTests {
+    public static void main(String[] args) {
+    }
+}
```
ashleyfrieze, line 6: Looks like this isn't needed.

mail-attachment-downloader/README.md
chanchal-detani: These changes are now incorporated.

mail-attachment-downloader/pom.xml
chanchal-detani: Done.

mail-attachment-downloader/src/main/java/com/baeldung/mail/attachment/utils/EmailAttachmentReceiver.java
chanchal-detani: Done

mail-attachment-downloader/src/main/java/com/baeldung/mail/attachment/utils/EmailAttachmentReceiver.java
chanchal-detani: Only folder is auto closeable and other ones are not, hence using only try block.

mail-attachment-downloader/src/test/java/com/baeldung/mail/attachment/ApplicationTests.java
chanchal-detani: Done

mail-attachment-downloader/README.md
chanchal-detani: Done

mail-attachment-downloader/pom.xml
chanchal-detani: Done

mail-attachment-downloader/src/main/java/com/baeldung/mail/attachment/utils/EmailAttachmentReceiver.java
chanchal-detani: Done

mail-attachment-downloader/src/main/java/com/baeldung/mail/attachment/utils/EmailAttachmentReceiver.java
chanchal-detani: Done

libraries-3/pom.xml
```diff
@@ -87,6 +87,11 @@
             <artifactId>error_prone_core</artifactId>
             <version>2.1.3</version>
         </dependency>
+        <dependency>
+			<groupId>com.sun.mail</groupId>
+			<artifactId>javax.mail</artifactId>
+			<version>1.6.2</version>
+		</dependency>
```
ashleyfrieze, line 94: It looks like this has been indented with `tab`. Please indent with spaces

libraries-3/src/main/java/com/baeldung/mail/attachment/EmailAttachmentReceiver.java
```diff
@@ -0,0 +1,108 @@
...
+        Properties properties = new Properties();
+
+        // server settings
+        properties.put(String.format(MAIL_PROTOCOL_HOST, this.protocol), this.host);
+        properties.put(String.format(MAIL_PROTOCOL_PORT, this.protocol), this.port);
+
+        // SSL settings
+        properties.put(String.format(MAIL_PROTOCOL_SOCKET_FACTORY_CLASS, this.protocol), JAVAX_NET_SSL_SSL_SOCKET_FACTORY);
+        properties.put(String.format(MAIL_PROTOCOL_SOCKET_FACTORY_FALLBACK, this.protocol), FALSE);
+        properties.put(String.format(MAIL_PROTOCOL_SOCKET_FACTORY_PORT, this.protocol), this.port);
```
ashleyfrieze, line 61: As this method is an instance method and not the constructor, the `this` is optional and should probably be removed to reduce the amount for the reader to read.

libraries-3/src/main/java/com/baeldung/mail/attachment/EmailAttachmentReceiver.java
```diff
@@ -0,0 +1,108 @@
...
+        // obtaining sessions
+        Session session = Session.getDefaultInstance(properties);
+        try {
+                // connecting to the mail server
+                Store store = session.getStore(this.protocol);
+                store.connect(this.userName, this.password);
+
+            // opens the inbox folder
+            Folder inbox = store.getFolder(INBOX);
+            inbox.open(Folder.READ_ONLY);
```
ashleyfrieze, line 72: There's something odd about the indentation here - I think lines 66-68 need bringing in a level.

libraries-3/src/main/java/com/baeldung/mail/attachment/EmailAttachmentReceiver.java
```diff
@@ -0,0 +1,108 @@
...
+        properties.put(String.format(MAIL_PROTOCOL_SOCKET_FACTORY_CLASS, this.protocol), JAVAX_NET_SSL_SSL_SOCKET_FACTORY);
+        properties.put(String.format(MAIL_PROTOCOL_SOCKET_FACTORY_FALLBACK, this.protocol), FALSE);
+        properties.put(String.format(MAIL_PROTOCOL_SOCKET_FACTORY_PORT, this.protocol), this.port);
+
+        // obtaining sessions
+        Session session = Session.getDefaultInstance(properties);
+        try {
+                // connecting to the mail server
+                Store store = session.getStore(this.protocol);
+                store.connect(this.userName, this.password);
```
ashleyfrieze, line 68: Same point about use of `this`

libraries-3/src/main/java/com/baeldung/mail/attachment/EmailAttachmentReceiver.java
```diff
@@ -0,0 +1,108 @@
...
+            for (Message message : messages) {
+                // filtering message with content type 'multipart'
+                if (message.getContentType()
+                    .equalsIgnoreCase(MULTIPART)) {
+                    checkForAttachmentsAndSave(message, directory);
+                }
+            }
+            // closing open connections
+            inbox.close();
+            store.close();
```
ashleyfrieze, line 83: We're not using `try-with-resources`. We could do. If not, then these `close` methods would have to be in the `finally` block in case of exceptions. The mail API can throw a lot of exceptions.

I'd recommend using the try-with resources. `Folder` is definitely `AutoCloseable`

libraries-3/src/main/java/com/baeldung/mail/attachment/EmailAttachmentReceiver.java
```diff
@@ -0,0 +1,108 @@
...
+                // this is the attachment.
+                MimeBodyPart mimeBodyPart = (MimeBodyPart) multipart.getBodyPart(partCount);
+                mimeBodyPart.saveFile(directory + File.separator + mimeBodyPart.getFileName());
+            }
+        }
+    }
+
+    private boolean isAttachment(int partCount, Multipart multipart) throws MessagingException {
+        MimeBodyPart mimeBodyPart = (MimeBodyPart) multipart.getBodyPart(partCount);
+        return Part.ATTACHMENT.equalsIgnoreCase(mimeBodyPart.getDisposition());
```
ashleyfrieze, line 106: When we use `equalsIgnoreCase`, it has to be possible for mixed case answers to be valid. Is that the case here?

libraries-3/src/main/java/com/baeldung/mail/attachment/EmailAttachmentReceiver.java
```diff
@@ -0,0 +1,122 @@
...
+import javax.mail.internet.MimeBodyPart;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class EmailAttachmentReceiver {
+
+
+
+
```
ashleyfrieze, line 27: We have some unnecessary whitespace here - let's remove it.

libraries-3/src/main/java/com/baeldung/mail/attachment/EmailAttachmentReceiver.java
```diff
@@ -0,0 +1,122 @@
...
+    final Logger LOGGER = LoggerFactory.getLogger(EmailAttachmentReceiver.class);
```
ashleyfrieze, line 28: By convention this would also be `static`

libraries-3/src/main/java/com/baeldung/mail/attachment/EmailAttachmentReceiver.java
```diff
@@ -0,0 +1,122 @@
...
+        Folder inbox = null;
+        try {
+            // connecting to the mail server
+            store = session.getStore(IMAP);
+            store.connect(userName, password);
+
+            // opens the inbox folder
+            inbox = store.getFolder(INBOX);
+            inbox.open(Folder.READ_ONLY);
+            List<Message> messages = asList(inbox.getMessages());
```
ashleyfrieze, line 79: I'm not sure why you're making `getMessages` into a `List`. You can do the `for` loop on an array without creating a list instead. It seems like an unnecessary copy between types.

libraries-3/src/main/java/com/baeldung/mail/attachment/EmailAttachmentReceiver.java
```diff
@@ -0,0 +1,122 @@
...
+
+        // SSL settings
+        properties.put(MAIL_IMAP_SOCKET_FACTORY_CLASS, JAVAX_NET_SSL_SSL_SOCKET_FACTORY);
+        properties.put(MAIL_IMAP_SOCKET_FACTORY_FALLBACK, FALSE);
+        properties.put(MAIL_IMAP_SOCKET_FACTORY_PORT, port);
+
+        // obtaining sessions
+        Session session = Session.getDefaultInstance(properties);
+        Store store = null;
+        Folder inbox = null;
```
ashleyfrieze, line 70: We should be using the `try-with-resources` pattern here to avoid having to write the `finally` statement. The mail API supports `AutoCloseable` - https://javaee.github.io/javamail/docs/api/javax/mail/Store.html

libraries-3/src/main/java/com/baeldung/mail/attachment/MailAttachmentDownloaderApplication.java
```diff
@@ -0,0 +1,16 @@
+package com.baeldung.mail.attachment;
+
+public class MailAttachmentDownloaderApplication {
+    public static void main(String[] args) {
+         String host = "imap.gmail.com";
+         String port = "993";
```
ashleyfrieze, line 6: The indentation is not consistent.

libraries-3/src/main/java/com/baeldung/mail/attachment/EmailAttachmentReceiver.java
chanchal-detani: > We have some unnecessary whitespace here - let's remove it.

Done

libraries-3/src/main/java/com/baeldung/mail/attachment/EmailAttachmentReceiver.java
chanchal-detani: > By convention this would also be static

Done

libraries-3/src/main/java/com/baeldung/mail/attachment/EmailAttachmentReceiver.java
chanchal-detani: > I'm not sure why you're making getMessages into a List. You can do the for loop on an array without creating a list instead. It seems like an unnecessary copy between types.

Done

libraries-3/src/main/java/com/baeldung/mail/attachment/MailAttachmentDownloaderApplication.java
chanchal-detani: > The indentation is not consistent.

Done

libraries-3/src/main/java/com/baeldung/mail/attachment/EmailAttachmentReceiver.java
chanchal-detani: If I use try with resource, in that case also I need to use finally block to close folder instance since try-with resource feature will not allow me to right below statements.
try (Store store = session.getStore(IMAP);
      store.connect(userName, password); // THIS STATEMENT IS REQUIRED OTHERWISE EXCEPTION SAYING STORE NOT // CONNECTED WILL BE THROWN
      Folder inbox = store.getFolder(INBOX)) {} catch(E e) {}

@ashleyfrieze : Please guide me on this?

ashleyfrieze: `Store` and `Folder` are both `AutoCloseable`. This means you can use `try-with-resources` for both. However, you will need to nest them. Ideally use a second function:

```java
try (Store store = session.getStore(IMAP)) {
   store.connect(...);
   useStore(store);
}

.
.
.
void useStore(Store store) {
   try (Folder folder = store.getFolder(INBOX)) {
      // etc
   }
}
```

There are other solutions, but they become a bit messier. You could have a function that creates the store AND logs into it so you can appear to do it with one single try, but if the login fails, it may not close things correctly

```java
try (Store store = createStore();
    Folder folder = store.getFolder(INBOX)) {
   ...
}


.
.
.
Store createStore() {
    Store store = session.getStore(IMAP);
    store.connect(...);
}
```

chanchal-detani:
Since i no longer be writing the articles, why should i share my code.

PR #9093, opened 2020-04-13
BAEL-3896: OpenAPI JSON Objects in Query Params

alex07pep:
OpenAPI JSON Objects in Query Params by alexandruflorin.peptan@gmail.com


spring-rest-http/README.md
```diff
@@ -13,3 +13,4 @@ The "REST With Spring" Classes: http://bit.ly/restwithspring
 - [Spring RequestMapping](https://www.baeldung.com/spring-requestmapping)
 - [Guide to DeferredResult in Spring](https://www.baeldung.com/spring-deferred-result)
 - [Using JSON Patch in Spring REST APIs](https://www.baeldung.com/spring-rest-json-patch)
+- [OpenAPI JSON Objects in Query Params](https://www.baeldung.com/openapi-json-objects-in-query-params)
```
KevinGilmore, line 16: Please revert this change. Since we don't know the final URL yet, your editor will add this once the article is published.

spring-rest-http/src/main/java/com/baeldung/jsonparam/jsonParamOpenApiSwaggerDefinition
```diff
@@ -0,0 +1,154 @@
...
+          content:
+            application/json:
+              schema:
+              	type: string
+          		example: '{"type":"foo","color":"green"}'
+      responses:
+        '200':
+          description: successful process
+
+"/api/tickets": {
```
KevinGilmore, line 30: Two things starting with this line: First, the indentation is off. This path and the paths below should be indented the same as /tickets above. Second, why do we start putting everything in double quotes at this point? Although it is allowed, convention is to not use any quotes in an OpenAPI spec except where necessary.

spring-rest-http/src/main/java/com/baeldung/jsonparam/jsonParamOpenApiSwaggerDefinition
```diff
@@ -0,0 +1,154 @@
+swagger: '2.0'
+...
```
KevinGilmore, line 2: Are we putting the ... here and other places below as a shorthand notation to say that in a real spec, there would be more here? That's fine for snippets in the article, but the code in the repository should reflect complete and syntactically correct examples. In fact, the article has 2.0 and 3.0 examples, so we should have one file here for the 2.0 spec and one for the 3.0 spec. Both should be valid specs.

PR #9110, opened 2020-04-16
Updated dependency assertj.version as the old one wasn't found

cristianvasiliu:


lor6:
Thanks @cristianvasiliu

PR #9122, opened 2020-04-17
tested the code with latest testng version 7.1.0

sureshnath:
I have verified the existing code also works with 7.10


lor6:
Thanks @sureshnath

PR #9135, opened 2020-04-18
BAEL-3965: Created examples for instance and static factory methods.

albanoj2:


eric-martin:
Test this please

spring-core-4/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-tx</artifactId>
+            <version>${spring.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-jdbc</artifactId>
+            <version>${spring.version}</version>
+        </dependency>
```
KevinGilmore, line 31: Since we're creating a new module, let's limit the dependencies to just what we need for the article. Looks like we don't need these two, and there could be others.

PR #9136, opened 2020-04-18
JAVA-913: Migrate parent-boot-1 articles to parent-boot-2

sampada07:


jjwt/pom.xml
```diff
@@ -30,6 +30,11 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-security</artifactId>
         </dependency>
+        
+        <dependency>
```
dkapil, line 34: is it needed? when we used boot-1 was it included as part of any other library and with boot-2 we have to mention it explicitly?

sampada07: I added the spring-boot-starter-web explicitly since without this dependency, we are getting compile errors of the kind javax.servlet.http.* cannot be resolved.

PR #9145, opened 2020-04-19
[BAEL-3967] Spring JPA @Embedded and @EmbeddedId 

shockwave9999:
[BAEL-3967] Spring JPA @Embedded and @EmbeddedId


persistence-modules/spring-data-jpa-5/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+
+        <dependency>
+            <groupId>mysql</groupId>
+            <artifactId>mysql-connector-java</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.postgresql</groupId>
+            <artifactId>postgresql</artifactId>
+        </dependency>
```
KevinGilmore, line 39: We could omit these for now since we aren't using them. They can be added later if/when needed.

shockwave9999: "updated pom.xml"

PR #9155, opened 2020-04-20
[BAEL-3489] Added Java-R integration examples.

aurasphere:


libraries-data-2/pom.xml
```diff
@@ -137,6 +155,13 @@
             <url>http://repo.numericalmethod.com/maven/</url>
             <layout>default</layout>
         </repository>
+        
+        <!-- Needed for Renjin -->
+		<repository>
```
JonCook, line 160: The indentation seems to be out here?

libraries-data-2/pom.xml
```diff
@@ -153,6 +178,27 @@
         <assertj.version>3.6.2</assertj.version>
         <slf4j.version>1.7.25</slf4j.version>
         <awaitility.version>3.0.0</awaitility.version>
+        <renjin.version>RELEASE</renjin.version>
+        <rcaller.version>3.0</rcaller.version>
+        <rserve.version>1.8.1</rserve.version>
     </properties>
 
+	<build>
+		<plugins>
```
JonCook, line 187: Also here the indentation is incorrect.

PR #9160, opened 2020-04-21
BAEL-3972: check user roles in Java

michael-pratt:
Add examples of checking user roles in Java

### DIFF
--- a/spring-security-modules/spring-security-core/src/main/java/com/baeldung/app/controller/TaskController.java
+++ b/spring-security-modules/spring-security-core/src/main/java/com/baeldung/app/controller/TaskController.java
@@ -48,8 +48,7 @@ public class TaskController {
      */
     @GetMapping("/manager")
     @PreAuthorize("hasRole('ROLE_MANAGER')")
-    public ResponseEntity<Iterable<Task>> getAlManagerTasks()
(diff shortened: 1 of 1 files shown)


spring-security-modules/spring-security-core/src/main/java/com/baeldung/app/controller/TaskController.java
```diff
@@ -30,4 +42,66 @@
...
         return ResponseEntity.ok().body(tasks);
     }
+
+    /**
+     * Example of restricting specific endpoints to specific roles using @PreAuthorize.
+     */
+    @GetMapping("/manager")
+    @PreAuthorize("hasRole('ROLE_MANAGER')")
+    public ResponseEntity<Iterable<Task>> getAlManagerTasks()
+    {
```
KevinGilmore, line 52: For consistency with this class and other Java code on the site, let's use the convention/standard of putting the opening brace at the end of the line rather than on a line by itself. Same applies to the other methods, not just this one :)

michael-pratt: @KevinGilmore Should be updated now

PR #9170, opened 2020-04-24
BAEL-3863 Introduction to Finagle

mdabrowski-eu:


finagle/README.md
```diff
@@ -0,0 +1,7 @@
+## Finagle
+
```
rpvilao, line 2: let's remove it, it's added by us in the end as we don't know yet the final url

finagle/src/main/java/com/baeldung/finagle/GreetingService.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.twitter.finagle.Service;
+import com.twitter.finagle.http.Request;
+import com.twitter.finagle.http.Response;
+import com.twitter.finagle.http.Status;
+import com.twitter.io.Buf;
+import com.twitter.io.Reader;
+import com.twitter.util.Future;
+
+public class GreetingService extends Service<Request, Response> {
```
rpvilao, line 11: could we convert the whole example to unit tests?

libraries-rpc/pom.xml
```diff
@@ -0,0 +1,30 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>libraries-rpc</artifactId>
+    <name>libraries-rpc</name>
+    <packaging>jar</packaging>
+
+    <parent>
+        <groupId>com.baeldung</groupId>
+        <artifactId>parent-modules</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
```
rpvilao, line 12: let's create properties for the versions

libraries-rpc/src/main/java/com/baeldung/rpc/finagle/LogFilter.java
```diff
@@ -0,0 +1,17 @@
...
+import com.twitter.finagle.SimpleFilter;
+import com.twitter.finagle.http.Request;
+import com.twitter.finagle.http.Response;
+import com.twitter.util.Future;
+
+public class LogFilter extends SimpleFilter<Request, Response> {
+    @Override
+    public Future<Response> apply(Request request, Service<Request, Response> service) {
+        System.out.println("Request host:" + request.host().getOrElse(() -> ""));
+        System.out.println("Request params:");
```
rpvilao, line 13: if the stdouts are irrelevant let's remove them

mdabrowski-eu: The whole point of this class is to log things so I wouldn't say they are irrelevant but maybe there is some standard logger library to use instead. If so, let me know, I will replace stdouts.

rpvilao: since you are importing the logback already, then I would say we use the logging facility

mdabrowski-eu: Ok, I replaced stdouts with a logger.

PR #9186, opened 2020-04-26
Replaced for loop with enhanced for

John-Richardson:


lor6:
Thanks @John-Richardson

PR #9221, opened 2020-05-03
BAEL-3995 - Spring Security with Okta

SmartyAnsh:


spring-security-modules/pom.xml
```diff
@@ -30,6 +30,7 @@
         <module>spring-security-mvc-login</module>
         <module>spring-security-mvc-persisted-remember-me</module>
         <module>spring-security-mvc-socket</module>
+        <module>spring-security-okta</module>
         <module>spring-security-oidc</module>
```
KevinGilmore, line 34: swap these two lines to keep the order alphabetical

SmartyAnsh: fixed

PR #9227, opened 2020-05-03
BAEL-3829: example of dynamic bean autowiring in Spring

maciejglowka:


spring-core-4/src/main/java/com/baeldung/dynamic/autowire/InterfaceDynamicAutowireService.java
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+@Service
+public class InterfaceDynamicAutowireService {
```
rpvilao, line 12: why interface in the name?

spring-core-4/src/main/java/com/baeldung/dynamic/autowire/UkRegionService.java
```diff
@@ -0,0 +1,16 @@
+package com.baeldung.dynamic.autowire;
+
+import org.springframework.stereotype.Service;
+
+@Service("uk")
```
rpvilao, line 5: since the names are global I would say this needs to be more specific, just uk seems a bad practice

maciejglowka: Ok, will change

spring-core-4/src/main/java/com/baeldung/dynamic/autowire/InterfaceDynamicAutowireService.java
maciejglowka: Wanted to show in the name, that this solution is based on interface autowiring, will change for a better name because this one can be missleading

PR #9238, opened 2020-05-04
Bael-3857 introduction to lock free data structures

Maiklins:


core-java-modules/core-java-concurrency-advanced-3/src/main/java/com/baeldung/lockfree/NonBlockingQueue.java
```diff
@@ -0,0 +1,81 @@
...
+        size.decrementAndGet();
+        return currentHead.getValue();
+    }
+
+    public int size() {
+        return this.size.get();
+    }
+
+    private class Node<T> {
+        private volatile T value;
```
pivovarit, line 60: I believe that should be final since we should not allow mutation inside a `Node` class other than for the sake of linking with next/previous `Node`.

```suggestion
        private final T value;
```

PR #9272, opened 2020-05-11
BAEL-2495 fixing InvalidDefinitionException

patkorek:


jackson-modules/jackson-conversions/src/main/java/com/baeldung/jackson/date/EventWithLocalDate.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.fasterxml.jackson.datatype.jsr310.deser.LocalDateDeserializer;
+import com.fasterxml.jackson.datatype.jsr310.ser.LocalDateSerializer;
+
+import java.time.LocalDate;
+
+
```
ashleyfrieze, line 11: Minor: extra blankspace here isn't needed - just one line's separation between `import` and `class`

jackson-modules/jackson-conversions/src/main/java/com/baeldung/jackson/date/EventWithLocalDate.java
```diff
@@ -0,0 +1,36 @@
...
+public class EventWithLocalDate {
+    public String name;
+
+    @JsonDeserialize(using = LocalDateDeserializer.class)
+    @JsonSerialize(using = LocalDateSerializer.class)
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd-MM-yyyy")
+    public LocalDate eventDate;
+
+    public EventWithLocalDate() {
+        super();
```
ashleyfrieze, line 21: this shouldn't be present - `super()` the base class is `Object` which doesn't have a special constructor to call.

jackson-modules/jackson-conversions/src/test/java/com/baeldung/jackson/date/JacksonDateUnitTest.java
```diff
@@ -161,6 +157,29 @@ public void whenSerializingJava8Date_thenCorrect() throws JsonProcessingExceptio
...
+    @Test
+    public void whenSerializingJava8DateAndReadingValue_thenCorrect() throws IOException {
+        final String stringDate = "\"2014-12-20\"";
+
+        final ObjectMapper mapper = new ObjectMapper();
+        mapper.registerModule(new JavaTimeModule());
+        mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
+
+        final String result2 = mapper.readValue(stringDate, LocalDate.class)
+            .toString();
```
ashleyfrieze, line 169: This is an odd way to test serialization: load it into an object and then do `toString` on that object. Can we not assert the fields of the LocalDate object? At a glance, this just looks like string processing, even though there's a date object in the middle.

PR #9299, opened 2020-05-16
BAEL-3745: Implement PropertySourceFactory to load YAML files

kwoyke:


kwoyke:
This needs to be moved to a new module.

PR #9300, opened 2020-05-16
BAEL-3996_Spring_Security_with_Auth0

SmartyAnsh:


spring-security-modules/spring-security-auth0/src/main/java/com/baeldung/auth0/controller/AuthController.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Autowired
+    private AuthenticationController authenticationController;
+
+    @Autowired
+    private AuthConfig config;
+
+    private static final String AUTH0_TOKEN_URL = "https://dev-example.auth0.com/oauth/token";
+
+    @RequestMapping(value = "/login", method = RequestMethod.GET)
```
KevinGilmore, line 39: Where possible, use the shortcut annotations like _GetMapping_ and _PostMapping_ in this controller as well as the others.

SmartyAnsh: It's fixed. Thanks

PR #9301, opened 2020-05-16
[JAVA-1657] Junit and Maven Surefire Plugin upgrade in core-java-modules

dupirefr:


core-java-modules/pom.xml
```diff
@@ -153,11 +153,18 @@
...
+                <version>${junit-jupiter.version}</version>
+                <scope>test</scope>
+            </dependency>
         </dependencies>
     </dependencyManagement>
 
     <properties>
-        <junit.version>4.13</junit.version>
         <maven-surefire-plugin.version>2.22.2</maven-surefire-plugin.version>
+        <junit.version>4.13</junit.version>
```
dkapil, line 167: <junit.version> property should not be needed now right? can we please remove it to verify that we are no accidentally using it anywhere down the line

PR #9311, opened 2020-05-17
used H2 database to validate the user

amit2103:


spring-5-security/src/main/resources/persistence.properties
```diff
@@ -0,0 +1,11 @@
+# jdbc.X
+jdbc.driverClassName=org.h2.Driver
+jdbc.url=jdbc:h2:mem:db;DB_CLOSE_DELAY=-1;INIT=CREATE SCHEMA IF NOT EXISTS USERS
+jdbc.user=sa
+jdbc.pass=sa
```
dkapil, line 5: this file should not be needed, spring boot should be able to cover all of this automatically, even if any property is still needed it should be added to application.properties with a spring boot equivalent of these properties for e.g. spring.datasource.username instead of jdbc.user, for more reference - https://docs.spring.io/autorepo/docs/spring-boot/1.3.0.M2/reference/html/boot-features-sql.html

spring-5-security/src/main/resources/application.properties
```diff
@@ -2,4 +2,10 @@ server.port=8081
 
 logging.level.root=INFO
 
-logging.level.com.baeldung.dsl.ClientErrorLoggingFilter=DEBUG
\ No newline at end of file
+logging.level.com.baeldung.dsl.ClientErrorLoggingFilter=DEBUG
+
+spring.datasource.driver-class-name = org.hsqldb.jdbc.JDBCDriver
```
dkapil, line 7: if we have added these we should not be needing persistence.properties, also please check the properties in my opinion for embedded h2 server we should not be needing these properties at all

PR #9366, opened 2020-05-26
BAEL3889 - Kafka Mock Producer

dev-chirag:


libraries-data-2/src/main/java/com/baeldung/kafka/producer/EvenOddPartitioner.java
```diff
@@ -0,0 +1,16 @@
...
+import org.apache.kafka.clients.producer.internals.DefaultPartitioner;
+import org.apache.kafka.common.Cluster;
+
+public class EvenOddPartitioner extends DefaultPartitioner {
+
+    @Override
+    public int partition(String topic, Object key, byte[] keyBytes, Object value, byte[] valueBytes, Cluster cluster) {
+
+        if (((String)key).length() % 2 == 0)
+            return 0;
```
KevinGilmore, line 12: use of braces is generally considered a best practice in this situation, even though not required by the language

PR #9370, opened 2020-05-26
Use ServiceLoader class to obtain service instance

mangstadt:
In the "main-app" module, the `HelloInterface` service instance, `HelloModules`, was just being treated like an ordinary class that was exported from the "hello.modules" module. The code was not treating the class as a service class. This is not a good example of how services are used.

This pull request makes use of the `ServiceLoader` class, which provides access to the `HelloInterface` instance as defined in the "hello.modules" module. This serves as a better example for how services are used.


dkapil:
Hi @mangstadt 

The intent of the sample code was to show how to make use of JDK 9's provides with and uses directives. The example that is taken to demonstrate this was of a service class.

With that said, we do agree that the way you suggested here is more appropriate for using services.
Thank you for your contribution.

PR #9383, opened 2020-05-28
BAEL-3341 Rename test dependent on database

andrebrowne:


andrebrowne:
@kwoyke @rpvilao I think this should fix the build. Let me know

PR #9389, opened 2020-05-28
BAEL-4039: Introducing Constant Folding & Dead Code Elimination

alimate:


jmh/src/main/java/com/baeldung/BenchMark.java
```diff
@@ -45,4 +51,42 @@ public void init() {
         // Do nothing
     }
 
+    @Benchmark
+    @OutputTimeUnit(TimeUnit.NANOSECONDS)
+    @BenchmarkMode(Mode.AverageTime)
+    public void doNothing() {}
```
eric-martin, line 57: } on the next line

PR #9431, opened 2020-06-04
BAEL-4006 | Liskov Substitution Principle in Java

rdevarakonda:


patterns/solid/src/main/java/com/baeldung/l/advanced/Car.java
```diff
@@ -0,0 +1,14 @@
+package com.baeldung.l.advanced;
+
+public abstract class Car {
+    // invariant: speed < limit;
+    protected int speed, limit;
```
ashleyfrieze, line 5: Should we have the `limit` declare in the more conventional method `protected int limit` rather than with this quirky form?

patterns/solid/src/main/java/com/baeldung/l/advanced/Car.java
```diff
@@ -0,0 +1,14 @@
...
+
+    protected abstract void turnOnEngine();
+
+    // postcondition: speed < limit
+    protected abstract void accelerate();
+
+    // postcondition: speed must reduce
+    protected abstract void brake();
```
ashleyfrieze, line 13: Looks like automagic throttling algorithm ;)

patterns/solid/src/main/java/com/baeldung/l/advanced/ElectricCar.java
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  @Override
+  protected void accelerate() {
+    //this acceleration is crazy!
+  }
+  
+  @Override
+  protected void brake() {
+      // Apply ElectricCar brake
```
ashleyfrieze, line 17: Comment format and indentation is not consistent.

patterns/solid/src/main/java/com/baeldung/l/advanced/Account.java
```diff
@@ -0,0 +1,10 @@
+package com.baeldung.l.advanced;
```
ashleyfrieze, line 1: Package name of `l` is probably not great. How about `liskov`?

patterns/solid/src/main/java/com/baeldung/l/advanced/ReadOnlyFileSystem.java
```diff
@@ -0,0 +1,16 @@
...
+import java.io.IOException;
+
+public class ReadOnlyFileSystem implements FileSystem {
+    public File[] listFiles(String path) {
+        // code to list files
+        return new File[0];
+    }
+
+    public void deleteFile(String path) throws IOException {
+        // Do nothing.
```
ashleyfrieze, line 13: Do nothing? or throw exception as per other examples?

patterns/solid/src/main/java/com/baeldung/l/advanced/Car.java
rdevarakonda: Fixed

patterns/solid/src/main/java/com/baeldung/l/advanced/ReadOnlyFileSystem.java
rdevarakonda: As you picked up during the article review, this was deliberate to illustrate a code smell. 
"Doing Nothing" with this comment 😉

patterns/solid/src/main/java/com/baeldung/l/advanced/ElectricCar.java
rdevarakonda: Fixed

patterns/solid/src/main/java/com/baeldung/l/advanced/Account.java
rdevarakonda: OK - this is a slightly tricky one.
These packages names were introduced for the examples used in https://www.baeldung.com/solid-principles article e.g. _com.baeldung.s, com.baeldung.o, com.baeldung.l_ and so on. 
I put my examples in a new package under_com.baeldung.l_ called _com.baeldung.l.advanced_. 

As much as I did not like the original package names, I did not want to include package name refactoring at least in the first pass of this PR. Now that you have reviewed the code for this article, I can do the refactoring as a separate commit. Please let me know what you think.

PR #9438, opened 2020-06-04
[BAEL-3841] declarative caching testing

adrian-maghear:


spring-caching/src/main/resources/schema.sql
```diff
@@ -1,3 +1,7 @@
+DROP TABLE ORDERDETAIL IF EXISTS;
```
adrian-maghear, line 1: had to add these because spring will try to load the ***.sql** initialization files for each integration test. This will result in `table already exists` errors and failing tests when running tests with maven `integration-heavy` profile. Actually this currently fails on master as it is.

The two sql files are: 
- not needed for testing, since the only test making use of these domain entities is actually mocking the repositories
- needed in the CacheApplication **only if one uncomments** the `@ComponentScan` annotation

Another solution would be to set the `spring.datasource.initialization-mode=never` in the test application.properties but that's not really a good option. 
Or separate all these caching articles into their own separate maven modules and this way they would not interfere with each other.

PR #9459, opened 2020-06-08
cgrappa@gmail.com / Hexagonal Architecture in Java sample article

charlygrappa:
The code for the sample article of an Hexagonal Architecture in Java.
Since it is for a sample article the only operation implemented is the one in the article. It is fully functional using an Embedded MongoDb instance and has unit tests for both the business logic, and a sample test for the controller.


charlygrappa:
Added the missing interface and renamed the implementing class.

PR #9503, opened 2020-06-15
BAEL-4172 - How to Turn Off Swagger-ui in Production

stancalau:
How do you turn off swagger-ui in production


stancalau:
Replaces with https://github.com/eugenp/tutorials/pull/9858

PR #9555, opened 2020-06-22
BAEL-4155 - First draft

developerDiv:


rpvilao:
could we please move all the code to https://github.com/eugenp/tutorials/tree/master/core-java-modules/core-java-networking-2 since that one already has 10 articles?

PR #9557, opened 2020-06-22
BAEL-1488: JSON-LD

mdhtr:
Serializing and deserializing with Jackson


json-2/src/main/java/com/baeldung/jsonld/deserialization/jsonldjava/jackson/JacksonDeserializationExample.java
```diff
@@ -0,0 +1,40 @@
+package com.baeldung.jsonld.deserialization.jsonldjava.jackson;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
```
JonCook, line 5: Let's remove this unused import

json-2/src/main/java/com/baeldung/jsonld/deserialization/jsonldjava/jackson/JacksonDeserializationExample.java
```diff
@@ -0,0 +1,40 @@
...
+           "    }" + 
+           "  }," + 
+           "  \"@type\": \"Person\"," + 
+           "  \"@id\": \"http://example.com/person/1234\"," + 
+           "  \"name\": \"Example Name\"," + 
+           "  \"knows\": \"http://example.com/person/2345\"" + 
+           "}";
+
+        Object jsonObject = JsonUtils.fromString(exampleJsonld);
+        Object compact = JsonLdProcessor.compact(jsonObject, new HashMap(), new JsonLdOptions());
```
JonCook, line 28: Let's fix the warning about raw types if possible.

json-2/src/main/java/com/baeldung/jsonld/deserialization/jsonldjava/jackson/JacksonDeserializationExample.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.jsonldjava.core.JsonLdOptions;
+import com.github.jsonldjava.core.JsonLdProcessor;
+import com.github.jsonldjava.utils.JsonUtils;
+
+public class JacksonDeserializationExample {
+    public static void main(String[] args) throws IOException {
```
JonCook, line 13: I'm not a big fan of these *Example type classes with the main method. Let's move this into a unit test using the test method names (givenX_whenY_thenZ) as mentioned in the author guidelines.

In general, we want as much coverage as possible on our new code. Putting it in a unit test is also easier for readers to run the code and also we validate that it is working as expected.

json-2/src/main/java/com/baeldung/jsonld/deserialization/jsonldjava/jackson/JacksonDeserializationExample.java
```diff
@@ -0,0 +1,40 @@
...
+        Object jsonObject = JsonUtils.fromString(exampleJsonld);
+        Object compact = JsonLdProcessor.compact(jsonObject, new HashMap(), new JsonLdOptions());
+        String compactContent = JsonUtils.toString(compact);
+
+        System.out.println(JsonUtils.toPrettyString(compact));
+
+        ObjectMapper objectMapper = new ObjectMapper();
+        JacksonExamplePerson person = objectMapper.readValue(compactContent, JacksonExamplePerson.class);
+
+        System.out.println(person.id);
```
JonCook, line 36: These system.out will likely become assert statements. But if you want to print, let's use a proper logger.

json-2/src/main/java/com/baeldung/jsonld/deserialization/jsonldjava/jackson/JacksonExamplePerson.java
```diff
@@ -0,0 +1,19 @@
+package com.baeldung.jsonld.deserialization.jsonldjava.jackson;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class JacksonExamplePerson {
```
JonCook, line 7: As these POJO classes are already in a separate package, can we not simply call the class Person?

json-2/src/main/java/com/baeldung/jsonld/serialization/hydrajsonld/HydraJsonldExamplePerson.java
```diff
@@ -0,0 +1,15 @@
+package com.baeldung.jsonld.serialization.hydrajsonld;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import de.escalon.hypermedia.hydra.mapping.Expose;
+import de.escalon.hypermedia.hydra.mapping.Vocab;
+
+@Vocab("http://example.com/vocab/")
+@Expose("person")
+public class HydraJsonldExamplePerson {
```
JonCook, line 10: As these POJO classes are already in a separate package, can we not simply call the class Person?

json-2/src/main/java/com/baeldung/jsonld/serialization/hydrajsonld/HydraJsonldSerializationExample.java
```diff
@@ -0,0 +1,52 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationConfig;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.databind.ser.BeanSerializerModifier;
+import com.fasterxml.jackson.databind.ser.std.BeanSerializerBase;
+
+import de.escalon.hypermedia.hydra.serialize.JacksonHydraSerializer;
+
+public class HydraJsonldSerializationExample {
+    public static void main(String[] args) throws IOException {
```
JonCook, line 17: The same comment applies about moving to a unit test.

json-2/src/main/java/com/baeldung/jsonld/serialization/hydrajsonld/HydraJsonldSerializationExample.java
```diff
@@ -0,0 +1,52 @@
...
+        HydraJsonldExamplePerson person = new HydraJsonldExamplePerson();
+        person.id = "http://example.com/person/1234";
+        person.name = "Example Name";
+
+        System.out.println(mapper.writerWithDefaultPrettyPrinter()
+            .writeValueAsString(person));
+    }
+
+    public static SimpleModule getJacksonHydraSerializerModule() {
+        return new SimpleModule() {
```
JonCook, line 31: There is a warning here about serial version Ids

json-2/src/main/java/com/baeldung/jsonld/serialization/jacksonjsonld/JacksonJsonldExamplePerson.java
```diff
@@ -0,0 +1,17 @@
...
+import ioinformarics.oss.jackson.module.jsonld.annotation.JsonldId;
+import ioinformarics.oss.jackson.module.jsonld.annotation.JsonldLink;
+import ioinformarics.oss.jackson.module.jsonld.annotation.JsonldNamespace;
+import ioinformarics.oss.jackson.module.jsonld.annotation.JsonldProperty;
+import ioinformarics.oss.jackson.module.jsonld.annotation.JsonldType;
+
+@JsonldNamespace(name = "s", uri = "http://schema.org/")
+@JsonldLink(rel = "s:knows", name = "knows", href = "http://example.com/person/2345")
+@JsonldType("s:Person")
+class JacksonJsonldExamplePerson {
```
JonCook, line 12: As these POJO classes are already in a separate package, can we not simply call the class Person?

json-2/src/main/java/com/baeldung/jsonld/serialization/jacksonjsonld/JacksonJsonLdSerializationExample.java
```diff
@@ -0,0 +1,25 @@
+package com.baeldung.jsonld.serialization.jacksonjsonld;
+
+import java.io.IOException;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import ioinformarics.oss.jackson.module.jsonld.HydraCollection;
```
JonCook, line 7: Let's clean up these unused imports.

json-2/src/main/java/com/baeldung/jsonld/serialization/jacksonjsonld/JacksonJsonLdSerializationExample.java
```diff
@@ -0,0 +1,25 @@
...
+import ioinformarics.oss.jackson.module.jsonld.JsonldGraph;
+import ioinformarics.oss.jackson.module.jsonld.JsonldModule;
+import ioinformarics.oss.jackson.module.jsonld.JsonldResource;
+
+public class JacksonJsonLdSerializationExample {
+    public static void main(String[] args) throws IOException {
```
JonCook, line 13: Same comment regarding the unit tests.

json-2/src/main/java/com/baeldung/jsonld/deserialization/jsonldjava/jackson/JacksonDeserializationExample.java
mdhtr: @JonCook I've converted everything into unit tests, but I'm not confident that I've found good names for these tests. Using assertions is definitely better than console prints, so I am happy in that regard, however, the Eclipse formatter doesn't allow multiline strings, so the inline JSON objects look terrible.

json-2/src/main/java/com/baeldung/jsonld/deserialization/jsonldjava/jackson/JacksonDeserializationExample.java
mdhtr: I must have some settings not properly set because I see no warnings. Could you point me in the right direction?

json-2/src/main/java/com/baeldung/jsonld/serialization/hydrajsonld/HydraJsonldSerializationExample.java
mdhtr: I must have some settings not properly set because I see no warnings. Could you point me in the right direction?

json-2/src/main/java/com/baeldung/jsonld/deserialization/jsonldjava/jackson/JacksonDeserializationExample.java
JonCook: The warning here is - HashMap is a raw type. References to generic type HashMap<K,V> should be parameterized. It can be fixed like - new HashMap<>()

json-2/src/main/java/com/baeldung/jsonld/serialization/hydrajsonld/HydraJsonldSerializationExample.java
JonCook: The warning here is - The serializable class does not declare a static final serialVersionUID field of type long. But I think we can leave this one. Don't worry.

json-2/src/main/java/com/baeldung/jsonld/deserialization/jsonldjava/jackson/JacksonDeserializationExample.java
JonCook: I think the test approach is good! Could we shorten the test names a little? I think you are on the right track. e.g
`givenAJsonLdObject_whenCompactIsUsedWithEmptyContext_thenItCanBeDeserializedIntoAJacksonAnnotatedJavaObject()`

Could become for example (just an idea):
`givenAJsonLdString_whenCompactIsUsedWithEmptyContext_thenDeserializedCorrectly()`

Let's also please move the Person class into the main src tree.

Finally, I agree the inline JSON does not look good. I'm afraid we will have to fix that manually without using the formatter.

mdhtr: @JonCook is `givenAJsonLdObject_whenCompactIsUsedWithEmptyContext_thenItCanBeDeserializedWithJackson()` okay? Are the two other test names okay like they are?

PR #9628, opened 2020-07-02
[BAEL-4074] Accessing Maven properties in Java

aurasphere:


maven-all/maven-2/pom.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
KevinGilmore, line 3: use 4-space indent instead of tabs

maven-modules/pom.xml
```diff
@@ -24,6 +24,7 @@
         <module>maven-profiles</module>
         <module>versions-maven-plugin</module>
         <module>version-collision</module>
+        <module>maven-properties</module>
```
KevinGilmore, line 27: We like to keep these alphabetical, so please move this line up a bit so that it's after maven-profiles but before versions-maven-plugin

PR #9639, opened 2020-07-03
BAEL 2881

charlygrappa:
Remove Trove&Colt and add HPPC


sergioKon:
Don't send it anymore

PR #9660, opened 2020-07-07
[BAEL-4213] Why are local variables thread safe

martinvw:
Added two small example for article about thread-safety for local variables.


martinvw:
@JonCook FYI

PR #9756, opened 2020-07-23
BAEL-4289

MajewskiKrzysztof:


core-java-modules/core-java-os/src/test/java/com/baeldung/screenshot/ScreenshotTest.java
```diff
@@ -0,0 +1,54 @@
...
+public class ScreenshotTest {
+
+    @Test
+    public void takeScreenshotOfMainScreen() throws Exception {
+        Rectangle screenRect = new Rectangle(Toolkit.getDefaultToolkit().getScreenSize());
+        BufferedImage capture = new Robot().createScreenCapture(screenRect);
+        File imageFile = new File("single-screen.bmp");
+        ImageIO.write(capture, "bmp", imageFile);
+        assertTrue(imageFile.exists());
+        imageFile.delete();
```
rpvilao, line 23: my suggestion here would be to save the file in a temp file and do not take care of the cleaning. If for some reason the deletion fails, the test is still valid, but it will fail to pass. alternatively you can do the deletion using an @After method, this keeping the test and assertion clean. In my opinion, I would go with the tmp version and do not take care of the cleaning :)

rpvilao: ah, this applies to all the tests of course :)

MajewskiKrzysztof: done :)

PR #9768, opened 2020-07-25
Replace StringBuffer with StringBuilder and add initial capacity

tweimer:
Replace StringBuffer with StringBuilder and add initial  capacity.
StringBuilder is faster because it is not synchronized.


lor6:
Thanks @tweimer Merged.

PR #9772, opened 2020-07-26
rutujavjoshi@gmail.com

rutujavjoshi:
Added code related to evaluation article ( A quick and practical example of Hexagonal Architecture in Java)
Added new code - src and tests : initial commit. 
Code added to ddd module ;


rutujavjoshi:
I had done two separate commits -
A> Code only (CI OK)
B> Tests (**_CI failure_**)


I replaced my local pom with the pom in the module and noticed that maven install failed for the tests.
Error being
'org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'org.springframework.boot.autoconfigure.data.**cassandra.CassandraDataAutoConfiguration': Unsatisfied dependency'**

When I removed 
 <dependency>
            <artifactId>spring-boot-starter-data-cassandra</artifactId>
        </dependency>
from the pom it is working. But I am not sure about the other code in the ddd module hence do not wish to update it.  Please  advice.

image

rutujavjoshi:
Updating code to follow unit test class naming conventions

rutujavjoshi:
Updating code to follow unit test class naming conventions - folder structure update

rutujavjoshi:
Hi,

Dropped the tests for now as they were still erroring out with master pom.
As an alternative can create separate module 'design' with my POM but skipping to  avoid redundancy, follow guidelines to reuse existing modules if possible.

PR #9786, opened 2020-07-28
Source code for hexagonal architecture sample application

gssachdeva:
Source code for hexagonal architecture sample application


gssachdeva:
Submitting another pull request with optimized code

PR #9826, opened 2020-08-05
JSON Parameters with Spring MVC - BAEL-4327

alex07pep:
JSON Parameters with Spring MVC - BAEL-4327 alexandruflorin.peptan@gmail.com

### DIFF
--- a/spring-mvc-basics-4/src/main/java/com/baeldung/jsonparam/Application.java
+++ b/spring-mvc-basics-4/src/main/java/com/baeldung/jsonparam/Application.java
@@ -1,0 +1,13 @@
+package com.baeldung.jsonparam;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+@SpringBootApplication
+public class Application{
+
+    public static void main(String[] args) {
+        SpringApplication.run(Application.class, args);
+    }
+
+}
--- a/spring-mvc-basics-4/src/main/java/com/baeldung/jsonparam/JsonParamController.java
+++ b/spring-mvc-basics-4/src/main/java/com/baeldung/jsonparam/JsonParamController.java
@@ -1,0 +1,55 @@
+package com.baeldung.jsonparam;
+
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.*;
+
+import java.io.IOException;
+
+@Controller
+@RestController
+@RequestMapping("/api")
+public class JsonParamController {
+
+    @GetMapping(value = "/message")
+    public String testQueryParamApi(@RequestParam("jsonParam") String jsonParam) {
+        // jsonParam={"level":3,"type":"foo","color":"green"}
+        ParamObjectDTO paramObjectDTO;
+        ObjectMapper objectMapper = new ObjectMapper();
+        try {
+            paramObjectDTO = objectMapper.readValue(jsonParam, ParamObjectDTO.class);
+            System.out.println(paramObjectDTO);
+            // use paramObjectDTO where you have {"level":3,"type":"foo","color":"green"} JSON data as Object
+        } catch (JsonParseException e) {
+            e.printStackTrace();
+        } catch (JsonMappingException e) {
+            e.printStackTrace();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        return jsonParam;
+    }
+
+
+    @PostMapping(value = "/message")
+    public String testBodyParamApi(@RequestBody String jsonParam) {
+        // jsonParam={"level":3,"type":"foo","color":"green"}
+        ParamObjectDTO paramObjectDTO;
+        ObjectMapper objectMapper = new ObjectMapper();
+        try {
+            paramObjectDTO = objectMapper.readValue(jsonParam, ParamObjectDTO.class);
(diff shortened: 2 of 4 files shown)


spring-mvc-basics-4/src/main/java/com/baeldung/jsonparam/Application.java
```diff
@@ -0,0 +1,13 @@
+package com.baeldung.jsonparam;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+@SpringBootApplication
+public class Application{
```
JonCook, line 7: Could we think of a better name? Application is very generic.

spring-mvc-basics-4/src/main/java/com/baeldung/jsonparam/JsonParamController.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.io.IOException;
+
+@Controller
+@RestController
+@RequestMapping("/api")
+public class JsonParamController {
+
+    @GetMapping(value = "/message")
+    public String testQueryParamApi(@RequestParam("jsonParam") String jsonParam) {
```
JonCook, line 17: Let's think of a better name for this method. We wouldn't normally start a method in our Controllers with testxxx. Likewise, jsonParam is very vague. What is this param?

spring-mvc-basics-4/src/main/java/com/baeldung/jsonparam/JsonParamController.java
```diff
@@ -0,0 +1,55 @@
...
+        }
+        return jsonParam;
+    }
+
+
+    @PostMapping(value = "/message")
+    public String testBodyParamApi(@RequestBody String jsonParam) {
+        // jsonParam={"level":3,"type":"foo","color":"green"}
+        ParamObjectDTO paramObjectDTO;
+        ObjectMapper objectMapper = new ObjectMapper();
```
JonCook, line 40: If we're using Spring can't we autowire this to make it testable? In general it is a good idea to avoid new-ing classes directly in methods.

spring-mvc-basics-4/src/main/java/com/baeldung/jsonparam/JsonParamController.java
```diff
@@ -0,0 +1,55 @@
...
+public class JsonParamController {
+
+    @GetMapping(value = "/message")
+    public String testQueryParamApi(@RequestParam("jsonParam") String jsonParam) {
+        // jsonParam={"level":3,"type":"foo","color":"green"}
+        ParamObjectDTO paramObjectDTO;
+        ObjectMapper objectMapper = new ObjectMapper();
+        try {
+            paramObjectDTO = objectMapper.readValue(jsonParam, ParamObjectDTO.class);
+            System.out.println(paramObjectDTO);
```
JonCook, line 23: Let's use a proper Logger

spring-mvc-basics-4/src/main/java/com/baeldung/jsonparam/JsonParamController.java
```diff
@@ -0,0 +1,55 @@
...
+            // use paramObjectDTO where you have {"level":3,"type":"foo","color":"green"} JSON data as Object
+        } catch (JsonParseException e) {
+            e.printStackTrace();
```
JonCook, line 26: If we are going to catch the exceptions, let's use a Logger and also consider catching them all together on one line if the action is the same for all of them.

spring-mvc-basics-4/src/main/java/com/baeldung/jsonparam/JsonParamController.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @GetMapping(value = "/message")
+    public String testQueryParamApi(@RequestParam("jsonParam") String jsonParam) {
+        // jsonParam={"level":3,"type":"foo","color":"green"}
+        ParamObjectDTO paramObjectDTO;
+        ObjectMapper objectMapper = new ObjectMapper();
+        try {
+            paramObjectDTO = objectMapper.readValue(jsonParam, ParamObjectDTO.class);
+            System.out.println(paramObjectDTO);
+            // use paramObjectDTO where you have {"level":3,"type":"foo","color":"green"} JSON data as Object
```
JonCook, line 24: This comment seems to be misplaced? Let's fix that.

spring-mvc-basics-4/src/main/java/com/baeldung/jsonparam/JsonParamController.java
```diff
@@ -0,0 +1,55 @@
...
+            e.printStackTrace();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        return jsonParam;
+    }
+
+
+    @PostMapping(value = "/message")
+    public String testBodyParamApi(@RequestBody String jsonParam) {
```
JonCook, line 37: The same comment applies regarding the method name and param name.

spring-mvc-basics-4/src/main/java/com/baeldung/jsonparam/ParamObjectDTO.java
```diff
@@ -0,0 +1,43 @@
+package com.baeldung.jsonparam;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class ParamObjectDTO {
```
JonCook, line 6: Let's give this domain object a better name, I'm not sure what it is supposed to represent? Your controller has /message in the URL but this class doesn't seem to represent a message. It is confusing, at least to me.

PR #9849, opened 2020-08-11
BAEL-4336 - @BeforeAll and @AfterAll Methods as a non-static 

meysam:


bfontana:
Discarded. Replaced by a new one.

PR #9854, opened 2020-08-12
AssertEquals 2 Lists ignore order

vishal1023:
Add unit tests to check equality of two lists without considering the order of elements in it


vishal1023:
@JonCook, I have implemented the comments you provide.

Issue #8361, opened 2019-12-13
Spring Cognito Tutorial Not functioning


Warfront1:
https://www.baeldung.com/spring-security-oauth-cognito

When following the guide, and cloning the latest version of the repo.
I am getting the following experience when attempting to log into my test cognito user account.

![image](https://user-images.githubusercontent.com/1920038/70772220-e1682500-1d41-11ea-9eed-efb8d7bd968e.png)

I have tested the credentials with the AWS hosted UI, and they work perfectly. 

During the tutorial it states

> Now we can start our application, go to http://localhost:8080, and click the login link.

However once starting this is actually what is displayed (take note of there not being a login link anywhere):
![image](https://user-images.githubusercontent.com/1920038/70772645-2b9dd600-1d43-11ea-8dfc-f093cdd9cd80.png)

I feel as if I most likely have a configuration issue, but the tutorial provides little information on any configuration details. In addition the entire org.springframework.boot.autoconfigure.security.oauth2.client package seems to have little documentation about the OAuth2ClientProperties/Provider to assist in tracking down the issue.
To compound on the issue, there is no error thrown anywhere when attempting to sign on, other than the UI experience above. 

The login POST above gives you a 302 with a referral to http://localhost:8080/login?error

Here is my current configuration [note multiple variations of the issuerUri were tested] 
```
spring:
  security:
    oauth2:
      client:
        registration:
          cognito:
            client-id: xxxxxxxxxxxxxxxxxxxxxxxxxx
            client-secret: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
            scope: openid
            redirectUriTemplate: "http://localhost:8080/login/oauth2/code/cognito"
            clientName: s
        provider:
          cognito:
#            issuerUri: https://cognito-idp.us-east-1.amazonaws.com/xxxxxxxx
            issuerUri: https://testpoolxxxxx.auth.us-east-1.amazoncognito.com
            usernameAttribute: cognito:email

```

jzheaux:
@Warfront1 sorry to hear you are having trouble with this - would you be able to provide your sample application as a GitHub repo? I'll check it out and see if I can make something of it.

Warfront1:
> @Warfront1 sorry to hear you are having trouble with this - would you be able to provide your sample application as a GitHub repo? I'll check it out and see if I can make something of it.

[...]

PR #9871, opened 2020-08-14
JAVA-1639: Upgrade spring-5-security-cognito to use latest Spring Boot

sampada07:
version


sampada07:
Also fixes https://github.com/eugenp/tutorials/issues/8361

PR #9873, opened 2020-08-15
BAEL-4107: Spring MVC Async vs Spring WebFlux

maryarm:
Code examples for BAEL-4107 load test.


spring-5-webflux/.gitignore
```diff
@@ -1,2 +1,3 @@
 # Files #
-*.log
\ No newline at end of file
+*.log
+*.ipr
```
eric-martin, line 3: Let's remove this gitignore change

PR #9887, opened 2020-08-17
BAEL-4106

ricardomoreirab:
application.yml vs application.properties for Spring Boot

### DIFF
--- a/spring-boot-modules/spring-boot-properties-3/README.md
+++ b/spring-boot-modules/spring-boot-properties-3/README.md
@@ -1,0 +1,7 @@
+## Spring Boot Properties
+
+This module contains articles about Properties in Spring Boot.
+
+### Relevant Articles:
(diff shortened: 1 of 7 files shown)


spring-boot-modules/spring-boot-properties-3/README.md
```diff
@@ -0,0 +1,7 @@
+## Spring Boot Properties
+
+This module contains articles about Properties in Spring Boot.
+
+### Relevant Articles:
+- [Using application.yml vs application.properties in Spring Boot](https://www.baeldung.com/TODO)
```
ricardomoreirab, line 6: Have to change the link after the article is published

KevinGilmore, line 6: let's omit this line; your editor will add the link after it is published

PR #9905, opened 2020-08-22
KTLN-154 : Collection Aggregate Operations in Kotlin

anirban99:


anirban99:
I removed the generic prints. But I would prefer to keep these `println("total = $total, it = $it")`. The reason is that when we run the program/tests, it helps to understand the sequence of elements being called for the different fold and reduce functions. What do you think about it?

core-kotlin-modules/core-kotlin-collections/src/main/kotlin/com/baeldung/kotlin/collections/AggregateOperations.kt
```diff
@@ -0,0 +1,107 @@
+package com.baeldung.kotlin.collections
```
KevinGilmore, line 1: Let's make the package more specific, e.g. com.baeldung.kotlin.collections.aggregate

anirban99: @KevinGilmore Moved it to a new module.

rpvilao:
@anirban99 there was an issue with the refactor, it literally created a folder with dots instead of the structure

anirban99:
@rpvilao Where did it create the folder with dots? Can you please point it out in the code?

rpvilao:
image
Aggregate Operations for example.

anirban99:
@rpvilao Fixed the package structure

PR #9912, opened 2020-08-24
bjakupovic: Understanding java.lang.Thread.State: WAITING (parking)

belms:
Two example classes for parking a thread

### DIFF
--- a/core-java-modules/core-java-concurrency-basic-2/src/main/java/com/baeldung/concurrent/parking/ParkingThreadDemo.java
+++ b/core-java-modules/core-java-concurrency-basic-2/src/main/java/com/baeldung/concurrent/parking/ParkingThreadDemo.java
@@ -6,12 +6,10 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.concurrent.*;
 
-class ParkingThreadDemo {
+public class ParkingThreadDemo {
     private static final Logger LOG = LoggerFactory.getLogger(ParkingThreadDemo.class);
 
-    public void parkingThread throws
-
-    InterruptedException {
+    public void parkingThread() throws InterruptedException {
         final ExecutorService executorPool = Executors.newFixedThreadPool(5);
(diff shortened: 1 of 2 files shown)


core-java-modules/core-java-concurrency-basic-2/src/main/java/com/baeldung/concurrent/parking/WaitParkComparison.java
```diff
@@ -0,0 +1,34 @@
...
+                LOG.debug("Thread 1 is blocked");
+                LockSupport.park();
+                LOG.debug("Thread 1 resumed");
+            }
+        };
+        Thread thread1 = new Thread(task1);
+        thread1.start();
+        Runnable task2 = () -> {
+            LOG.debug("Thread 2 running ");
+            synchronized (lockObject) {
```
ashleyfrieze, line 26: As this is multi-threaded code, there's a chance that thread 2 will reach this point before thread 1 is parked,

core-java-modules/core-java-concurrency-basic-2/src/main/java/com/baeldung/concurrent/parking/ParkingThreadDemo.java
```diff
@@ -0,0 +1,47 @@
...
+
+        // consumer:
+        for (int i = 0; i < 20; i++) {
+            collection.add(executorPool.submit(new Callable<Long>() {
+                @Override
+                public Long call() throws Exception {
+                    while (true) {
+                        Long item = queue.take();
+                        if (item.intValue() == 0) {
+                            break;
```
ashleyfrieze, line 25: In this code, each of the 20 consumers will stop on hitting a zero. This is not incredibly easy to follow. This code is also quite monolithic and deeply nested. It is very abstract and seems to serve no real purpose.

Can we make a simpler more practical example?

ashleyfrieze (changes requested):

PR #9917, opened 2020-08-24
BAEL-4224: Hidden inputs with thymeleaf

michael-pratt:


spring-thymeleaf-3/src/main/java/com/baeldung/thymeleaf/blog/BlogDTO.java
```diff
@@ -0,0 +1,66 @@
+package com.baeldung.thymeleaf.blog;
```
rpvilao, line 1: just a thought, could we use lombok here?

michael-pratt: We could definitely use it here, but it wasn't already imported the module and I wanted to keep the example code as much aligned to the article as possible.

spring-thymeleaf-3/src/main/java/com/baeldung/thymeleaf/blog/BlogController.java
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+
+import java.util.Random;
+
+@Controller
+public class BlogController {
+
+    @GetMapping("/blog/new")
+    public String newBlogPost(Model model)
+    {
```
KevinGilmore, line 14: As per guidelines, the opening brace should be at the end of line 13, not by itself.

michael-pratt: Moved to same line

PR #9961, opened 2020-09-01
java-compareTo-guide

erkpsingh24:


core-java-modules/core-java-lang-3/src/main/java/com/baeldung/compareTo/CollectionJava.java
```diff
@@ -0,0 +1,85 @@
...
+    nameSortedList.add(new Person(8, "Josh"));
+    nameSortedList.add(new Person(2, "Kallis"));
+    nameSortedList.add(new Person(5, "Karen"));
+   
+    Collections.sort(list, new PersonNameComparator());
+    Assert.assertArrayEquals(list.toArray(), nameSortedList.toArray());
+    
+}
+    
+    public static void main(String args[]) {
```
KevinGilmore, line 80: After moving the code, you should write unit tests, not a main method.

KevinGilmore (changes requested):
PR does not build. Files are in the wrong packages, one doesn't have a package declared, and one should be in src/test/java and be converted to unit tests. See individual comments for more info.

PR #9980, opened 2020-09-06
Fix Gatling test used for https://www.baeldung.com/gatling-jmeter-grinder-comparison

slandelle:
The Gatling results in https://www.baeldung.com/gatling-jmeter-grinder-comparison are completely broken:

* dynamic parameter injection is incorrect, always the same value is injected, causing the application to actually crash because of the duplicates
* one of the response parameter capture (check) is incorrect and should be optional
* the test includes 1 second pauses between consecutive requests. The tests with the other tools don't have them. The result is that the Gatling throughput is artificially limited.

On my local machine, I get 4100 requests executed in ~3 seconds (instead of the 31s in your article), which give an average throughput of 1025 requests/second.

Generally speaking the performance figures in the article are very suspicious:

```
  | Successful Requests | Errors | Total Test Time (s) | Average Response Time (ms) | Peak Throughput
-- | -- | -- | -- | -- | --
Gatling | 4100 Requests | 100 (soft)* | 31 | 64.5 | 132 req/s
JMeter | 4135 Requests | 0 | 35 | 81 | 1080 req/s
The Grinder | 5000 Requests | 0 | 5.57 | 65.72 | 1284 req/s
```

* How come JMeter gets 4135 requests in 35s and get a throughput of 1080 req/s?!?!
* How do you measure "Peak Throughput"?
* Those tests are actually super short. With only 5,000 iterations, the JIT compiler hasn't kicked in yet. You should run longer tests to get meaningful results.

### DIFF
--- a/testing-modules/load-testing-comparison/src/main/resources/scripts/Gatling/GatlingScenario.scala
+++ b/testing-modules/load-testing-comparison/src/main/resources/scripts/Gatling/GatlingScenario.scala
@@ -20,7 +20,7 @@ class RewardsScenario extends Simulation {
 	.repeat(10){
 		exec(http("transactions_add")
 		  .post("/transactions/add/")
-		  .body(StringBody("""{ "customerRewardsId":null,"customerId":""""+ randCustId() + """","transactionDate":null }""")).asJson
+		  .body(StringBody(_ => s"""{ "customerRewardsId":null,"customerId":"${randCustId()}","transactionDate":null }""")).asJson
 		.check(jsonPath("$.id").saveAs("txnId"))
 		.check(jsonPath("$.transactionDate").saveAs("txtDate"))
 		.check(jsonPath("$.customerId").saveAs("custId")))
--- a/testing-modules/load-testing-comparison/src/main/resources/scripts/Gatling/GatlingScenario.scala
+++ b/testing-modules/load-testing-comparison/src/main/resources/scripts/Gatling/GatlingScenario.scala
@@ -28,7 +28,7 @@ class RewardsScenario extends Simulation {
 		
 		.exec(http("get_reward")
 		  .get("/rewards/find/${custId}")
-		  .check(jsonPath("$.id").saveAs("rwdId")))
+		  .check(jsonPath("$.id").optional.saveAs("rwdId")))
 		.pause(1)
 		
 		.doIf("${rwdId.isUndefined()}"){
--- a/testing-modules/load-testing-comparison/src/main/resources/scripts/Gatling/GatlingScenario.scala
+++ b/testing-modules/load-testing-comparison/src/main/resources/scripts/Gatling/GatlingScenario.scala
@@ -7,7 +7,7 @@ import scala.concurrent.duration._
 
 class RewardsScenario extends Simulation {
 
-  def randCustId() = Random.nextInt(99)
+  def randCustId() = Random.nextInt(100000)
   
   val httpProtocol = http.baseUrl("http://localhost:8080")
 					    .acceptHeader("text/html,application/json;q=0.9,*/*;q=0.8")
--- a/testing-modules/load-testing-comparison/src/main/resources/scripts/Gatling/GatlingScenario.scala
+++ b/testing-modules/load-testing-comparison/src/main/resources/scripts/Gatling/GatlingScenario.scala
@@ -24,12 +24,10 @@ class RewardsScenario extends Simulation {
 		.check(jsonPath("$.id").saveAs("txnId"))
 		.check(jsonPath("$.transactionDate").saveAs("txtDate"))
 		.check(jsonPath("$.customerId").saveAs("custId")))
-		.pause(1)
 		
 		.exec(http("get_reward")
 		  .get("/rewards/find/${custId}")
 		  .check(jsonPath("$.id").optional.saveAs("rwdId")))
-		.pause(1)
 		
 		.doIf("${rwdId.isUndefined()}"){
 			exec(http("rewards_add")
@@ -41,7 +39,6 @@ class RewardsScenario extends Simulation {
 		.exec(http("transactions_add")
 		  .post("/transactions/add/")
 		  .body(StringBody("""{ "customerRewardsId":"${rwdId}","customerId":"${custId}","transactionDate":"${txtDate}" }""")).asJson)
-		.pause(1)
 		
 		.exec(http("get_reward")
 		  .get("/transactions/findAll/${rwdId}"))
--- a/testing-modules/load-testing-comparison/src/main/resources/scripts/Gatling/GatlingScenario.scala
+++ b/testing-modules/load-testing-comparison/src/main/resources/scripts/Gatling/GatlingScenario.scala
@@ -7,7 +7,7 @@ import scala.concurrent.duration._
 
 class RewardsScenario extends Simulation {
 
-  def randCustId() = Random.nextInt(100000)
+  def randCustId() = java.util.concurrent.ThreadLocalRandom.current().nextInt(1, 10000)
   
   val httpProtocol = http.baseUrl("http://localhost:8080")
 					    .acceptHeader("text/html,application/json;q=0.9,*/*;q=0.8")


slandelle:
@eugenp PTAL

slandelle:
@lor6 Thanks for merging. What's your plan regarding the broken published results? Do you plan on updating the existing post, or publish an update? Any ETA?

lor6:
@slandelle  I've updated the results here, using 100 instead of 10 users: https://www.baeldung.com/gatling-jmeter-grinder-comparison 

One issue is that Gatling is running fewer requests than the other tools, which is because of the if here: https://github.com/eugenp/tutorials/blob/master/testing-modules/load-testing-comparison/src/main/resources/scripts/Gatling/GatlingScenario.scala#L33 It seems that after the first run this always evaluates to false, even if the request above doesn't return a result: https://github.com/eugenp/tutorials/blob/master/testing-modules/load-testing-comparison/src/main/resources/scripts/Gatling/GatlingScenario.scala#L31 Is there a way to set the variable back to undefined?

Let me know if you see any other issues.

Thanks.

slandelle:
@lor6 Sorry, I don't understand.

> It seems that after the first run this always evaluates to false, even if the request above doesn't return a result: https://github.com/eugenp/tutorials/blob/master/testing-modules/load-testing-comparison/src/main/resources/scripts/Gatling/GatlingScenario.scala#L31

The test tries to create a new customer based on a random id (between 1 and 10,000 like the JMeter test), and then adds a new reward if there's not already one.
Every test run will generate 1,000 customers.

Now, the application under test uses an in-memory h2 database. This means that if you run tests without restarting the application in between, you'll increase the probability of hitting a customerId that already exists and hence having an already existing award.

In short. You should restart the application before running a test with any of the 3 tests tools so you start with the same state and don't compare apples and oranges.

Did I get it right?

lor6:
@slandelle Yes, that's correct, but I've restarted the app before running the test with each tool so each starts with a clean db. The issue with the Gatling script is that after the first run, the rwdId variable never evaluates to undefined again even if the previous request that saves the id to the rwdId returns null.

slandelle:
@lor6 OK, sorry, now I get it. Please see https://github.com/eugenp/tutorials/pull/10146

Modern versions would make it easier but the post target old versions.
Could you please merge and let me know?

lor6:
Thanks @slandelle It seems ok now. 

Let me know if you see any other issues in the post: https://www.baeldung.com/gatling-jmeter-grinder-comparison

slandelle:
Thanks @lor6. However, yes, I see other issues:

**Number of requests:**

The 3 tests don't generate the same number of requests while the loop is configured with a fixed number of iterations. This means the 3 tests don't do the same thing.

This is most likely because the random generators are not configured the same way:
* [Grinder](https://github.com/eugenp/tutorials/blob/master/testing-modules/load-testing-comparison/src/main/resources/scripts/The%20Grinder/grinder.py#L28): `Random#nextInt()`
* [Gatling](https://github.com/eugenp/tutorials/blob/master/testing-modules/load-testing-comparison/src/main/resources/scripts/Gatling/GatlingScenario.scala#L10): `Random#nextInt(1, 10000)`
* [JMeter](https://github.com/eugenp/tutorials/blob/master/testing-modules/load-testing-comparison/src/main/resources/scripts/JMeter/Test%20Plan.jmx#L203-L204): `<stringProp name="maximumValue">10000</stringProp>`

As a result: the Grinder is very unlikely to suffer from `custId` collisions. That's why you get exactly `50000` requests with the Grinder and less with the other ones.
You should fix this and have the same behavior for the 3 tests.

**Typo in 3.3**

`The 10 runs` => `The 100 runs`

**Test duration**

The tests last ~1 minutes. This is not a realist usage for load testing where tests lasts way longer. All 3 tools run on the JVM but might benefit differently from the JIT that barely has time to kick in. I would recommend running tests with more iterations, eg 1,000 instead of 100.

**Tools versions**

Performance comparison results only hold for the versions of the 3 tools that were tested.
The article doesn't mention them explicitly, nor their release dates.

If I'm to trust:

* the [pom.xml for the Gatling version](https://github.com/eugenp/tutorials/blob/master/testing-modules/load-testing-comparison/pom.xml#L121): the Gatling version used is 2.2.5, released on 2017-04-20
* The [JMeter test plan for the JMeter version](https://github.com/eugenp/tutorials/blob/master/testing-modules/load-testing-comparison/src/main/resources/scripts/JMeter/Test%20Plan.jmx#L2): 5.1-SNAPSHOT.20181219, so a custom made version from 2018-12-19 (JMeter 5.0 was released on 2018-09-18 while JMeter 5.1 was released on 2019-02-19)

I can't find the Grinder version. 

So if those versions are correct, there are 20 months between the JMeter and Gatling versions.
You can imagine the improvements that happened during this period of time. Gatling 2 has reached end of life in October, 2018.

As you're revisiting the article, I would recommend:
* updating all the tool versions to the latest ones (I can provide a PR for upgrading Gatling)
* mentioning explicitly the tested versions and their respective release dates

lor6:
@slandelle I'm running the tests again with the latest versions, but if there's any improvement you suggest for the Gatling script, please add a PR.

I'll up the numbers and add the versions used in the text.

slandelle:
@lor6 PR incoming.

Regarding the `nextInt`, I would really recommend removing the bound in all the tools. There seems to be a race condition (spring-data, hibernate, h2?) in the application that can cause it to crash on duplicated entries:

```
javax.persistence.NonUniqueResultException: query did not return a unique result: 2
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:128) ~[hibernate-core-5.4.20.Final.jar:5.4.20.Final]
	at org.hibernate.query.internal.AbstractProducedQuery.getSingleResult(AbstractProducedQuery.java:1588) ~[hibernate-core-5.4.20.Final.jar:5.4.20.Final]
	at org.hibernate.query.criteria.internal.compile.CriteriaQueryTypeQueryAdapter.getSingleResult(CriteriaQueryTypeQueryAdapter.java:111) ~[hibernate-core-5.4.20.Final.jar:5.4.20.Final]
	at org.springframework.data.jpa.repository.query.JpaQueryExecution$SingleEntityExecution.doExecute(JpaQueryExecution.java:196) ~[spring-data-jpa-2.3.3.RELEASE.jar:2.3.3.RELEASE]
	at org.springframework.data.jpa.repository.query.JpaQueryExecution.execute(JpaQueryExecution.java:88) ~[spring-data-jpa-2.3.3.RELEASE.jar:2.3.3.RELEASE]
	at org.springframework.data.jpa.repository.query.AbstractJpaQuery.doExecute(AbstractJpaQuery.java:154) ~[spring-data-jpa-2.3.3.RELEASE.jar:2.3.3.RELEASE]
	at org.springframework.data.jpa.repository.query.AbstractJpaQuery.execute(AbstractJpaQuery.java:142) ~[spring-data-jpa-2.3.3.RELEASE.jar:2.3.3.RELEASE]
	at org.springframework.data.repository.core.support.QueryExecutorMethodInterceptor$QueryMethodInvoker.invoke(QueryExecutorMethodInterceptor.java:195) ~[spring-data-commons-2.3.3.RELEASE.jar:2.3.3.RELEASE]
	at org.springframework.data.repository.core.support.QueryExecutorMethodInterceptor.doInvoke(QueryExecutorMethodInterceptor.java:152) ~[spring-data-commons-2.3.3.RELEASE.jar:2.3.3.RELEASE]
	at org.springframework.data.repository.core.support.QueryExecutorMethodInterceptor.invoke(QueryExecutorMethodInterceptor.java:130) ~[spring-data-commons-2.3.3.RELEASE.jar:2.3.3.RELEASE]
```

slandelle:
@lor6 PTAL #10157

lor6:
Thanks @slandelle  The post is updated.

slandelle:
Thanks @lor6!
However, IMHO, you should really align the random behavior on the JMeter and Gatling tests (remove the 10,000 upper bound) to match the behavior of the Grinder test and not compare apples and oranges. 

Also, the numbers in 4.2 don't add up, eg:
* Gatling: 46348 / 59 = 786, not 772
* JMeter: 46369 / 67 = 692, not 694
* Grinder: 50000 : 74 = 676, not 684

lor6:
Hi @slandelle Yep, the upper bound is now removed in all the tests.

I think you're seeing a cached version of the text, as I've now upped the number of runs to 1000 and there are close to 500k requests running. Please check again in a few hours.

slandelle:
@lor6 Awesome, thanks a lot!

FYI, if someday you want to spend time and have more realistic results, here are a few tips:
* The load test tool and the applications should not sit on the same machine. Otherwise, first both compete for CPU, secondly loopback (localhost) is very different from actual network and doesn't produce the same results.
* Use a very fast test application so you're sure this is not the bottleneck. The application we use internally can withstand 200,000 requests per second.
* Run on Linux, not Windows. Java NIO's implementation on Windows is not great. Gatling works best on Linux.

slandelle:
@lor6 I was very surprised with the latest results and tried to reproduce on my side.

In turns out that, whatever the tool, the throughput is quickly dropping. The throughput is initially higher with Gatling, but then the drop is worse, things might be related (Gatling harming the application more).

In short: the current implementation of the test application is completely not suited for a load testing benchmark.
I suspect the culprit is h2 (we've had bad experiences with in in the past too): fine for functional tests, very bad for a demo application under load.
IMHO, if you want to fix this, you should use something like a ConcurrentHashMap, not a h2 database.

slandelle:
response-times-and-throughput

lor6:
@slandelle Yes, I noticed the same behavior. The throughput really starts to drop after 400k requests. For reference, I did run these on a Linux machine, though the initial commands in the text were Windows-specific. I'll update those as well, as Linux is the more common choice anyway.

I could try with MySQL instead, since that could be used for a production db. I'll test with a different db, and with a ConcurrentHashMap, and let you know.

Thanks.

slandelle:
Hi @lor6 
I got it sorted.

2 issues:
* the JSON payloads defined in the Gatling script don't exactly match what the app expects causing an overhead, please see #10167 
* the h2 database doesn't have any index, so the findBy queries get slower and slower as the tables grow. I'm not familiar with SpringBoot and JPA to add those indexes myself. I went with in-memory maps and got the expected stable (and high) throughput, please see #10168

Regards,

PR #10043, opened 2020-09-16
[BAEL-4532]a.getClass() vs A.class

sk1418:


core-java-modules/core-java-lang-3/src/main/java/com/baeldung/getclassobject/Animal.java
```diff
@@ -0,0 +1,5 @@
+package com.baeldung.getclassobject;
+
+public class Animal {
+   protected int numberOfEyes;
```
eric-martin, line 4: 4 space indent (I've fixed the article)

sk1418: Thanks.
Fixed.

core-java-modules/core-java-lang-3/pom.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project
-    xmlns="http://maven.apache.org/POM/4.0.0"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+        xmlns="http://maven.apache.org/POM/4.0.0"
```
KevinGilmore, line 3: As a general guideline, we shouldn't make changes that are whitespace-only.

KevinGilmore (commented):
Looks like the changes to pom.xml are only changes to the whitespace.

core-java-modules/core-java-lang-3/pom.xml
sk1418: @KevinGilmore Thanks. I've reverted the changes to this file.

PR #10049, opened 2020-09-18
BAEL-4156 Does a method's signature in Java include its return type?

stancalau:
BAEL-4156 Does a method's signature in Java include its return type?


core-java-modules/core-java-lang-oop-methods/src/main/java/com/baeldung/signature/OverloadingErrors.java
```diff
@@ -0,0 +1,72 @@
+package com.baeldung.signature;
+
+import java.io.Serializable;
+
+public class OverloadingErrors<T extends Serializable> {
+
+    public void print() {
+        System.out.println("Signature is: method()");
```
eric-martin, line 8: Signature is: print()

PR #10051, opened 2020-09-18
[BAEL-4511] Docker: RUN, CMD or ENTRYPOINT

adrian-maghear:


adrian-maghear:
Not needed for the Docker article

PR #10058, opened 2020-09-20
Bael 4457 list of json objects with rest template

Trixi-Turny:


spring-resttemplate-2/pom.xml
```diff
@@ -48,15 +48,15 @@
...
     <build>
         <plugins>
-        <plugin>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-maven-plugin</artifactId>
-        </plugin>
+            <plugin>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-maven-plugin</artifactId>
+            </plugin>
```
ashleyfrieze, line 59: There's no actual change to this file, so let's revert this one.

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/consumer/controller/UserConsumerController.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.List;
+
+@RestController
+public class UserConsumerController {
+
+    @Autowired
+    private final RestTemplate restTemplate;
+
+    public UserConsumerController(RestTemplate restTemplate) {
```
ashleyfrieze, line 23: You don't need to both provide constructor injection and field injection. You may find that you can take `@Autowired` off the field and things work. Depending on the version of Spring you may need to add `@Autowired` to the constructor, though these days it's not needed.

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/consumer/controller/UserConsumerController.java
```diff
@@ -0,0 +1,49 @@
...
+        this.restTemplate = restTemplate;
+    }
+    private static final String BASE_URL="http://localhost:8080/users";
```
ashleyfrieze, line 26: `static` members should be the first thing in a class - this should be moved up top.

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/consumer/controller/UserConsumerController.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @GetMapping(value="/usersAsListOfObjects")
+    public @ResponseBody
+    Object[] getUsersAsObjects(){
+
+        ResponseEntity<Object[]> responseEntity = restTemplate.getForEntity(BASE_URL, Object[].class);
+        Object[] objects = responseEntity.getBody();
```
ashleyfrieze, line 33: Is it also possible to serialize the response as `User[]` too?

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/model/User.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.List;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class User {
+    private int id;
+    private String name;
+    private List<Address> addressList;
+
+    public User(int id, String name, List<Address> addressList) {
```
ashleyfrieze, line 14: This POJO is more complex than it needs to be for the example. We're just looking for a situation where there's a list of objects coming back from an endpoint. This object has a list of other things in it. That seems off topic.

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/service/UserService.java
```diff
@@ -0,0 +1,9 @@
+package com.baeldung.resttemplate.json.service;
+
+import com.baeldung.resttemplate.json.model.User;
+
+import java.util.List;
+
+public interface UserService {
+    List<User> getUsers() ;
+}
```
ashleyfrieze, line 9: While it may be good practice to extract interfaces from services, this interface seems to complicate this simple example. We could just as easily do without it.

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/service/UserServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+@Service
+public class UserServiceImpl implements UserService {
+    @Override
+    public List<User> getUsers() {
+        return List.of(
+                new User(1, "user1", new ArrayList<Address>(
```
ashleyfrieze, line 16: It might be cleaner to use `Arrays.asList` here, statically importing it. You don't need to assign that to a new `ArrayList`.

spring-resttemplate-2/src/test/java/com/baeldung/resttemplate/json/consumer/controller/UserConsumerControllerUnitTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.internal.bytebuddy.matcher.ElementMatchers.is;
+import static org.junit.Assert.assertEquals;
+import static org.springframework.test.web.client.match.MockRestRequestMatchers.method;
+import static org.springframework.test.web.client.match.MockRestRequestMatchers.requestTo;
+import static org.springframework.test.web.client.response.MockRestResponseCreators.withStatus;
+
+@SpringBootTest
+@ExtendWith(MockitoExtension.class)
```
ashleyfrieze, line 34: I don't see any use of Mockito in this test.

spring-resttemplate-2/src/test/java/com/baeldung/resttemplate/json/consumer/controller/UserConsumerControllerUnitTest.java
```diff
@@ -0,0 +1,81 @@
...
+                .andRespond(withStatus(HttpStatus.OK)
+                                    .contentType(MediaType.APPLICATION_JSON)
+                                    .body(userJson)
+                );
+
+        //when
+        List<User> actual = tested.getUsersAsPOJO();
+        mockServer.verify();
+        assertEquals(actual.size(), expected.size());
+        assertEquals(actual.get(0).getName(), expected.get(0).getName());
```
ashleyfrieze, line 77: Maybe use an assertion where you compare the full list, rather than just checking bits of it. AssertJ's `assertThat(...).containsExactly(...)` would do it, but you can use JUnit's `assertThat(object, hamcrestMatcher)` too if you prefer.

spring-resttemplate-2/src/test/java/com/baeldung/resttemplate/json/provider/controller/UserControllerUnitTest.java
```diff
@@ -0,0 +1,52 @@
...
+import com.baeldung.resttemplate.json.service.UserService;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import static org.mockito.Mockito.when;
+
+
```
ashleyfrieze, line 16: redundant whitespace

spring-resttemplate-2/src/test/java/com/baeldung/resttemplate/json/provider/controller/UserControllerUnitTest.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+
+@ExtendWith(MockitoExtension.class)
+class UserControllerUnitTest {
```
ashleyfrieze, line 23: It may be necessary to unit test controllers in production code, but it's overkill for this example.

spring-resttemplate-2/src/test/java/com/baeldung/resttemplate/json/consumer/controller/UserConsumerControllerUnitTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private UserConsumerController tested = new UserConsumerController(restTemplate);
+
+    @Before
+    public void init() {
+        mockServer = MockRestServiceServer.createServer(restTemplate);
+    }
+
+    @Test
+    public void whenGetUsersAsObjects_thenOK() {
```
ashleyfrieze, line 48: You're only testing one of the two endpoints here.

ashleyfrieze (changes requested):

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/model/User.java
Trixi-Turny: I think the ticket or the stackoverflow link mentioned processing nested objects - I think it is a good example to see

spring-resttemplate-2/pom.xml
```diff
@@ -48,16 +48,7 @@
...
     </dependencies>
-    
-    <build>
-        <plugins>
-        <plugin>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-maven-plugin</artifactId>
-        </plugin>
-        </plugins>
-    </build>
```
ashleyfrieze, line 61: I don't think it's a good idea to remove this plugin from the existing example.

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/consumer/controller/UserConsumerController.java
```diff
@@ -0,0 +1,59 @@
...
+    private static final String BASE_URL="http://localhost:8080/users";
+    private final RestTemplate restTemplate;
+
+    public UserConsumerController(RestTemplate restTemplate) {
+        this.restTemplate = restTemplate;
+    }
+
+    @GetMapping(value="/usersAsArrayOfObjects")
+    public @ResponseBody
+    Object[] getUsersAsObjects(){
```
ashleyfrieze, line 31: `public` seems to be in the wrong place here. I'd expect annotations before `public`. There's a missing space between `()` and `{`

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/consumer/controller/UserConsumerController.java
```diff
@@ -0,0 +1,59 @@
...
+
+        ResponseEntity<Object[]> responseEntity = restTemplate.getForEntity(BASE_URL, Object[].class);
+        Object[] objects = responseEntity.getBody();
+        MediaType contentType = responseEntity.getHeaders().getContentType();
+        HttpStatus statusCode = responseEntity.getStatusCode();
+        return objects;
+    }
+
+
+    @GetMapping(value="/usersAsArrayOfPOJO")
```
ashleyfrieze, line 41: the url doesn't match the behaviour

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/consumer/controller/UserConsumerController.java
```diff
@@ -0,0 +1,59 @@
...
+    public @ResponseBody
+    List<String> getPostCodesFromUserArray(){
```
ashleyfrieze, line 43: same spacing and ordering from the previous. I'd expect:

```java
@GetMapping(...)
@ResponseBody
public List<Foo> getXYZ() {
   ...
}
```

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/consumer/controller/UserConsumerController.java
```diff
@@ -0,0 +1,59 @@
...
+
+        ResponseEntity<User[]> responseEntity = restTemplate.getForEntity(BASE_URL, User[].class);
+        User[] userArray = responseEntity.getBody();
+        MediaType contentType = responseEntity.getHeaders().getContentType();
+        HttpStatus statusCode = responseEntity.getStatusCode();
+        return Arrays.stream(userArray).flatMap(user->user.getAddressList().stream()).map(address->address.getPostCode()).collect(Collectors.toList());
```
ashleyfrieze, line 49: let's break down the streaming operations into a new line by each major function:

```java
return Arrays.stream(userArray)
  .flatMap(user->user.getAddressList().stream())
  .map(Address::getPostCode)  // should use function reference where possible too
  .collect(Collectors.toList());
```

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/consumer/controller/UserConsumerController.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @GetMapping(value="/usersAsListOfPojo")
```
ashleyfrieze, line 52: misleading name

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/model/User.java
ashleyfrieze: Ok.

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/provider/controller/UserController.java
```diff
@@ -0,0 +1,23 @@
...
+public class UserController {
+
+    private final UserService userService;
+
+    public UserController(UserService userService) {
+        this.userService = userService;
+    }
+    @RequestMapping(value = "/users", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
+    public @ResponseBody
+    List<User> getUsers() {
```
ashleyfrieze, line 19: Is this an essential part of the example? Are you not mocking it out in the tests/example code.

We're only interested in the consumer of this API, not the producer.

spring-resttemplate-2/src/test/java/com/baeldung/resttemplate/json/consumer/controller/UserConsumerControllerUnitTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Test
+    public void whenGetUsersAsPOJO_thenOK() {
+        // Given
+        String url = "http://localhost :8080/users";
+        List<String> expected = Arrays.asList("user1", "user2");
+
+        String userJson = "[{\"id\":1,\"name\":\"user1\",\"addressList\":[{\"addressLine1\":\"address1_addressLine1\",\"addressLine2\":\"address1_addressLine2\",\"town\":\"address1_town\",\"postCode\":\"user1_address1_postCode\"}," +
+                                  "{\"addressLine1\":\"address2_addressLine1\",\"addressLine2\":\"address2_addressLine2\",\"town\":\"address2_town\",\"postCode\":\"user1_address2_postCode\"}]}," +
+                                  "{\"id\":2,\"name\":\"user2\",\"addressList\":[{\"addressLine1\":\"address1_addressLine1\",\"addressLine2\":\"address1_addressLine2\",\"town\":\"address1_town\",\"postCode\":\"user2_address1_postCode\"}]}]";
```
ashleyfrieze, line 105: If it's the same user JSON each time, maybe extract to a constant?

ashleyfrieze (changes requested):

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/service/UserService.java
Trixi-Turny: We don't need to include this bit in the article, but as you say it is good practice so I would prefer to leave it in unless you REALLY object it. :)

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/consumer/controller/UserConsumerController.java
Trixi-Turny: Thanks! bad copy and paste..

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/provider/controller/UserController.java
Trixi-Turny: Thank you! I realise this is not needed now.

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/consumer/service/UserConsumerServiceImpl.java
```diff
@@ -0,0 +1,90 @@
...
+        this.restTemplate = restTemplate;
+    }
+
+    @Override
+    public List<String> processUserDataFromObjectArray() {
+        ObjectMapper mapper = new ObjectMapper();
+        ResponseEntity<Object[]> responseEntity = restTemplate.getForEntity(BASE_URL, Object[].class);
+        Object[] objects = responseEntity.getBody();
+        return Arrays.stream(objects)
+                       .map(object -> mapper.convertValue(object, User.class))
```
ashleyfrieze, line 32: These are very heavily indented. Please check the indentation style. I suspect they should be indented by 2 additional spaces, rather than 12.

ashleyfrieze (changes requested):

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/consumer/service/UserConsumerServiceImpl.java
ashleyfrieze: Quick reminder on this one.

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/consumer/service/UserConsumerServiceImpl.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Override
+    public List<String> processUserDataFromUserList() {
+        ResponseEntity<List<User>> responseEntity = restTemplate.exchange(
+                BASE_URL,
+                HttpMethod.GET,
+                null,
+                new ParameterizedTypeReference<List<User>>() {
+                }
```
ashleyfrieze, line 53: Please look at all of the continuation indentation

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/model/Address.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public Address(String addressLine1, String addressLine2, String town, String postCode) {
+        this.addressLine1 = addressLine1;
+        this.addressLine2 = addressLine2;
+        this.town = town;
+        this.postCode = postCode;
+    }
+
+    public Address() {
+        super();
```
ashleyfrieze, line 21: This `super()` is unnecessary - let's remove it.

spring-resttemplate-2/src/test/java/com/baeldung/resttemplate/json/consumer/service/UserConsumerServiceImplUnitTest.java
```diff
@@ -0,0 +1,147 @@
...
+        // Then
+        mockServer.verify();
+        assertEquals(actual.size(), expected.size());
+        assertThat(actual).containsExactly(expected.get(0), expected.get(1));
+    }
+
+    @Test
+    public void whenProcessUserDataAsList_thenOK() {
+        // Given
+        String url = "http://localhost :8080/users";
```
ashleyfrieze, line 84: The space in the URL is unexepcted

spring-resttemplate-2/src/test/java/com/baeldung/resttemplate/json/consumer/service/UserConsumerServiceImplUnitTest.java
```diff
@@ -0,0 +1,148 @@
...
+                .andRespond(withStatus(HttpStatus.OK)
+                                    .contentType(MediaType.APPLICATION_JSON)
+                                    .body(USER_JSON)
+                );
+        List<String> actual = tested.processNestedUserDataFromUserList();
+
+        // Then
+        mockServer.verify();
+        assertEquals(actual.size(), expected.size());
+        assertThat(actual).containsExactly(expected.get(0), expected.get(1), expected.get(2));
```
ashleyfrieze, line 146: A couple of mild stylistic points here:

1. `// Given` and `// When` and `// Then` appeared somewhere on the internet as a commenting standard in tests, and on the whole it adds very little. It may be good to finish the sentence `// Given a user service that returns three users` but otherwise, it's just boilerplate

_Small aside here, your `mockServer.expect` is actually part of your `Given` clause in the above, but appears as a `When`!_

2. There's no need to assert both the size and the exact content of a list - the second automatically implies the first.
2.5. If you're using `assertJ`, and I recommend it a lot, then let's not mix it with JUnit's own `assertEquals`

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/model/Address.java
```diff
@@ -0,0 +1,29 @@
...
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class Address {
+    private final String addressLine1;
+    private final String addressLine2;
+    private final String town;
+    private final String postCode;
+
+    @JsonCreator
+    public Address(
+        @JsonProperty("addressLine1") String addressLine1,
```
eric-martin, line 17: use a two-space indent for continuations

spring-resttemplate-2/src/main/java/com/baeldung/resttemplate/json/model/User.java
```diff
@@ -0,0 +1,30 @@
...
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class User {
+    private final int id;
+    private final String name;
+    private final List<Address> addressList;
+
+    @JsonCreator
+    public User(
+            @JsonProperty("id") int id,
```
eric-martin, line 17: same here

spring-resttemplate-2/src/test/java/com/baeldung/resttemplate/json/consumer/service/UserConsumerServiceImplUnitTest.java
```diff
@@ -0,0 +1,127 @@
...
+    public void init() {
+        mockServer = MockRestServiceServer.createServer(restTemplate);
+    }
+
+    @Test
+    public void whenProcessUserDataAsObjects_thenOK() {
+        String url = "http://localhost:8080/users";
+        List<String> expected = Arrays.asList("user1", "user2");
+
+        mockServer.expect(ExpectedCount.once(),
```
eric-martin, line 42: same here

PR #10077, opened 2020-09-23
Bjakupovic iteration forloop

belms:
Related to : http://jira.baeldung.com/browse/BAEL-4560


core-java-modules/core-java/src/main/java/com/baeldung/iterationcounter/ForLoopIterationCounter.java
```diff
@@ -0,0 +1,49 @@
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class ForLoopIterationCounter {
+    private final List<String> colors = Arrays.asList("red", "blue", "yellow", "green");
```
ashleyfrieze, line 8: This probably ought to be `private static final` and have an uppercase name - e.g. `COLORS`

core-java-modules/core-java/src/main/java/com/baeldung/iterationcounter/ForLoopIterationCounter.java
```diff
@@ -0,0 +1,49 @@
...
+    private void forEachCounter() {
+        int counter = 0;
+        for (String color : colors) {
+            counter++;
+            System.out.println("Value:" + color + ", iteration: " + counter);
+        }
+    }
+
+    public void listIterator() {
+        for (ListIterator<String> color = colors.listIterator(); color.hasNext(); ) {
```
ashleyfrieze, line 19: This is an ugly `for` loop as it has no last clause. It may be better expressed as a `while`. It's not a `forEach` loop

core-java-modules/core-java/src/main/java/com/baeldung/iterationcounter/ForLoopIterationCounter.java
```diff
@@ -0,0 +1,49 @@
...
+            int i = color.nextIndex();
+            System.out.println("Value:" + color.next() + ", iteration: " + (i + 1));
+        }
+    }
+
+    private void forEachLambdaAtomicCounter() {
+        final AtomicInteger indexHolder = new AtomicInteger();
+        colors.forEach(color -> {
+            final int index = indexHolder.incrementAndGet();
```
ashleyfrieze, line 28: The use of `final` here is overkill and adds noise to the text

core-java-modules/core-java/src/main/java/com/baeldung/iterationcounter/ForLoopIterationCounter.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public void listIterator() {
+        for (ListIterator<String> color = colors.listIterator(); color.hasNext(); ) {
+            int i = color.nextIndex();
+            System.out.println("Value:" + color.next() + ", iteration: " + (i + 1));
+        }
+    }
+
+    private void forEachLambdaAtomicCounter() {
+        final AtomicInteger indexHolder = new AtomicInteger();
```
ashleyfrieze, line 26: This also doesn't need to be `final`. It's clearly `effectively final`

ashleyfrieze (changes requested):

core-java-modules/core-java/src/main/java/com/baeldung/iterationcounter/ForLoopIterationCounter.java
belms: should I then remove this section from the code and from the article?

ashleyfrieze: It could be provided as an example around a `while` loop if you think it adds something.

core-java-modules/core-java/src/main/java/com/baeldung/iterationcounter/ForLoopIterationCounter.java
```diff
@@ -0,0 +1,47 @@
...
+        });
+        return counter.intValue();
+    }
+
+    public void forEachLambda(List<String> list) {
+        forEachWithFunctionalInterface(list, (element, i) -> LOGGER.info("Value:" + element + ", iteration: " + (i + 1)));
+    }
+
+    @FunctionalInterface
+    public interface loopWithIndex<T> {
```
ashleyfrieze, line 36: Types don't start with lowercase. We could, reasonably use `BiConsumer` here too. Maybe `LoopWithIndex<T>` could extend `BiConsumer<T, Integer>` or maybe the `forEachWithFunctionalInterface` could just take the `BiConsumer`

`withFunctionalInterface` is a bad name too. It's a Consumer, so `ItemAndIndexConsumer` is a better name for the function to use and the type to use.

core-java-modules/core-java/src/test/java/com/baeldung/iterationcounter/ForLoopIterationCounterUnitTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ForLoopIterationCounterUnitTest {
+    private static final List<String> COLORS = Arrays.asList("red", "blue", "yellow", "green");
+
+
+    @Mock
+    private ForLoopIterationCounter iterationCounter = new ForLoopIterationCounter();;
```
ashleyfrieze, line 24: This seems confused. Two `;`s.

Is it a mock? If so, why are you `new`ing it? Why are you mocking the class under test?

core-java-modules/core-java/src/test/java/com/baeldung/iterationcounter/ForLoopIterationCounterUnitTest.java
```diff
@@ -0,0 +1,48 @@
...
+        int counter = forLoopIterationCounter.forEachLambdaAtomicCounter(COLORS);
+
+        assertEquals(counter, COLORS.size());
+    }
+
+    @Test
+    public void given_listOfStrings_when_forEachLambda_then_returnNumberOfIterations() {
+        iterationCounter.forEachLambda(COLORS);
+
+        verify(iterationCounter, times(1)).forEachLambda(COLORS);
```
ashleyfrieze, line 46: This proves nothing. This proves a method was called on a mock. It doesn't prove that your algorithm did anything.

Also, `times(1)` is redundant. `verify` defaults to `once` unless you specify another `times`

belms: I know, but the purpose of this lambda expression is that we don't have any variables that we use after the for loop that are declared before the loop. So there is no return value. Because if I understood the problem, the goal is to know the iteration while you are in the loop, and this is a perfect example, and I don't think I should change it.  I'll try and figure out how to test this.

core-java-modules/core-java/src/test/java/com/baeldung/iterationcounter/ForLoopIterationCounterUnitTest.java
belms: I needed a mock value for the last useless test, but I'll change it once I figure out how to test that lambda.

ashleyfrieze:
Finished on a different PR

PR #10139, opened 2020-10-07
[BAEL-4556] How to Check if a Java Program Is Running in 64-Bit JVM or 32-Bit JVM

MeenaGawande:
How to Check if a Java Program Is Running in 64-Bit JVM or 32-Bit JVM


java-native/src/main/java/com/baeldung/jvm/JVMBitVersion.java
```diff
@@ -0,0 +1,28 @@
+package com.baeldung.jvm;
```
KevinGilmore, line 1: Let's be a bit more specific (no pun intended) with the package name, e.g. com.baeldung.jvmbitversion

java-native/src/test/java/com/baeldung/jvm/JVMBitVersionUnitTest.java
```diff
@@ -0,0 +1,46 @@
...
+import com.sun.jna.Platform;
+
+public class JVMBitVersionUnitTest {
+
+    private static JVMBitVersion jvmVersion;
+
+    @Before
+    public void setup() {
+        jvmVersion = new JVMBitVersion();
+    }
```
KevinGilmore, line 17: Initializing a static variable in a @Before-annotated method is not considered a good practice because unit tests can be run in parallel. In this case, it's not a big deal since the class is immutable, but I'd suggest either changing the @Before to @BeforeAll or simply making the jvmVersion variable an instance variable (not static).

java-native/src/test/java/com/baeldung/jvm/JVMBitVersionUnitTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Test
+    public void testGetUsingSystemClass() {
```
KevinGilmore, line 20: Use the givenX_whenY_thenZ naming convention for the test methods (the givenX part is optional).

java-native/src/test/java/com/baeldung/jvm/JVMBitVersionUnitTest.java
```diff
@@ -0,0 +1,46 @@
+package com.baeldung.jvm;
```
KevinGilmore, line 1: Change this package as well, so that it's the same as that of the main class.

java-native/src/test/java/com/baeldung/jvmbitversion/JVMBitVersionUnitTest.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Test
+    public void whenUsingSystemClass_thenOutputIsAsExpected() {
+        if (System.getProperty("sun.arch.data.model") == "64") {
+            assertEquals("64-bit", jvmVersion.getUsingSystemClass());
+        } else if (System.getProperty("sun.arch.data.model") == "32") {
+            assertEquals("32-bit", jvmVersion.getUsingSystemClass());
+        }
+
```
KevinGilmore, line 27: let's remove the blank line before this closing brace as per guidelines

PR #10166, opened 2020-10-15
BAEL-4212 (String) or .toString()?

Saber-k:


core-java-modules/core-java-string-operations-3/src/main/java/com/baeldung/tostring/StringCastUtils.java
```diff
@@ -0,0 +1,15 @@
+package com.baeldung.tostring;
+
+public class StringCastUtils {
+    public static String castToString(Object object) {
+        if (object instanceof String)
```
JonCook, line 5: Let's make sure we include parenthesis for the if statements. Generally we follow that rule for our samples.

core-java-modules/core-java-string-operations-3/src/test/java/com/baeldung/tostring/ToStringUnitTest.java
```diff
@@ -0,0 +1,102 @@
+package com.baeldung.tostring;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
JonCook, line 5: Let's avoid * imports and import only the things we need.

core-java-modules/core-java-string-operations-3/src/main/java/com/baeldung/tostring/StringCastUtils.java
```diff
@@ -0,0 +1,15 @@
...
+            return (String) object;
+        return null;
+    }
+
+    public static String getToString(Object object) {
```
JonCook, line 10: Can we think of a better name for this method? getToString sounds a bit weird to me. What about something like objectToString or something a bit more meaningful? Open to ideas.

JonCook (changes requested):
Hey @Saber-k - I left a few comments for review. Thanks.

core-java-modules/core-java-string-operations-3/src/main/java/com/baeldung/tostring/StringCastUtils.java
Saber-k: what about - getStringRepresentation

PR #10170, opened 2020-10-15
The Value of 0xFF Number and Its Uses With &amp; Operation in Java Ar…

abdallahsawan:
…ticle by Abdallah Sawan


java-numbers-4/src/test/java/com/baeldung/number_0xff/Number0xffUnitTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.junit.Test;
+import static org.junit.Assert.assertEquals;
+
+public class Number0xffUnitTest {
+
+    @Test
+    public void test0xFFAssignedToInteger() {
+        int x = 0xff;
+        int expectedValue = 255;
+        assertEquals(x, expectedValue);
```
KevinGilmore, line 12: The expected value should always go before the actual in the assertEquals call.

KevinGilmore (commented):
Please change order of assertEquals arguments to match the method signature: assertEquals(expected, actual)

PR #10174, opened 2020-10-17
Storing Files Indexed by a Database

GilvanOrnelas:


persistence-modules/spring-boot-persistence-2/src/main/java/com/baeldung/db/indexing/FileSystemRepository.java
```diff
@@ -0,0 +1,36 @@
...
+    String RESOURCES_DIR = FileSystemRepository.class.getResource("/")
+        .getPath() + "/";
+
+    default Image saveToFileSystem(Image image) throws Exception {
+        Path newFile = Paths.get(RESOURCES_DIR + new Date().getTime() + "-" + image.name);
+        Files.createDirectories(newFile.getParent());
+
+        Files.write(newFile, image.content);
+
+        image.setName(image.name);
```
ashleyfrieze, line 23: Why are you setting the image name to itself? and why do you have both a setter and a public property on `Image`? I'd rather than we just used getters and setters?

persistence-modules/spring-boot-persistence-2/src/main/java/com/baeldung/db/indexing/FileSystemRepository.java
```diff
@@ -0,0 +1,36 @@
...
+        image.setLocation(newFile.toAbsolutePath()
+            .toString());
+        image.setContent(null);
```
ashleyfrieze, line 26: On the whole, I think it would make cleaner code if the image you saved was a different object than the input. This works, which is fine, but it's kind of dirty as the caller provides you an image with its bytes and you delete them.

Better this should construct `Image imageToSave = new Image(image.getName(), newFile.toAbsolutePath());`

Also note - constructors are better than setters!

persistence-modules/spring-boot-persistence-2/src/main/java/com/baeldung/db/indexing/FileSystemRepository.java
```diff
@@ -0,0 +1,36 @@
...
+
+        return save(image);
+    }
+
+    default byte[] findInFileSystem(Long id) throws Exception {
+        Path image = Paths.get(findById(id).get().location);
```
ashleyfrieze, line 32: Perhaps this should return `Optional<byte[]>` with the code

```java
return findById(id)
   .map(Image::getLocation) // use a getter!!
   .map(this::readFile);
```

with a function `readFile` that does your `readAllBytes`.

Then this code would handle what happens when the file is missing.

persistence-modules/spring-boot-persistence-2/src/main/java/com/baeldung/db/indexing/ImageController.java
```diff
@@ -0,0 +1,39 @@
...
+        dbImage.setContent(image.getBytes());
+
+        return imageRepository.save(dbImage)
+            .getId();
+    }
+
+    @GetMapping(value = "/image/{imageId}", produces = MediaType.IMAGE_JPEG_VALUE)
+    Resource downloadImage(@PathVariable Long imageId) {
+        byte[] image = imageRepository.findById(imageId)
+            .get()
```
ashleyfrieze, line 35: Similar issue here with calling `get` when the `findById` may return `empty`.

persistence-modules/spring-boot-persistence-2/src/test/java/com/baeldung/db/indexing/FileSystemImageIntegrationTest.java
```diff
@@ -0,0 +1,58 @@
...
+@SpringBootTest(classes = ImageUploaderApplication.class)
+@AutoConfigureMockMvc
+@FixMethodOrder
+class FileSystemImageIntegrationTest {
+
+    @Autowired
+    MockMvc mockMvc;
+
+    @Test
+    @Order(1)
```
ashleyfrieze, line 32: Bad practice to use ordered tests - if a test case requires something from earlier, then it's probably one test case.

persistence-modules/spring-boot-persistence-2/src/test/java/com/baeldung/db/indexing/FileSystemImageIntegrationTest.java
```diff
@@ -0,0 +1,58 @@
...
+        InputStream image = classLoader.getResourceAsStream("baeldung.jpeg");
+
+        MockMultipartHttpServletRequestBuilder multipartRequest = MockMvcRequestBuilders.multipart("/file-system/image")
+            .file(new MockMultipartFile("image", "baeldung", MediaType.TEXT_PLAIN_VALUE, image));
+
+        MvcResult result = mockMvc.perform(multipartRequest)
+            .andExpect(status().isOk())
+            .andReturn();
+
+        Assertions.assertThat(result.getResponse()
```
ashleyfrieze, line 44: static import `assertThat`

persistence-modules/spring-boot-persistence-2/src/test/java/com/baeldung/db/indexing/FileSystemImageIntegrationTest.java
```diff
@@ -0,0 +1,58 @@
...
+            .getContentAsString())
+            .isEqualTo("1");
+    }
+
+    @Test
+    @Order(2)
+    void givenBaeldungImage_whenDownloadIt_shallReturnTheImage() throws Exception {
```
ashleyfrieze, line 51: You can see that this test is part of the other one as its `given` describes something that's not in its body.

ashleyfrieze (changes requested):

persistence-modules/spring-boot-persistence-2/src/test/java/com/baeldung/db/indexing/FileSystemImageIntegrationTest.java
GilvanOrnelas: I find this topic discussable. I sure understand your point. But I see a trade-off here. If you consider that your tests should be as clean as possible, we should follow the Single Responsibility Principle. And asserting two different flows (save and find) is a clear violation of it. 
Sure, if we create complex tests relying on order, it may become a mess. But the same is valid for huge test methods with many assertions.

With the order approach, we also have the benefit of a strict view of where is the problem, without depending on logs.

Anyway, I removed the ordering but kept both methods. Now the download test is relying on a mock. Kind of a middle ground. 

As an integration test, I still see more ups than downs in ordering our tests. For these specific cases.

persistence-modules/spring-boot-persistence-2/src/test/java/com/baeldung/db/indexing/FileSystemImageIntegrationTest.java
GilvanOrnelas: From my point of view, the given is something expected for the test run as expected. Usually an argument, but it may be a pre-condition too.
Anyway, it is a little different now.

ashleyfrieze: @GilvanOrnelas you're right that `given` means _I expect this precondition for a test_. However, in test code, it's usual that the `given` part of the test is within the body of the test method, setting up that precondition so that each test can be run independently. In your previous version, for example, it wasn't possible to only execute the second test, as it depended on the execution of the first.

I've seen this done before, sometimes for very good reasons. I even wrote a JUnit framework once to allow tests to express their dependencies, and for the test execution order to be determined from them, rather than have to be hard-coded using `Order`.

However, on the whole, it's well worth avoiding this pattern.

persistence-modules/spring-boot-persistence-2/src/test/java/com/baeldung/db/indexing/FileSystemImageIntegrationTest.java
```diff
@@ -0,0 +1,65 @@
...
+                .isEqualTo("1");
+    }
+
+    @Test
+    void givenBaeldungImage_whenDownloadIt_shallReturnTheImage() throws Exception {
+        ClassLoader classLoader = ClassLoader.getSystemClassLoader();
+        String imagePath = classLoader.getResource("baeldung.jpeg")
+            .getFile();
+
+        when(fileLocationService.find(1L))
```
ashleyfrieze, line 56: As a bonus, you could use `BDDMockito.given` rather than `when`. This is minor and optional. As you're expressing the `given` part of the test, it's awkward to be doing so using functions called `when`. That's a question of preference, though.

persistence-modules/spring-boot-persistence-2/src/test/java/com/baeldung/db/indexing/FileSystemImageIntegrationTest.java
GilvanOrnelas: Cool. I would love to see your "dependency JUnit tests".

TestNG has an interesting feature - dependsOnMethods. Defining dependencies is a little better than forcing an order. I wonder if your framework looks like it.

persistence-modules/spring-boot-persistence-2/src/test/java/com/baeldung/db/indexing/FileSystemImageIntegrationTest.java
GilvanOrnelas: I didn't know BDDMockito. A little more readable, same effort. Amazing! Thanks.

ashleyfrieze (approved):

persistence-modules/spring-boot-persistence-2/src/test/java/com/baeldung/db/indexing/FileSystemImageIntegrationTest.java
```diff
@@ -0,0 +1,69 @@
...
+class FileSystemImageIntegrationTest {
+
+    @Autowired
+    MockMvc mockMvc;
+
+    @MockBean
+    FileLocationService fileLocationService;
+
+    @Test
+    void givenJpegImage_whenUploadIt_shallReturnItsId() throws Exception {
```
KevinGilmore, line 34:
```suggestion
    void givenJpegImage_whenUploadIt_thenReturnItsId() throws Exception {
```

persistence-modules/spring-boot-persistence-2/src/test/java/com/baeldung/db/indexing/FileSystemImageIntegrationTest.java
```diff
@@ -0,0 +1,69 @@
...
+            .andExpect(status().isOk())
+            .andReturn();
+
+        assertThat(result.getResponse()
+            .getContentAsString())
+                .isEqualTo("1");
+    }
+
+    @Test
+    void givenBaeldungImage_whenDownloadIt_shallReturnTheImage() throws Exception {
```
KevinGilmore, line 51:
```suggestion
    void givenBaeldungImage_whenDownloadIt_thenReturnTheImage() throws Exception {
```

persistence-modules/spring-boot-persistence-2/src/test/java/com/baeldung/db/indexing/ImageIntegrationTest.java
```diff
@@ -0,0 +1,72 @@
...
+class ImageIntegrationTest {
+
+    @Autowired
+    MockMvc mockMvc;
+
+    @MockBean
+    ImageDbRepository imageRepository;
+
+    @Test
+    void givenBaeldungJpegImage_whenUploadIt_shallReturnItsId() throws Exception {
```
KevinGilmore, line 36:
```suggestion
    void givenBaeldungJpegImage_whenUploadIt_thenReturnItsId() throws Exception {
```

persistence-modules/spring-boot-persistence-2/src/test/java/com/baeldung/db/indexing/ImageIntegrationTest.java
```diff
@@ -0,0 +1,72 @@
...
+            .andExpect(status().isOk())
+            .andReturn();
+
+        assertThat(result.getResponse()
+            .getContentAsString())
+                .isEqualTo("1");
+    }
+
+    @Test
+    void givenExistingImage_whenDownloadIt_shallReturnHttpStatusOk() throws Exception {
```
KevinGilmore, line 53:
```suggestion
    void givenExistingImage_whenDownloadIt_thenReturnHttpStatusOk() throws Exception {
```

KevinGilmore (commented):
See suggested method renames (change "shall" to "then")

persistence-modules/spring-boot-persistence-2/src/main/java/com/baeldung/db/indexing/Image.java
```diff
@@ -0,0 +1,62 @@
...
+import javax.persistence.GeneratedValue;
+import javax.persistence.Id;
+import javax.persistence.Lob;
+
+@Entity
+class Image {
+
+    @Id
+    @GeneratedValue
+    long id;
```
KevinGilmore, line 13: It's generally considered better to use a wrapper type (Long) for the id instead of the primitive (long).

PR #10232, opened 2020-11-04
Feature/bael 4687 customize jackson json mapper

adinarolea:


spring-boot-modules/spring-boot-jackson/src/main/java/com/baeldung/boot/jackson/config/CoffeeConfiguration.java
```diff
@@ -0,0 +1,60 @@
...
+
+@Configuration
+public class CoffeeConfiguration {
+    public static final String dateTimeFormat = "dd-MM-yyyy HH:mm";
+    private LocalDateTimeSerializer localDateTimeSerializer = new LocalDateTimeSerializer(DateTimeFormatter.ofPattern(dateTimeFormat));
+
+    @Bean
+    public MappingJackson2HttpMessageConverter mappingJackson2HttpMessageConverter() {
+        Jackson2ObjectMapperBuilder builder = new Jackson2ObjectMapperBuilder()
+                .serializers(localDateTimeSerializer)
```
JonCook, line 25: The indentation here should be 4 spaces not 8. This applies throughout the whole PR.

JonCook (changes requested):
Hi @adinarolea 

Nice example. Just a couple of things:
- As mentioned in the Jira ticket this should go in new spring-boot-data-2 module similar to: https://github.com/eugenp/tutorials/tree/master/spring-boot-modules/spring-boot-data
- The indentation isn't correct (8 spaces vs 4 in several places)

You may wish to take a look at our eclipse formatter:
- https://github.com/eugenp/tutorials/blob/master/eclipse/formatter.xml

Thanks

spring-boot-modules/spring-boot-jackson/src/main/java/com/baeldung/boot/jackson/config/CoffeeConfiguration.java
adinarolea: I am not sure what do you want to say, I am using IntelIJ with Eclipse keymap. 
Before public method we have 4 spaces.

JonCook: It is the second line of the indentation which is wrong. In this case it should be:

```
@Configuration
public class CoffeeHttpConverterConfiguration {

    @Bean
    public MappingJackson2HttpMessageConverter mappingJackson2HttpMessageConverter() {
        Jackson2ObjectMapperBuilder builder = new Jackson2ObjectMapperBuilder()
            .serializers(localDateTimeSerializer)
            .serializationInclusion(JsonInclude.Include.NON_NULL);
        return new MappingJackson2HttpMessageConverter(builder.build());
    }
}
```

Likewise for your controller and I think most of the classes:
```
@RestController
public class CoffeeController {

    @GetMapping("/coffee")
    public Coffee getCoffee(@RequestParam(required = false) String brand, @RequestParam(required = false) String name) {
        return new Coffee().setBrand(brand)
            .setDate(LocalDateTime.now())
            .setName(name);
    }
}
```
If you want to break the method signature across two lines, it is OK.

I don't understand what you mean by " I am using IntelIJ with Eclipse keymap."

You can use our eclipse of IntelliJ formatter directly I think:

- https://github.com/eugenp/tutorials/blob/master/eclipse/formatter.xml
- https://github.com/eugenp/tutorials/blob/master/intelliJ/intelliJ-formatter.xml

PR #10260, opened 2020-11-17
Fixed Spring Boot CRUD Application example

hdlopez:
**Issue**

On the [CRUD application example](https://www.baeldung.com/spring-boot-crud-thymeleaf), on `UserController` many functions are redirecting to index (redirect:/index) but the `/index` mapping does not exist.

**Solution**

* Added index mapping on *UserController*
* Refactoring the `findAll` query to make it only on index redirection
* Added unit test for new index mapping


lor6:
Thanks @hdlopez

PR #10270, opened 2020-11-21
GraphQL code fixes and sample queries

tomas-skalicky:
- GraphQL schema files was missing -> added
- PostResolver contained a bug of making Author in Post optional -> fixed
- Security disabled for easier experimenting
- server port changed to 8081 for easier experimenting
- sample queries for faster results


lor6:
Thanks @tomas-skalicky

PR #10298, opened 2020-12-01
BAEL-3948 - Fix test(s) in spring-batch which leaves repository.sqlite changed

JonCook:


JonCook:
The problem was this strange test class which is not referenced in any article.
'''
public class SpringContextTest {

    @Test
    public void testMain() throws Exception {
        App.main(null);
    }
}
'''
I have removed it and now the {{repository.sqlite}} file is left untouched.

PR #10336, opened 2020-12-23
JAVA-3496: Add integration tests for UserRepository

kwoyke:


persistence-modules/spring-data-jpa-query-2/src/test/java/com/baeldung/spring/data/jpa/query/UserRepositoryIntegrationTest.java
```diff
@@ -0,0 +1,162 @@
+package com.baeldung.spring.data.jpa.query;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
```
rozagerardo, line 4: Is there a reason not to use JUnit 5?

persistence-modules/spring-data-jpa-query-2/src/main/java/com/baeldung/spring/data/jpa/query/UserApplication.java
```diff
@@ -0,0 +1,13 @@
+package com.baeldung.spring.data.jpa.query;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+@SpringBootApplication
+public class UserApplication {
```
rozagerardo, line 7: Have you checked this is working? I am getting an error when trying to run this application>
```
***************************
APPLICATION FAILED TO START
***************************

Description:

The bean 'articleRepository', defined in com.baeldung.spring.data.jpa.query.datetime.ArticleRepository defined in @EnableJpaRepositories declared on JpaRepositoriesRegistrar.EnableJpaRepositoriesConfiguration, could not be registered. A bean with that name has already been defined in com.baeldung.spring.data.jpa.query.datetime.ArticleRepository defined in @EnableJpaRepositories declared on JpaRepositoriesRegistrar.EnableJpaRepositoriesConfiguration and overriding is disabled.

Action:

Consider renaming one of the beans or enabling overriding by setting spring.main.allow-bean-definition-overriding=true
```
Besides, do we actually need it? I think we're showing all the functionality in the test, I don't see much use in declaring this application. I might be missing something, of course :)

persistence-modules/spring-data-jpa-query-2/src/test/java/com/baeldung/spring/data/jpa/query/UserRepositoryIntegrationTest.java
kwoyke: Well none of the tests in this module use JUnit 5 (yet). So I think it's better to stick to the pattern for now and create a new task to upgrade all the tests to JUnit 5.

persistence-modules/spring-data-jpa-query-2/src/main/java/com/baeldung/spring/data/jpa/query/UserApplication.java
kwoyke: Yeah that's a good catch - the application interfered with a com.baeldung.spring.data.jpa.query.**datetime** sub package.
I created a single common app for both.

The only reason for the SpringBootApplication I created is that I want to benefit from @DataJpaTest in the tests. Without the app in place we get the "Unable to Find @SpringBootConfiguration ..." error and we'd have to provide some config file anyway. I think the app is a bit cleaner approach. Especially it's now used by both tests - UserRepositoryIntegrationTest and ArticleRepositoryIntegrationTest.

kwoyke: Just noticed we do have an article describing this issue :) See https://www.baeldung.com/spring-boot-unable-to-find-springbootconfiguration-with-datajpatest

persistence-modules/spring-data-jpa-query-2/src/test/java/com/baeldung/spring/data/jpa/query/UserRepositoryIntegrationTest.java
rozagerardo: sounds good @kwoyke 👍

persistence-modules/spring-data-jpa-query-2/src/main/java/com/baeldung/spring/data/jpa/query/UserApplication.java
rozagerardo: Cool, thanks for the info @kwoyke

rozagerardo:
Thanks for the comments @kwoyke . I think the PR looks good, but I don't have write permissions for this repo, so I can't merge it myself. Can you do it? Otherwise I can ping someone :)

PR #10342, opened 2020-12-25
Added changes to show transaction roll back

amitiw4u:


persistence-modules/spring-jpa-2/src/main/java/com/baeldung/spring/transaction/CourseService.java
```diff
@@ -0,0 +1,44 @@
...
+            throw new DataIntegrityViolationException("Throwing exception for demoing Rollback!!!");
+        } catch (Exception e) {
+            TransactionAspectSupport.currentTransactionStatus().setRollbackOnly();
+        }
+    }
+    
+    @Transactional(noRollbackFor = { SQLException.class })
+    public void createCourseDeclarativeWithNoRollBack(Course course) throws SQLException {
+        courseDao.create(course);
+        throw new SQLException("Throwing exception for demoing Rollback!!!");
```
eric-martin, line 41: Let's change the message to "Throwing exception without rollback"

PR #10371, opened 2020-12-30
Java 2824 fix tests in java 9 and above modules

Maiklins:


core-java-modules/core-java-os/src/main/java/com/baeldung/java9/process/OutputStreamExample.java
```diff
@@ -6,8 +6,7 @@
 public class OutputStreamExample {
 
     public static void main(String[] args) {
-        Logger log = Logger.getLogger(OutputStreamExample.class.getName());
-        log.log(Level.INFO, Integer.toString(sum(1,2)));
+        System.out.println(sum(1,2));
```
Maiklins, line 9: The test checks the output stream, which does not work when the result is logged.

PR #10378, opened 2021-01-01
Use shouldNotFilter to exclude URLs for a Filter

tapankavasthi:
Jira ticket:
[BAEL-4703](http://jira.baeldung.com/browse/BAEL-4703)

Summary:
The existing article uses an out-dated way is used to excluded filtering
for certain URLs, instead, `shouldNotFilter` should be used.


spring-web-modules/spring-mvc-basics-3/src/main/java/com/baeldung/exclude_urls_filter/filter/HeaderValidatorFilter.java
```diff
@@ -14,17 +14,17 @@
...
-	    }
 	    String countryCode = request.getHeader("X-Country-Code");
 	    if (!"US".equals(countryCode)) {
 	        response.sendError(HttpStatus.BAD_REQUEST.value(), "Invalid Locale");
 	        return;
 	    }
-	
 	    filterChain.doFilter(request, response);
 	}
+
```
eric-martin, line 24: Let's fix the whitespace in the annotation and method

PR #10379, opened 2021-01-02
BAEL-3636 - introduction to alibaba sentinel

Amitbhave:
PR for [BAEL-3636](http://jira.baeldung.com/browse/BAEL-3636)
A simple example where the flow rule is set. So the controller will serve only one request per second and will invoke fallback if requests exceed.
@davidmartinezbarua


spring-cloud/spring-cloud-sentinel/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+    <dependency>
+      <groupId>com.alibaba.csp</groupId>
+      <artifactId>sentinel-annotation-aspectj</artifactId>
+      <version>1.8.0</version>
+    </dependency>
+  </dependencies>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
```
KevinGilmore, line 34: Use 4-space indent in XML files

spring-cloud/spring-cloud-sentinel/src/main/java/org/spring/cloud/sentinel/SentinelApplication.java
```diff
@@ -0,0 +1,11 @@
+package org.spring.cloud.sentinel;
```
KevinGilmore, line 1: As per the guidelines, the package structure should begin at com.baeldung. For this tutorial, we should use the package com.baeldung.spring.cloud.sentinel.

KevinGilmore (commented):
Please see notes about indentation and package naming.

spring-cloud/spring-cloud-sentinel/pom.xml
Amitbhave: Done..

spring-cloud/spring-cloud-sentinel/src/main/java/org/spring/cloud/sentinel/SentinelApplication.java
Amitbhave: My bad! Changed to com.baeldung.spring.cloud.sentinel

spring-cloud/spring-cloud-sentinel/pom.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0"?>
+<project
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
```
KevinGilmore, line 3: Please use spaces instead of tabs for indentation in this file.

PR #10385, opened 2021-01-04
Bael-4684-Prevent Cross-Site Scripting (XSS) in a Spring application

hmdrzsharifi:


spring-5-security/src/main/java/com/baeldung/xss/XSSFilter.java
```diff
@@ -0,0 +1,50 @@
...
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+        throws IOException, ServletException {
+
+        XSSRequestWrapper wrappedRequest = new XSSRequestWrapper((HttpServletRequest) request);
+
+        String body = IOUtils.toString(wrappedRequest.getReader());
+        if (!"".equals(body)) {
+            JSONObject oldJsonObject = new JSONObject(body);
+            JSONObject newJsonObject = new JSONObject();
+            for (String key : oldJsonObject.keySet()) {
```
ashleyfrieze, line 39: This function doesn't look like it should be here. We have a `stripXSS` function to manipulate an individual value, surely there should be a helper which iterates over a whole JSON object to strip XSS from its fields?

The other concern here is that this only strips XSS from first-level fields, and it assumes that the object doesn't have arrays or child objects.

spring-5-security/src/main/java/com/baeldung/xss/XSSUtils.java
```diff
@@ -0,0 +1,21 @@
+package com.baeldung.xss;
+
+import org.jsoup.Jsoup;
+import org.jsoup.safety.Whitelist;
+import org.owasp.esapi.ESAPI;
+
+public class XSSUtils {
+    private XSSUtils() {
+    }
```
ashleyfrieze, line 9: Remove - not needed. I know some linters ask for this, but they're wrong!

spring-5-security/src/main/java/com/baeldung/xss/XSSUtils.java
```diff
@@ -0,0 +1,21 @@
...
+
+    public static String stripXSS(String value) {
+        if (value != null) {
```
ashleyfrieze, line 12: switch to `if (value == null) { return null; }` and stop redefining `value` throughout this function.

spring-5-security/src/test/java/com/baeldung/xss/PersonControllerUnitTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @LocalServerPort
+    int randomServerPort;
+
+    private static String createPersonUrl;
+    private static RestTemplate restTemplate;
+    private static HttpHeaders headers;
+    private static UriComponentsBuilder builder;
+    private final ObjectMapper objectMapper = new ObjectMapper();
+    private static JSONObject personJsonObject;
```
ashleyfrieze, line 27: It looks like these `private static` variables should actually be local variables in the test. They're most definitely not global variables for the test fixture!

ashleyfrieze (changes requested):

spring-5-security/src/main/java/com/baeldung/xss/PersonController.java
```diff
@@ -0,0 +1,30 @@
...
+    @PostMapping(value = "/person")
+    private ResponseEntity<String> savePerson(@RequestHeader Map<String, String> headers,
+        @RequestParam String param, @RequestBody Person body) {
+        JSONObject response = new JSONObject();
+        headers.forEach((key, value) -> response.put(key, value));
+        response.put("firstName", body.getFirstName());
+        response.put("lastName", body.getLastName());
+        response.put("age", body.getAge());
+        response.put("param", param);
+        return new ResponseEntity<String>(response.toString(), HttpStatus.OK);
```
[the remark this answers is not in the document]
hmdrzsharifi: I don't pass person object to the user. I need a custom json object that contains request headers, params and body.

spring-5-security/src/main/java/com/baeldung/xss/PersonController.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.Map;
+
+@RestController
+@RequestMapping("/personService")
+public class PersonController {
+
+    @PostMapping(value = "/person")
+    private ResponseEntity<String> savePerson(@RequestHeader Map<String, String> headers,
+        @RequestParam String param, @RequestBody Person body) {
+        JSONObject response = new JSONObject();
```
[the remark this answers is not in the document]
hmdrzsharifi: I don't pass person object to the user. I need a custom json object that contains request headers, params and body.

spring-5-security/src/main/java/com/baeldung/xss/SecurityConf.java
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.config.annotation.web.builders.WebSecurity;
+import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
+
+@Configuration
+public class SecurityConf extends WebSecurityConfigurerAdapter {
+
+    @Override
+    public void configure(WebSecurity web) {
+        web.ignoring().antMatchers("/**");
```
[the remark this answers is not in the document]
hmdrzsharifi: Done

spring-5-security/src/main/java/com/baeldung/xss/XSSFilter.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+        throws IOException, ServletException {
+
+        XSSRequestWrapper wrappedRequest = new XSSRequestWrapper((HttpServletRequest) request);
+
+        String body = IOUtils.toString(wrappedRequest.getReader());
+        if (!"".equals(body)) {
```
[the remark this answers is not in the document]
hmdrzsharifi: I have used StringUtils.isBlank(body)

spring-5-security/src/main/java/com/baeldung/xss/XSSFilter.java
hmdrzsharifi: We need it here before sending the request to the requestWrapper. Otherwise, we lose the request body.

spring-5-security/src/main/java/com/baeldung/xss/XSSUtils.java
hmdrzsharifi: Done.

spring-5-security/src/main/java/com/baeldung/xss/XSSUtils.java
hmdrzsharifi: Done.

spring-5-security/src/test/java/com/baeldung/xss/PersonControllerUnitTest.java
hmdrzsharifi: Done.

spring-5-security/src/main/java/com/baeldung/xss/XSSFilter.java
hmdrzsharifi: yes. because it matches the Person class in this article.

spring-5-security/src/main/java/com/baeldung/xss/XSSUtils.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public static String stripXSS(String value) {
+        if (value == null) {
+            return null;
+        }
+        value = ESAPI.encoder()
+            .canonicalize(value);
+        value = value.replaceAll("\0", "");
+        value = Jsoup.clean(value, Whitelist.none());
+        return value;
```
ashleyfrieze, line 17: It's generally bad form to keep redefining a variable - especially one which is also an input parameter to a function. Can we rewrite?

E.g.

```java
// use fluency to reduce need for intermediate variables
String normalized = ESAPI.encoder()
  .canonicalize(value)
  .replaceAll("\0", "");
  
// use a return rather than assign-then-return
return Jsoup.clean(normalized, Whitelist.none());
```

hmdrzsharifi: Done.

ashleyfrieze (approved):

ashleyfrieze:
No longer needed

PR #10389, opened 2021-01-04
[BAEL-4601] How to convert a Hibernate proxy to a real entity object

emyasa:
* Adds Domain Models
* Adds Integration Tests

### DIFF
--- a/persistence-modules/java-jpa-3/src/test/java/com/baeldung/jpa/hibernateunproxy/HibernateProxyIntegrationTest.java
+++ b/persistence-modules/java-jpa-3/src/test/java/com/baeldung/jpa/hibernateunproxy/HibernateProxyIntegrationTest.java
@@ -27,7 +27,7 @@ public class HibernateProxyIntegrationTest {
     @Test
     public void givenPaymentReceipt_whenAccessingPayment_thenVerifyType() {
         PaymentReceipt paymentReceipt = entityManager.find(PaymentReceipt.class, 3L);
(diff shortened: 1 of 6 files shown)


persistence-modules/java-jpa-3/src/test/java/com/baeldung/jpa/hibernateunproxy/HibernateProxyIntegrationTest.java
```diff
@@ -0,0 +1,75 @@
...
+    public static void setup() {
+        EntityManagerFactory factory = Persistence.createEntityManagerFactory("jpa-h2-hibernate-unproxy");
+        entityManager = factory.createEntityManager();
+        populateH2DB();
+    }
+
+    @Test
+    public void givenPaymentReceipt_whenAccessingPayment_thenVerifyType() {
+        PaymentReceipt paymentReceipt = entityManager.find(PaymentReceipt.class, 3L);
+        Assert.assertEquals(true, paymentReceipt.getPayment() instanceof HibernateProxy);
```
KevinGilmore, line 30: Instead of Assert.assertEquals(true, expr), just use Assert.assertTrue(expr). Do this for all three examples that are using this method.

emyasa: Thanks! I've pushed the changes; also, marking this as resolved.

PR #10414, opened 2021-01-10
Update packages to match declaration

wolkenschieber:
Some tests declared package `com.baeldung`, but were located in directory `org.baeldung`. This PR moves the affected tests into the `com.baeldung` directory.


lor6:
Thanks @wolkenschieber Merged.

PR #10421, opened 2021-01-14
[JAVA-3717] Update articles for 2.4.0 - other properties-related articles

rozagerardo:
JIRA: http://team.baeldung.com/browse/JAVA-3717


spring-boot-modules/spring-boot-properties-3/src/main/resources/application-multidocument-integration.properties
```diff
@@ -0,0 +1,4 @@
+spring.datasource.password: 'password'
```
lor6, line 1: should this be "=" instead of ":"?

PR #10475, opened 2021-02-10
Update README.md

michaelzangl:
Typo in Article Title (is in the original article as well, but that is not editable)


lor6:
Thanks @michaelzangl

PR #10488, opened 2021-02-14
BAEL-3938: Add new section on Java 10 Collectors.toUnmodifiableList()

mdhtr:
I had to change the java version to 10 to support Collectors.toUnmodifiableList()

I also added an example of using Java 9's List.copyOf() which is not mentioned in the article. My intention with this was to follow the existing pattern, where there are more cases presented in the tests than in the article itself.

Besides these is the test that is included in the article, following the pattern of the existing tests.


mdhtr:
@ashleyfrieze I'm force-pushing to fix the compile error that fails the CI check. It only affects the first commit where the pom.xml is modified, so I hope if you comment on the tests in the second commit that won't be affected.

PR #10490, opened 2021-02-14
BAEL-4362 - Generate WSDL Stubs with Maven

saleco:
Initial version of codebase for BAEL-4362


saleco:
It will be done in another module

PR #10500, opened 2021-02-18
characters(...) fills StringBuilder instead of String

Kubiac:
I startet to implement an XML parser from this tutorial and ran into a problem with empty elements, e.g. `<elem></elem>`. I found this SO post (https://stackoverflow.com/questions/44570604/sax-parser-is-not-working-properly-when-xml-input-is-given-as-stream-and-some-xm) and realised, that characters(...) can be called zero or multiple times, so one should fixed it conceptually by using a StringBuffer or StringBuilder.


lor6:
Thanks @Kubiac I've merged the PR.

PR #10509, opened 2021-02-21
Feature/bael 4618 final performance

dstr89:


core-java-modules/pom.xml
```diff
@@ -77,6 +77,7 @@
         <module>core-java-lang</module>
         <module>core-java-lang-2</module>
         <module>core-java-lang-3</module>
+		<module>core-java-lang-4</module>
```
KevinGilmore, line 80: use spaces instead of tabs to indent

dstr89: done

PR #10511, opened 2021-02-21
deque vs. stack article

sk1418:


core-java-modules/core-java-collections-4/src/main/java/dequestack/ArrayLifoStack.java
```diff
@@ -0,0 +1,91 @@
+package dequestack;
```
eric-martin, line 1: package com.baeldung.collections.dequestack;

core-java-modules/core-java-collections-4/src/main/java/dequestack/LifoStack.java
```diff
@@ -0,0 +1,12 @@
+package dequestack;
```
eric-martin, line 1: package com.baeldung.collections.dequestack;

core-java-modules/core-java-collections-4/src/main/java/dequestack/ArrayLifoStack.java
sk1418: fixed

core-java-modules/core-java-collections-4/src/main/java/dequestack/LifoStack.java
sk1418: fixed

PR #10619, opened 2021-04-01
Implement a functional logout for spring-boot-keycloak

skinkie:
As title. I think this is a very valuable addition to the tutorial, since this has been one of the most troublesome things in the past.


lor6:
Thanks @skinkie Merged.

PR #10622, opened 2021-04-02
JAVA-3596: Upgrade exec-maven-plugin dependency

kwoyke:


testing-modules/junit-5/pom.xml
```diff
@@ -105,21 +105,6 @@
...
-                <version>${exec-maven-plugin.version}</version>
-                <executions>
-                    <execution>
-                        <goals>
-                            <goal>java</goal>
-                        </goals>
-                    </execution>
-                </executions>
-                <configuration>
-                    <mainClass>com.baeldung.TestLauncher</mainClass>
```
lor6, line 120: I think we should keep this explicit definition in the junit-5 module, as this is the only one that has a TestLauncher class, for this article: https://www.baeldung.com/junit-5-extensions 
and remove the configuration from the junit-5-migration and test-containers modules

kwoyke: Yeah, that sounds good @lor6 Thanks for catching that!

PR #10649, opened 2021-04-12
Add soft delete example

hendrosteven:


bfontana:
will be integrated into some of the existing JPA modules.

PR #10655, opened 2021-04-14
BAEL-4844: Code for the AspectJ class annotation article.

bradhandy:
Code for the article about advising methods in an annotated class.


bradhandy:
Closing because the most recent commit was not showing in this PR.

PR #10675, opened 2021-04-17
[BAEL-4096] Retry in WebFlux

hkhan:


spring-5-webflux/src/test/java/com/baeldung/spring/retry/ExternalConnectorIntegrationTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private void verifyNumberOfGetRequests(int times) {
+        IntStream.range(0, times).forEach(i -> {
+            RecordedRequest recordedRequest = getRecordedRequest();
+            assertThat(recordedRequest.getMethod()).isEqualTo("GET");
+            assertThat(recordedRequest.getPath()).isEqualTo("/data/ABC");
+        });
+    }
+
+    private RecordedRequest getRecordedRequest() {
```
ashleyfrieze, line 93: This would be a lot cleaner if it just did a `throws Exception`, and if the `verifyNumberOfGetRequests` used a `for` loop and also allows `throws Exception`. While `IntStream` is nice for loops, it's not really much cleaner. Similarly, catching exceptions in tests, rather than propagating `throws` is a bit hard to reason about.

spring-5-webflux/src/main/java/com/baeldung/spring/retry/ExternalConnector.java
```diff
@@ -0,0 +1,34 @@
...
+
+@Component
+@AllArgsConstructor
+public class ExternalConnector {
+
+    private static final String PATH_BY_ID = "/data/{id}";
+
+    private final WebClient webClient;
+
+    public Mono<String> getData(String stockId) {
```
ashleyfrieze, line 21: It would be better if this code used a two space continuation wrap, and if the wider lines broke at the nearest `,` or `.` to enable reading.

ashleyfrieze (changes requested):

spring-5-webflux/src/test/java/com/baeldung/spring/retry/ExternalConnectorIntegrationTest.java
hkhan: changed to use ```for``` loop and ```exception``` clean up

spring-5-webflux/src/main/java/com/baeldung/spring/retry/ExternalConnector.java
hkhan: Reformatted using eclipse formatter (in Intellij). 

The continuation wrap seems to be set to 4 in the eclipse formatter scheme:

intellij-code-settings

PR #10683, opened 2021-04-22
Adding soft delete

hendrosteven:


hendrosteven:
https://drafts.baeldung.com/wp-admin/post.php?post=99912&action=edit

PR #10720, opened 2021-05-01
BAEL-4900 AttributeOverwrite explained

polomos:


persistence-modules/spring-data-jpa-annotations/src/main/java/com/baeldung/attribute/overwrite/entity/Car.java
```diff
@@ -0,0 +1,51 @@
...
+
+@Entity
+@AttributeOverride(name = "identifier", column = @Column(name = "VIN"))
+public class Car extends Vehicle {
+
+    private String model;
+    private String name;
+    @Embedded
+    @AttributeOverrides({
+            @AttributeOverride(name = "name", column = @Column(name = "BRAND_NAME", length = 5)),
```
eric-martin, line 15: Line continuations should use a two space indent

polomos: Done.

PR #10789, opened 2021-05-20
Refactored RenameFileUnitTest with Temporary Directory

eas5:
This is a test refactoring. No original assertion was changed nor the original code presented any failing steps.

Problem:
Tests that manipulate external file resources need to guarantee resource integrity and availability. Test suite design and maintainability may suffer from the addition of assurance steps to guarantee parallel execution scenarios and resource leakage from failed executions.

Solution:
The use of @TempDir annotation assures a temporary directory being created and cleaned up for every test method execution, thus simplifying test maintenance steps.


lor6:
Thanks @eas5 That would be the better way to test this in a real-world scenario. In this case, the code is meant to support a tutorial on working with files in Java, so we prefer to keep it simple and focus only on the File API.

PR #10828, opened 2021-05-30
Committing an example of hexagonal architecture

Amitercse:
Making this pull request for my evaluation article- https://drafts.baeldung.com/?p=105033&preview=true.
I created a new project because inside pattern there was already existing example of hexagonal architecture and since it is my first post for evaluation so I didn't create class there but created new project.


Amitercse:
Since this request was created for evaluation of my article and need not to be merged to master so closing it as instructed in hiring mail.

PR #10835, opened 2021-05-30
Hexagonal architecture demo application.

sandeephkamath:


sandeephkamath:
As this is a demo PR closing it as per instruction

PR #10837, opened 2021-05-30
Example code for BAEL-4955 - Understanding the Data Transfer Object (DTO) Pattern

wnoizumi:


persistence-modules/spring-data-jpa-filtering/src/main/java/com/baeldung/projection/view/PersonAddressDTO.java
```diff
@@ -0,0 +1,21 @@
+package com.baeldung.projection.view;
+
+public class PersonAddressDTO {
+
+    private final String firstName;
```
ashleyfrieze, line 5: Ditch this `final` or make both `final`. I prefer to avoid `final` unless I'm very specifically making an immutable object.

persistence-modules/spring-data-jpa-filtering/src/test/java/com/baeldung/projection/JpaProjectionIntegrationTest.java
```diff
@@ -57,4 +61,13 @@ public void whenUsingDynamicProjections_thenObjectWithRequiredPropertiesIsReturn
         assertThat(personView.getFirstName()).isEqualTo("John");
         assertThat(personDto.getFirstName()).isEqualTo("John");
     }
+    
+    @Test
+    public void givenDTOFromTwoTables_whenUsingClassBasedProjections_thenDtoWithRequiredPropertiesIsReturned() {
+        List<PersonAddressDTO> dtos = personRepository.findAllPersonsWithAddress();
+        assertEquals(1, dtos.size());
```
ashleyfrieze, line 68: If you're using `assertJ`'s `assertThat`, then change this to `assertThat(dtos).hasSize(1)`

ashleyfrieze (changes requested):

PR #10838, opened 2021-05-30
[BAEL-4406] Add JaCoCo exclusions

ukhan1980:


maven-modules/maven-plugins/maven-jacoco/src/main/java/com/baeldung/generated/Customer.java
```diff
@@ -0,0 +1,11 @@
+package com.baeldung.generated;
+
+@Generated
+public class Customer {
+    // every thing in this class will be excluded from jacoco report because of @Generated
```
doodoroma, line 5: *everything

PR #10840, opened 2021-05-31
BAEL-4856 Download a binary file using OkHttp

iulian-manda:


libraries-http-2/src/main/java/com/baeldung/okhttp/download/BinaryFileDownloadApplication.java
```diff
@@ -0,0 +1,26 @@
+package com.baeldung.okhttp.download;
+
+import okhttp3.OkHttpClient;
+
+import java.io.FileOutputStream;
+
+public class BinaryFileDownloadApplication {
```
JonCook, line 7: I think we can remove this class? It isn't mentioned in the article and its better to use unit tests to drive and test our functionality.

JonCook (changes requested):
Hi @iulian-manda 

I added a comment to the PR. The other main problem we need to address is that formatting is incorrect on all the code. We're using 4 space indentation. We have an eclipse formatter you can use at:

- https://github.com/eugenp/tutorials/tree/master/eclipse

I believe there is also one for IntelliJ knocking around as well.

Thanks

libraries-http-2/src/main/java/com/baeldung/okhttp/download/BinaryFileDownloadApplication.java
iulian-manda: Definitely, I will remove it, thank you.

PR #10869, opened 2021-06-05
Update Graph.java

omertopuz:
for dfsWithoutRecursion, when a node popped from the stack it should be checked that the current node is visited or not. Since stack has duplicated nodes, some nodes is visired doubly.


lor6:
Thanks @omertopuz I've merged the PR.

PR #10875, opened 2021-06-06
code for baeldung article 

mainak9042:
code for article "Connection timeout vs read timeout for Java sockets"


core-java-modules/core-java-networking-3/src/main/java/com/baeldung/timeout/TcpServerSocket.java
```diff
@@ -0,0 +1,44 @@
+package com.baeldung.timeout;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+//A Java program for a Server
```
KevinGilmore, line 6: Omit comment line

PR #10893, opened 2021-06-09
BAEL-5004 Fixed overflow issues for Large numbers

Simfa11:


KevinGilmore (changes requested):
Revert changes to gradlew.bat and remove the two empty files.

PR #10894, opened 2021-06-10
switch the tomcat plugin to jetty because it's no longer compatible

apixandru:
The currently configured tomcat plugin no longer works with the current dependencies. Changed it to jetty and the project is runnable from maven now.


kwoyke:
Fixed with https://github.com/eugenp/tutorials/pull/11201

PR #10942, opened 2021-06-21
BAEL-4913 - Spring Native

SmartyAnsh:


spring-native/src/main/java/com/baeldung/springnativeintro/SpringNativeApp.java
```diff
@@ -0,0 +1,9 @@
+package com.baeldung.springnativeintro;
+
+public class SpringNativeApp {
+    
+    public static void main(String[] args) {
+        System.out.println("Hello World!, This is Baledung Spring Native Application");
```
KevinGilmore, line 6:
```suggestion
        System.out.println("Hello, World! This is a Baledung Spring Native Application");
```

KevinGilmore (commented):
See suggestion on Hello World message

spring-native/src/main/java/com/baeldung/springnativeintro/SpringNativeApp.java
SmartyAnsh: Done

PR #10957, opened 2021-06-25
[BAEL-4823] Maven plugin management

hkhan:


maven-modules/plugin-management/submodule-1/src/test/java/com/baeldung/CopiesAdditionalResourcesUnitTest.java
```diff
@@ -0,0 +1,20 @@
...
+import java.io.File;
+import java.net.URL;
+import java.util.Objects;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class CopiesAdditionalResourcesUnitTest {
+
+    @Test
+    public void givenAdditionalResource_whenCopyingFromSourceToDestination_thenShouldBeInDestination() {
```
ashleyfrieze, line 14: As we're using `JUnit5`, by convention the `class` and `void` method do not need the `public` modifier. They can just be package-private.

maven-modules/plugin-management/submodule-1/src/test/java/com/baeldung/CopiesAdditionalResourcesUnitTest.java
```diff
@@ -0,0 +1,20 @@
...
+        URL resource = getClass().getClassLoader().getResource("json/include.json");
+        File destinationFile = new File(Objects.requireNonNull(resource).getFile());
```
ashleyfrieze, line 16: As much as `Objects.requireNonNull` is a good programming practice, it complicates this example - let's remove it.

ashleyfrieze (changes requested):
Couple of minor changes.

PR #10995, opened 2021-07-04
source code for bealdung article "Find IP address of client"

mainak9042:
source code for a server and a client application for the baeldung article


core-java-modules/core-java-networking-3/src/main/java/com/baeldung/timeout/TcpServerSocket.java
```diff
@@ -3,6 +3,10 @@
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
+<<<<<<< HEAD
+//A Java program for a Server
+=======
+>>>>>>> upstream/master
```
ashleyfrieze, line 9: Note: there's a strange bit of git merge conflict here that needs resolving.

core-java-modules/core-java-networking-3/src/main/java/com/baeldung/clientaddress/ApplicationServer.java
```diff
@@ -0,0 +1,57 @@
...
+                    out.println("Hello Client !!");
+                } else {
+                    out.println("Not valid message !! ");
+                    break;
+                }
+            }
+            closeIO();
+            stopServer();
+        } catch (IOException e) {
+
```
ashleyfrieze, line 44: empty catch block. How about we remove the try/catch and make the method `throws IOException`?

ashleyfrieze (changes requested):
The reader has no way of running this code if they download it.

Can we add an IntegrationTest that runs it? or a `main` method for them to use?

core-java-modules/core-java-networking-3/src/main/java/com/baeldung/timeout/TcpServerSocket.java
mainak9042: i don't know what happened here. This is from my previous article. I removed the comment

core-java-modules/core-java-networking-3/src/main/java/com/baeldung/clientaddress/ApplicationServer.java
mainak9042: done

core-java-modules/core-java-networking-3/src/main/java/com/baeldung/clientaddress/ApplicationServer.java
```diff
@@ -0,0 +1,60 @@
...
+public class ApplicationServer {
+
+    private ServerSocket serverSocket;
+    private Socket conncetedSocket;
+    private PrintWriter out;
+    private BufferedReader in;
+
+    public void startServer(int port) throws IOException{
+        try {
```
ashleyfrieze, line 19: No catch block

core-java-modules/core-java-networking-3/src/main/java/com/baeldung/clientaddress/ApplicationServer.java
```diff
@@ -0,0 +1,60 @@
...
+            serverSocket = new ServerSocket(port);
+            conncetedSocket = serverSocket.accept();
```
ashleyfrieze, line 21: Typo - should be `connectedSocket`

ashleyfrieze (changes requested):

PR #11009, opened 2021-07-07
[BAEL-4478] IP address is in given range or not using java

KarthickSridhar:


core-java-modules/core-java-networking-3/src/test/java/com/baeldung/ipingivenrange/IPWithGivenRangeCheckUnitTest.java
```diff
@@ -0,0 +1,35 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.junit.jupiter.api.Test;
+
+import com.baeldung.ipingivenrange.IPWithGivenRangeCheck;
+
+class IPWithGivenRangeCheckUnitTest {
+
+    @Test
+    void givenIPAddresses_whenValid_thenCheckInputIPIsInTheRange() throws Exception {
```
doodoroma, line 12: I propose to have one test per behaviour. It would help also to identify the case by the name of the test. 
For example
`givenIPv4Addresses_whenIsInRange_thenReturnsTrue`
`givenIPv4Addresses_whenIsNotInRange_thenReturnsFalse`
`givenIPv6Addresses_whenIsInRange_thenReturnsTrue`
`givenIPv6Addresses_whenIsNotInRange_thenReturnsFalse`

PR #11038, opened 2021-07-15
Also consider string interpolation

GoossensMichael:
The above example only works when logging a string without using string interpolation. For example the following log statement: log.debug("This is a log from {}", "David"). This would result in event.getMessage() returning "This is a log from {}" while you would like to see the processed form "This is a log from David". The getMessage().toString() only returns the former while toString() on the event will return the latter. Furthermore the toString() method is redundant on getMessage() as it already is a string.


lor6:
Thanks @GoossensMichael I've merged the PR.

PR #11043, opened 2021-07-17
BAEL-5030 Printing thread info in log file using log4j

leocampos:
Printing thread info in log file using log4j

### DIFF
--- a/logging-modules/log4j2/src/main/java/com/baeldung/logging/log4j2/threadinfo/ApplicationThread.java
+++ b/logging-modules/log4j2/src/main/java/com/baeldung/logging/log4j2/threadinfo/ApplicationThread.java
@@ -1,0 +1,36 @@
+package com.baeldung.logging.log4j2.threadinfo;
+
+import org.apache.logging.log4j.ThreadContext;
+
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+import java.util.function.BooleanSupplier;
+
+public class ApplicationThread extends Thread {
+    private final User user;
+
+    public ApplicationThread(User user) {
+        this.user = user;
+        user.connect();
+    }
+
+    @Override
+    public void run() {
+        ThreadContext.put("name", user.getName());
+        BooleanSupplier poll;
+
+        do {
+            poll = user.actionPoll();
+
+            sleepRandom();
+        } while (poll == null || poll.getAsBoolean());
+    }
+
+    private void sleepRandom() {
+        try {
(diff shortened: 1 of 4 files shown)


logging-modules/log4j2/src/main/java/com/baeldung/logging/log4j2/threadinfo/Demonstration.java
```diff
@@ -0,0 +1,41 @@
+package com.baeldung.logging.log4j2.threadinfo;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class Demonstration {
+    private static final Logger logger = LogManager.getLogger(Demonstration.class);
```
ashleyfrieze, line 7: As this is `static final` it should have an uppercase name - i.e. `LOGGER`

logging-modules/log4j2/src/main/java/com/baeldung/logging/log4j2/threadinfo/ApplicationThread.java
```diff
@@ -0,0 +1,36 @@
+package com.baeldung.logging.log4j2.threadinfo;
+
+import org.apache.logging.log4j.ThreadContext;
+
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+import java.util.function.BooleanSupplier;
+
+public class ApplicationThread extends Thread {
+    private final User user;
```
ashleyfrieze, line 10: While the `final` here is good practice, we often omit them in sample code for brevity.

logging-modules/log4j2/src/main/java/com/baeldung/logging/log4j2/threadinfo/ApplicationThread.java
```diff
@@ -0,0 +1,36 @@
...
+    @Override
+    public void run() {
+        ThreadContext.put("name", user.getName());
+        BooleanSupplier poll;
+
+        do {
+            poll = user.actionPoll();
+
+            sleepRandom();
+        } while (poll == null || poll.getAsBoolean());
```
ashleyfrieze, line 26: I can't help but feel that this is a bit complex for an example. I see what you're doing, but you're mixing the actions, which are really `Runnable` with the terminate condition, which is really a `keepGoing -> true/false`

I like the action list, but I think the code would be easier to understand if you made the action list of `Runnable` and set a flag for whether the user should stop running.

ashleyfrieze (changes requested):

PR #11045, opened 2021-07-18
Added code to demo expected package error

saikatcse03:


core-java-modules/core-java-lang-4/src/main/java/com/baeldung/bookstore/Book.java
```diff
@@ -0,0 +1,17 @@
...
+package com.baeldung.bookstore;
+
+public class Book {
+
+	private String title;
+	private String author;
+	private long isbn;
+
+	public Book(final String title, final String author, final long isbn) {
+		super();
```
ashleyfrieze, line 11: Unnecessary call to `super()` - please remove

core-java-modules/core-java-lang-4/src/main/java/com/baeldung/bookstore/Book.java
```diff
@@ -0,0 +1,17 @@
+//incorrect package com.baeldung declaration;
+package com.baeldung.bookstore;
+
+public class Book {
+
+	private String title;
```
ashleyfrieze, line 6: These files are formatted with tabs, where it should be spaces.

ashleyfrieze (changes requested):

core-java-modules/core-java-lang-4/src/main/java/com/baeldung/bookstore/Book.java
saikatcse03: I have reformatted using 4 spaces.

core-java-modules/core-java-lang-4/src/main/java/com/baeldung/bookstore/Book.java
saikatcse03: Have removed the super call.

core-java-modules/core-java-lang-4/src/main/java/com/baeldung/bookstore/Book.java
```diff
@@ -0,0 +1,15 @@
+//incorrect package com.baeldung declaration;
```
ashleyfrieze, line 1: I think we need to illustrate the precise incorrect package in this comment - there's words around the "wrong" declaration.

Let's put a line of comment to explain that the next line is commented out to avoid this code failing to build... then let's have the wrong package declaration prefixed by a //

We need to make it clear to reader that if they comment out the package declaration below and uncomment out the one we've left, that they'll replicate the problem.

saikatcse03: I've added few comments as per your review.

PR #11060, opened 2021-07-21
Could com.baeldung:apache-rocketmq:1.0-SNAPSHOT drop off redundant dependencies to loose weight?

Celebrate-future:
@johnA1331 Hi, I am a user of project **_com.baeldung:apache-rocketmq:1.0-SNAPSHOT_**. I found that its pom file introduced **_60_** dependencies. However, among them, **_10_** libraries (**_16%_**) have not been used by your project (the redundant dependencies are listed below). Reduce these useless dependencies can help prevent conflicts between library versions. MeanWhile, it can minimize the total added size to projects. It can also help enable advanced scenarios for this project. 
This PR helps **_com.baeldung:apache-rocketmq:1.0-SNAPSHOT_** lose weight :) I have tested the revised configuration in my local environment. It is safe to remove the unused libraries.

Best regards

### DIFF
--- a/apache-rocketmq/pom.xml
+++ b/apache-rocketmq/pom.xml
@@ -36,7 +36,7 @@
         </dependency>
         
     </dependencies>
-    
+
     <properties>
         <geode.core>1.6.0</geode.core>
         <rocketmq.version>2.0.4</rocketmq.version>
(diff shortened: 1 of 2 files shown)


lor6:
Thanks @Celebrate-future That's true, but we prefer to go with the simplest configuration recommended by the Apache RocketMQ team, which is using the Boot starter they provide.

But you're welcome to make this change in your custom project.

PR #11063, opened 2021-07-22
BAEL-5049 spring-cloud-dapr

gboersma:
Code for article https://drafts.baeldung.com/ja


spring-cloud/spring-cloud-dapr/greeting/src/main/java/org/example/hello/GreetingController.java
```diff
@@ -0,0 +1,24 @@
+package org.example.hello;
+
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/")
+public class GreetingController {
+    @RequestMapping(value = "/", method = RequestMethod.GET)
```
KevinGilmore, line 10: Assuming it's supported in dapr, we should prefer the GetMapping annotation on these controller methods.

KevinGilmore (commented):
Use GetMapping instead of RequestMapping on the controller methods if they only support GET.

PR #11111, opened 2021-08-10
hexagonal architecture first's commit

TonyChinwe:
The first commit from the hexagonal architecture article


TonyChinwe:
After adding the POM file to hexagonal architecture

TonyChinwe:
After renaming the methods of test cases

TonyChinwe:
After renaming my test cases method and correcting one typo

TonyChinwe:
After correcting some more typos after changing the test cases method name

PR #11123, opened 2021-08-13
[JAVA-6221] Fix TimeApi unit test

hkhan:


kwoyke:
Hi @hkhan , PR looks good.
One minor thing though - since you already fix the order of the params in the assertEquals (which is great BTW), would you please fix these in the remaining methods as well?

hkhan:
Hi @kwoyke - sure. 
I've updated the tests - actually using Assertj now (this removes the use of deprecated date API for checking fields and makes tests more readable).

PR #11125, opened 2021-08-15
Java 6390

freelansam:


apache-kafka/pom.xml
```diff
@@ -0,0 +1,181 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-streams</artifactId>
+            <version>${kafka.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>${slf4j.version}</version>
```
kwoyke, line 29: Let's use org.slf4j.version property from the parent pom.xml.

apache-kafka/src/main/java/com/baeldung/flink/model/InputMessage.java
```diff
@@ -55,13 +56,12 @@ public InputMessage(String sender, String recipient, LocalDateTime sentAt, Strin
...
+        if (this == o)
+            return true;
+        if (o == null || getClass() != o.getClass())
+            return false;
         InputMessage message1 = (InputMessage) o;
-        return Objects.equal(sender, message1.sender) &&
-                Objects.equal(recipient, message1.recipient) &&
-                Objects.equal(sentAt, message1.sentAt) &&
-                Objects.equal(message, message1.message);
+        return Objects.equal(sender, message1.sender) && Objects.equal(recipient, message1.recipient) && Objects.equal(sentAt, message1.sentAt) && Objects.equal(message, message1.message);
```
kwoyke, line 64: Let's keep the original formatting - each Objects.equal in its own line.

apache-kafka/pom.xml
freelansam: Done!

apache-kafka/src/main/java/com/baeldung/flink/model/InputMessage.java
freelansam: Done, please check!

PR #11135, opened 2021-08-18
username - yashraj121296

yashraj121296:
submitting demo code for write-up assignment on the topic "A quick and practical example of Hexagonal Architecture in Java".


yashraj121296:
added a new module for hexagonal Architecture demo application Application user name

yashraj121296:
added a new module for hexagonal Architecture demo application Application in the second commit

PR #11141, opened 2021-08-18
Bael 5082

hajk1:
code for the BAEL-5082 jira issue


hajk1:
close in order to open another PR

PR #11161, opened 2021-08-24
JAVA-3587 Update Jackson version in the main pom

Maiklins:


apache-libraries/src/test/java/com/baeldung/apache/beam/intro/WordCountUnitTest.java
```diff
@@ -9,8 +9,8 @@
 
 public class WordCountUnitTest {
     
-    @Test
-    // @Ignore
+//    @Test
```
kwoyke, line 12: Is there any reason we want to ignore this test?

Maiklins: @kwoyke I have re-run the build including the test and now it works fine.

PR #11195, opened 2021-09-02
JAVA-5947 - Removed the large Geo Location File - GeoLite2-City.mmdb from code an…

anuragkumawat:
…d it's references.


anuragkumawat:
JAVA-5947 - Removed the large Geo Location File - GeoLite2-City.mmdb from code and updated the tests with comment to download file and run manual tests

PR #11273, opened 2021-10-01
BAEL-5143

andbi:
BAEL-5143 Introduction to the Evrete Rule Engine, sources and tests


rule-engines/evrete/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+
+    <parent>
+        <groupId>com.baeldung</groupId>
+        <artifactId>rule-engines</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+
+    <build>
+
+    </build>
```
KevinGilmore, line 23: you can omit the empty build section

rule-engines/evrete/src/main/java/com/baeldung/evrete/introduction/IntroductionAJR.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Random;
+
+public class IntroductionAJR {
+    public static void main(String[] args) throws IOException {
+        ClassLoader classLoader = IntroductionAJR.class.getClassLoader();
+        KnowledgeService service = new KnowledgeService();
+        URL rulesetUrl = classLoader.getResource("rules/CustomerRuleset.java");
+        Knowledge knowledge = service.newKnowledge(
+                "JAVA-SOURCE",
+                rulesetUrl
```
KevinGilmore, line 22: Use four-space indent per level in this and other Java files.

rule-engines/pom.xml
```diff
@@ -17,6 +17,7 @@
         <module>easy-rules</module>
         <module>openl-tablets</module>
         <module>rulebook</module>
+        <module>evrete</module>
```
KevinGilmore, line 20: Move this line up after easy-rules so that the modules are listed alphabetically

KevinGilmore (approved):

PR #11284, opened 2021-10-04
BAEL-4858] Trusting all certificates using OkHttp

vunamtien:


libraries-http-2/src/main/java/com/baeldung/okhttp/ssl/TrustAllCertsClient.java
```diff
@@ -0,0 +1,46 @@
...
+
+                    @Override
+                    public void checkServerTrusted(java.security.cert.X509Certificate[] chain, String authType) {
+                    }
+
+                    @Override
+                    public java.security.cert.X509Certificate[] getAcceptedIssuers() {
+                        return new java.security.cert.X509Certificate[]{};
+                    }
+                }
```
KevinGilmore, line 30: This block should begin only 4 (not 8) spaces indented from line 16.

KevinGilmore (changes requested):
Adjust indentation in one block (see notes on file)

libraries-http-2/src/main/java/com/baeldung/okhttp/ssl/TrustAllCertsClient.java
vunamtien: Thank you, Kevin Gilmore.
I fixed it.

KevinGilmore (approved):

PR #11304, opened 2021-10-09
specifying relative path

raghudinkavijaykumar:
Fixes issue 11200.


lor6:
The issue was fixed in the meantime, closing here.

PR #11365, opened 2021-10-24
Bael 4241

saikatcse03:
Added All class listing code

### DIFF
--- a/core-java-modules/core-java-jvm-2/pom.xml
+++ b/core-java-modules/core-java-jvm-2/pom.xml
@@ -33,6 +33,16 @@
             <artifactId>jol-core</artifactId>
             <version>${jol-core.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.reflections</groupId>
+            <artifactId>reflections</artifactId>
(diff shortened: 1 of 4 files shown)


core-java-modules/core-java-jvm-2/pom.xml
```diff
@@ -33,6 +33,16 @@
...
+        <dependency>
+            <groupId>org.reflections</groupId>
+            <artifactId>reflections</artifactId>
+            <version>0.9.12</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>31.0.1-jre</version>
+        </dependency>
```
KevinGilmore, line 45: Let's extract the versions into properties.

KevinGilmore (commented):
In pom.xml, extract dependency version strings into properties.

core-java-modules/core-java-jvm-2/pom.xml
saikatcse03: Now extracting from properiies

PR #11417, opened 2021-11-05
[BAEL-5211] - How to test a Java Optional

lucaCambi77:
A few tests on how to test a Java Optional

### DIFF
--- a/testing-modules/testing-assertions/src/test/java/com/baeldung/optional/OptionalUnitTest.java
+++ b/testing-modules/testing-assertions/src/test/java/com/baeldung/optional/OptionalUnitTest.java
@@ -25,7 +25,9 @@ class OptionalUnitTest {
 
     @Test
     public void givenOptionalWithValue_whenAssertEqualsOptionalObject_thenOk() {
-        assertEquals(Optional.of(OPTIONAL_RETURN_VALUE), Optional.of(OPTIONAL_RETURN_VALUE));
+        Optional<String> expected = Optional.of(OPTIONAL_RETURN_VALUE);
+        Optional<String> actual = Optional.of(OPTIONAL_RETURN_VALUE);
+        assertEquals(expected, actual);
     }
 
     @Test
--- a/testing-modules/testing-assertions/src/test/java/com/baeldung/optional/OptionalUnitTest.java
+++ b/testing-modules/testing-assertions/src/test/java/com/baeldung/optional/OptionalUnitTest.java
@@ -9,61 +9,48 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-public class OptionalUnitTest {
+class OptionalUnitTest {
 
-    private final String stringToTest = "testString";
+    private static final String optionalReturnString = "optionalReturnString";
 
     @Test
     public void givenOptionalWithValue_whenAssertThatIsNotEmpty_thenOk() {
-        assertThat(optionalWithValueMethodToTest()).isNotEmpty();
+        assertThat(Optional.of(optionalReturnString)).isNotEmpty();
     }
 
     @Test
     public void givenOptionalWithValue_whenAssertThatHasValue_thenOk() {
-        assertThat(optionalWithValueMethodToTest()).hasValue(stringToTest);
+        assertThat(Optional.of(optionalReturnString)).hasValue(optionalReturnString);
     }
 
     @Test
     public void givenOptionalWithValue_whenAssertEqualsOptionalObject_thenOk() {
-        assertEquals(Optional.of(stringToTest), optionalWithValueMethodToTest());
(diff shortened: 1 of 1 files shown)


testing-modules/testing-assertions/src/test/java/com/baeldung/optional/OptionalUnitTest.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Optional;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class OptionalUnitTest {
+
+    private final String stringToTest = "testString";
```
ashleyfrieze, line 14: This should probably be `private static final String EXPECTED_STRING` - in that there's no instance-level aspect to it.

Similarly, this isn't the string we're testing - it's the result we're expecting in our tests.

testing-modules/testing-assertions/src/test/java/com/baeldung/optional/OptionalUnitTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.Optional;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class OptionalUnitTest {
```
ashleyfrieze, line 12: By convention JUnit 5 tests don't have `public` modifiers on methods or the class

testing-modules/testing-assertions/src/test/java/com/baeldung/optional/OptionalUnitTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Test
+    public void givenEmptyOptional_whenAssertIsNotPresent_thenOk() {
+        Optional<String> optional = optionalIsEmptyMethodToTest();
+        assertFalse(optional.isPresent());
+    }
+
+    private Optional<String> optionalWithValueMethodToTest() {
+        return Optional.of(stringToTest);
+    }
```
ashleyfrieze, line 64: You've `DRY`ed out the tests by doing extract method here of a common bit of data, but it makes the test harder to follow. Inline these. In general, a trivial bit of code like this is easier re-written every time in a test, and that's especially true for example snippets in articles.

testing-modules/testing-assertions/src/test/java/com/baeldung/optional/OptionalUnitTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @Test
+    public void givenOptionalWithValue_whenAssertEqualsGet_thenOk() {
+        Optional<String> optional = optionalWithValueMethodToTest();
+        assertEquals(stringToTest, optional.get());
+    }
+
+    @Test
+    public void givenOptionalWithValue_whenAssertTrueIsPresentAndGet_thenOk() {
+        Optional<String> optional = optionalWithValueMethodToTest();
+        assertTrue(optional.isPresent() && stringToTest.equals(optional.get()));
```
ashleyfrieze, line 40: This is a bad test - it's a bad bit of advice. If this fails, we don't know why.

ashleyfrieze (changes requested):

testing-modules/testing-assertions/src/test/java/com/baeldung/optional/OptionalUnitTest.java
lucaCambi77: yes now that i see it i agree

testing-modules/testing-assertions/src/test/java/com/baeldung/optional/OptionalUnitTest.java
lucaCambi77: I agree but I was on to follow the StackOverflow example

testing-modules/testing-assertions/src/test/java/com/baeldung/optional/OptionalUnitTest.java
lucaCambi77: yes that is true

testing-modules/testing-assertions/src/test/java/com/baeldung/optional/OptionalUnitTest.java
lucaCambi77: ok i'll change it accordingly

testing-modules/testing-assertions/src/test/java/com/baeldung/optional/OptionalUnitTest.java
ashleyfrieze: Let's make our own version of the example that suits our needs.

SO examples aren't always the best ideas ;)

PR #11444, opened 2021-11-12
fixing missing parameter

janmadle:
Fixed not working example - Added missing zipParameters to .addFile() function


lor6:
Thanks @janmadle

PR #11452, opened 2021-11-14
Update PostRestController.java

G0ldmember:
The id is already part of the postDto in this case. And the controller method does not have a @PathVariable annotation to evaluate the /{id} parameter. Guess the id should be transmitted in the RequestBody in the PostDto Object? Otherwise, id would be redundant, once in the PathVariable and once in the PostDto.


kwoyke:
This issue has already been addressed in https://github.com/eugenp/tutorials/pull/11634
We actually decided to use the path variable id.
Even though we have this duplication - id value in the path and in the body - I think it's better than presenting the PUT method with /posts uri instead of /posts/{id}.

PR #11462, opened 2021-11-17
BAEL-5208: Get the name of the currently executing test in JUnit

Yashasvii:


testing-modules/junit-5-advanced/src/test/java/com/baeldung/getTestName/JUnit4ParameterizedTestNameUnitTest.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(Parameterized.class)
+public class JUnit4ParameterizedTestNameUnitTest {
+
+    private final String input;
+    private final String expected;
```
ashleyfrieze, line 18: Let's add a new line after these fields. There's also no strong reason to make them `final`. It's not _wrong_ but it doesn't add much.

ashleyfrieze (changes requested):
Minor change to one of the files

PR #11487, opened 2021-11-21
[JAVA-8353] Update the test to wait for assertion with timeout

hkhan:


core-java-modules/core-java-exceptions-3/src/test/java/com/baeldung/exceptions/illegalmonitorstate/IllegalMonitorStateExceptionUnitTest.java
```diff
@@ -1,7 +1,10 @@
 package com.baeldung.exceptions.illegalmonitorstate;
 
+import com.google.common.util.concurrent.Uninterruptibles;
```
kwoyke, line 3: Let's get rid of this unnecessary import.

PR #11527, opened 2021-11-28
Update schema.sql

kuldiegor:
fix: All SQL commands (statements) must be separated by a line break. If there is one command then there is one line.


kwoyke:
There are multiple schema.sql files in the module that actually contain multi-line DDL statements. All the tables are properly created and all the tests work fine as well.
I don't think we need to adjust any of the schema.sql in the spring-jdbc module.

PR #11537, opened 2021-12-01
BAEL-3763: Improvement: Add a section about the ISO_INSTANT formatter

azhwani:
Simple PR

### DIFF
--- a/core-java-modules/core-java-datetime-string/src/test/java/com/baeldung/datetime/DateTimeFormatterUnitTest.java
+++ b/core-java-modules/core-java-datetime-string/src/test/java/com/baeldung/datetime/DateTimeFormatterUnitTest.java
@@ -165,7 +165,7 @@ public class DateTimeFormatterUnitTest {
     public void shouldPrintFormattedZonedDateTime() {
         ZonedDateTime zonedDateTime = ZonedDateTime.of(2021, 02, 15, 0, 0, 0, 0, ZoneId.of("Europe/Paris"));
         String formattedZonedDateTime = DateTimeFormatter.ISO_INSTANT.format(zonedDateTime);
-        Assert.assertEquals(formattedZonedDateTime, "2021-02-14T23:00:00Z");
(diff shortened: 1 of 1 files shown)


core-java-modules/core-java-datetime-string/src/test/java/com/baeldung/datetime/DateTimeFormatterUnitTest.java
```diff
@@ -159,4 +160,33 @@ public void shouldExpectAnExceptionIfDateTimeStringNotMatchPattern() {
         DateTimeFormatter zonedFormatter = DateTimeFormatter.ofPattern("dd.MM.yyyy HH:mm z");
         ZonedDateTime.from(zonedFormatter.parse("31.07.2016 14:15"));
     }
+
+    @Test
+    public void shouldPrintFormattedZonedDateTime() {
+        ZonedDateTime zonedDateTime = ZonedDateTime.of(2021, 02, 15, 0, 0, 0, 0, ZoneId.of("Europe/Paris"));
+        String formattedZonedDateTime = DateTimeFormatter.ISO_INSTANT.format(zonedDateTime);
+        Assert.assertEquals(formattedZonedDateTime, "2021-02-14T23:00:00Z");
```
KevinGilmore, line 168: swap the order of arguments so that it's assertEquals(expected, actual) -- do this for the other test that uses assertEquals as well. And for both, add an empty line before the assertion.

KevinGilmore (changes requested):
Swap order of arguments in assertEquals; add blank line before assertEquals

core-java-modules/core-java-datetime-string/src/test/java/com/baeldung/datetime/DateTimeFormatterUnitTest.java
azhwani: Done

KevinGilmore (approved):

PR #11539, opened 2021-12-01
[JAVA-8703] Fix Live test to allow isolated execution

hkhan:


testing-modules/spring-testing-2/src/test/java/com/baeldung/dynamicproperties/ArticleTestFixtureLiveTest.java
```diff
@@ -23,8 +27,11 @@ void givenAnArticle_whenPersisted_thenShouldBeAbleToReadIt() {
         article.setContent("Today's applications...");
 
         articleRepository.save(article);
-        Article persisted = articleRepository.findAll().get(0);
-        assertThat(persisted.getId()).isNotNull();
+
+        List<Article> allArticles = articleRepository.findAll();
+        assertThat(allArticles).hasSize(1);
```
hkhan, line 32: I've added this assertion to verify the DB state and ensure that we're using a separate DB instance, per test class. Without this, we could unintentionally run the test against the same instance as another Live test and the test would still pass!

PR #11554, opened 2021-12-06
BAEL-5204: Change the font color of a cell with Apache POI

HarisHashim:
Class CellStyler that set font styles as well as unit test


HarisHashim:
@eric-martin ready for code review

PR #11580, opened 2021-12-14
[JAVA-8377] Split spring-boot-testing module

hkhan:


spring-boot-modules/spring-boot-testing-2/pom.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>spring-boot-testing-2</artifactId>
+    <name>spring-boot-testing-2</name>
+    <packaging>war</packaging>
```
kwoyke, line 8: Let's stick to jar.

spring-boot-modules/spring-boot-testing-2/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <finalName>spring-boot-testing</finalName>
+        <resources>
```
kwoyke, line 31: I don't think we need to explicitly set the filtering to true.

spring-boot-modules/spring-boot-testing-2/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+            <resource>
+                <directory>src/main/resources</directory>
+                <filtering>true</filtering>
+            </resource>
+        </resources>
+        <plugins>
```
kwoyke, line 37: Do we really need any of the following plugins in this simple module?

spring-boot-modules/spring-boot-testing-2/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+            </plugin>
+        </plugins>
+    </build>
+
+    <properties>
+        <!-- The main class to start by executing java -jar -->
+        <start-class>com.baeldung.boot.Application</start-class>
+        <git-commit-id-plugin.version>2.2.4</git-commit-id-plugin.version>
+        <gmavenplus-plugin.version>1.6</gmavenplus-plugin.version>
+        <spring-boot.version>2.5.0</spring-boot.version>
```
kwoyke, line 87: This way you explicitly overwrite the Boot version defined in the parent-boot-2. Any reason for that?

kwoyke:
One additional general comment. I don't think we need the maven wrapper here. As a rule of thumb we don't use it across our modules. I know it has some benefits and also a few modules "use" it. But let's keep it simple unless there's a good reason to use the wrapper.

PR #11606, opened 2021-12-20
BAEL-4316 Sample apps to showcase how to add files from outside docker context

swapanpramanick2004:
BAEL-4316 Sample apps to showcase how to add files from outside docker context

- Sample App
- Dockerfiles
- Docker-compose file
- Scripts


docker/docker-sample-app/README.md
```diff
@@ -1,3 +1,3 @@
 ### Relevant Articles:
 
-- How to Get Docker-Compose to Always Use the Latest Image
+- How to Include Files Outside of Docker’s Build Context
```
ashleyfrieze, line 3: Let's remove this - we don't update the article list as part of these PRs

ashleyfrieze (changes requested):

docker/docker-sample-app/README.md
ashleyfrieze: Quick reminder that this needs reverting

ashleyfrieze (changes requested):

docker/docker-sample-app/README.md
swapanpramanick2004: Ohh, sorry, it was actually a mistake. I thought I did this change in the directory docker/docker-sample-ecomm-app and the change would go away when I deleted the directory. But it seems I changed thee wrong file when I made this change in the first place.

PR #11640, opened 2021-12-30
[JAVA-8364] Split spring-5-reactive-2 module

hkhan:


spring-5-reactive-3/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+        <dependency>
+            <groupId>io.projectreactor</groupId>
+            <artifactId>reactor-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <profiles>
+        <profile>
+            <id>integration-lite-first</id>
```
kwoyke, line 43: Do we really need this extra config? Shouldn't Spring Boot pick up logback-test.xml "out of the box"?

spring-5-reactive-3/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+	<include
+		resource="org/springframework/boot/logging/logback/base.xml" />
+	<appender name="LISTAPPENDER"
```
kwoyke, line 5: There's no ListAppender in this module.

hkhan: removed

spring-5-reactive-3/pom.xml
hkhan: yes, you're right, logback-test.xml is used by default - fixed.

kwoyke:
retest this please

PR #11657, opened 2022-01-08
[JAVA-9347] Split spring-rest-http module

hkhan:


spring-web-modules/spring-rest-http-2/src/test/java/com/baeldung/requesttimeout/RequestTimeoutUnitTest.java
```diff
@@ -3,13 +3,15 @@
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.context.junit4.SpringRunner;
 import org.springframework.web.reactive.function.client.WebClient;
 import org.springframework.web.reactive.function.client.WebClientRequestException;
 
 @SpringBootTest
+@ActiveProfiles("prod")
```
hkhan, line 12: Had to fix the test folder structure and activate this profile in order to switch off swagger (springfox). It appears that this version of SpringBoot has issues with springfox (which causes the app to fail at startup)

PR #11661, opened 2022-01-08
BAEL-5289 Java IdentityHashMap Class

SeshuTechie:


java-collections-maps-3/src/main/java/com/baeldung/map/identity/IdentityHashMapDemonstrator.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(title, year);
+        }
+
+        @Override
+        public String toString() {
+            return "Book{" +
```
doodoroma, line 84: Line continuation should have 2-space indentation.

PR #11680, opened 2022-01-12
BAEL-5263 : Access Entity Manager with Spring Boot and Spring Data

lucaCambi77:
- [x] user repository and custom find using EntityManager
- [x] integration test

### DIFF
--- a/spring-boot-modules/spring-boot-data-2/src/main/java/com/baeldung/boot/customRepository/repository/CustomUserRepositoryImpl.java
+++ b/spring-boot-modules/spring-boot-data-2/src/main/java/com/baeldung/boot/customRepository/repository/CustomUserRepositoryImpl.java
@@ -8,7 +8,7 @@ import javax.persistence.PersistenceContext;
 
 import com.baeldung.boot.customRepository.model.User;
 
-public class CustomUserRepositoryImp implements CustomUserRepository {
+public class CustomUserRepositoryImpl implements CustomUserRepository {
 
     @PersistenceContext
     private EntityManager entityManager;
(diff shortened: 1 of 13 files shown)


spring-boot-modules/spring-boot-data-2/src/main/java/com/baeldung/boot/customRepository/repository/CustomUserRepositoryImpl.java
```diff
@@ -0,0 +1,27 @@
...
+import com.baeldung.boot.customRepository.model.User;
+
+public class CustomUserRepositoryImpl implements CustomUserRepository {
+
+    @PersistenceContext
+    private EntityManager entityManager;
+
+    @Override
+    public User customFindMethod(Long id) {
+        return (User) entityManager.createQuery("from User u where u.id = :id")
```
andrzejsydor, line 18:
```suggestion
        return (User) entityManager.createQuery("FROM User u WHERE u.id = :id")
```

spring-boot-modules/spring-boot-data-2/src/test/java/com/baeldung/boot/customRepository/CustomRepositoryUnitTest.java
```diff
@@ -0,0 +1,28 @@
...
+import com.baeldung.boot.customRepository.repository.UserRepository;
+
+@SpringBootTest(classes = CustomRepositoryApplication.class)
+public class CustomRepositoryUnitTest {
+
+    @Autowired
+    private UserRepository userRepository;
+
+    @Test
+    public void givenCustomRepository_whenInvokeCustomFindMethod_thenEntityIsFound() {
```
andrzejsydor, line 19: junit5 tests could be default package scope
```suggestion
    void givenCustomRepository_whenInvokeCustomFindMethod_thenEntityIsFound() {
```

PR #11691, opened 2022-01-13
maven to gradle

ankit0706:
A small java maven project which is converted into gradle


gradle/maven-to-gradle/src/main/java/com/sample/javacode/DisplayTime.java
```diff
@@ -0,0 +1,21 @@
...
+public class DisplayTime {
+
+	public static void main(String[] args) {
+		Date date = new java.util.Date();
+		System.out.println(date);
+		String gmt = date.toString();
+		boolean flag = StringUtils.isEmpty(gmt);
+		if(flag) {
+			System.out.println("No time found");
+		}else {
```
andrzejsydor, line 16: please format code

ankit0706: I have imported the eclipse formatter into my IDE and formatted the code. Please let me know if it's fine now.

JavierPeris: The format is correct now.

gradle/maven-to-gradle/pom.xml
```diff
@@ -0,0 +1,15 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
JavierPeris, line 1: This file is using tabs, please convert to spaces.

ankit0706: I tried to fix it. Please let me know if it is fine now.

JavierPeris: Looks good!

PR #11762, opened 2022-01-31
BAEL-4513 Gradle source vs target compatibility

kpentaris:
Draft url: https://drafts.baeldung.com/wp-admin/post.php?post=126456 
Jira link: http://jira.baeldung.com/browse/BAEL-4513


gradle/gradle-source-vs-target-compatibility/src/main/java/com/baeldung/helloworld/HelloWorldApp.java
```diff
@@ -0,0 +1,9 @@
+package com.baeldung.helloworld;
+
+public class HelloWorldApp {
+
+  public static void main(String[] args) {
+    System.out.println("Hello World!");
+  }
+
+}
```
JavierPeris, line 9: Please use the formatter included in the repo.

PR #11777, opened 2022-02-04
[JAVA-9585] Convert Cassandra integration tests to Live tests

hkhan:


persistence-modules/spring-data-cassandra/src/test/java/com/baeldung/SpringContextIntegrationTest.java
```diff
@@ -25,17 +25,17 @@
 
 @RunWith(SpringJUnit4ClassRunner.class)
 @ContextConfiguration(classes = CassandraConfig.class)
-public class SpringContextTest {
+public class SpringContextIntegrationTest {
```
kwoyke, line 28: Please leave it as SpringContextTest. This kind of tests are "special" and they are explicitly included in the profiles (in the main pom.xml).

hkhan: Yes, I've seen it in the live-all profile (or is it used somewhere else as well?) 

However, it looks like it'll be part of the same profile even as IntegrationTest. 

In addition it will get run as part of integration tests profiles which might be useful to ensure that context is not broken (especially as we're moving all the tests to LiveTest).

Should I rename this back to SpringContextTest?

hkhan: yes sure

PR #11778, opened 2022-02-04
JAVA-7535 Moved ebook specific code snippets to maven-simple module

dkapil:


maven-modules/maven-simple/maven-dependency/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+		<artifactId>maven-modules</artifactId>
+		<version>0.0.1-SNAPSHOT</version>
+	</parent>
+
+	<dependencyManagement>
+		<dependencies>
+			<dependency>
+				<groupId>org.apache.commons</groupId>
+				<artifactId>commons-lang3</artifactId>
+				<version>3.12.0</version>
```
kwoyke, line 21: Can we use commons-lang3.version property from the main pom.xml?

dkapil: done

PR #11813, opened 2022-02-13
BAEL-5039: Add a new section on using a datepicker to submit date values

achraftt:


glmartin (changes requested):
In the HTML, the indentation should be 4 spaces.

PR #11818, opened 2022-02-14
fix improper package import

ixuzhi:


lor6:
Thanks @ixuzhi But this is actually the correct package used in this article: https://www.baeldung.com/spring-boot-soap-web-service The files will be generated in the springsoap/gen location when you run the Maven build.
The files in the springsoap/client/gen location are used in https://www.baeldung.com/spring-soap-web-service

PR #11865, opened 2022-02-25
aws-lambda: todo-reminder: added cron trigger as stated in the article

chillbirdo:
The corresponding article ( https://www.baeldung.com/java-enterprise-aws-lambda#2-example-problem ) says: "Let's create an app that runs every few minutes". However, it seems that a scheduled event trigger has been forgotten, so the function is never invoked.
I added a scheduled event that invokes the function every 10 minutes. Tested on AWS.

### DIFF
--- a/aws-lambda/todo-reminder/template.yaml
+++ b/aws-lambda/todo-reminder/template.yaml
@@ -20,3 +20,9 @@ Resources:
         Variables:
           PARAM1: VALUE
           ENV_NAME: !Ref EnvironmentName
+      Events:
+        ScheduledEvent:
+          Type: Schedule
+          Properties:
+            Schedule: cron(0/10 * * * ? *)
+            Enabled: True


aws-lambda/todo-reminder/template.yaml
```diff
@@ -19,4 +19,10 @@ Resources:
...
           PARAM1: VALUE
-          ENV_NAME: !Ref EnvironmentName
\ No newline at end of file
+          ENV_NAME: !Ref EnvironmentName
+      Events:
+        ScheduledEvent:
+          Type: Schedule
+          Properties:
+            Schedule: cron(0/10 * * * ? *)
+            Enabled: True
```
ashleyfrieze, line 28: This is definitely a valid addition to the template. However, the focus of the article was primarily the implementation of the lambda. There are lots of ways this Lambda might be invoked periodically, and the article didn't go into that.

We _could_ add this change, and even update the article, but it's kind of off topic.

chillbirdo: When I was going through your article I was happy to read that we would code a function that runs every few minutes, because it was exactly what I was looking for. After studying and running the code I found out that it does not do that, so I came back to the article wondering whether I had missed or misunderstood something. So it was kind of confusing.
 The rest of the article was **awesome** and helped me lot! Many thanks, cheers, Chillbirdo

PR #11901, opened 2022-03-07
[BAEL-5420] Added value object creation

mayank4all:


core-java-modules/core-java-lang-4/src/main/java/com/baeldung/employee/EmployeeBean.java
```diff
@@ -0,0 +1,48 @@
+package com.baeldung.employee;
+
+import java.io.Serializable;
+import java.time.LocalDate;
+
+public class EmployeeBean implements Serializable {
```
andrzejsydor, line 6: just `Employee`? Why Bean?
```suggestion
public class Employee implements Serializable {
```

core-java-modules/core-java-lang-4/src/main/java/com/baeldung/employee/EmployeeVO.java
```diff
@@ -0,0 +1,40 @@
+package com.baeldung.employee;
+
+import java.time.LocalDate;
+import java.util.Objects;
+
+public class EmployeeVO {
+    public String firstName;
```
andrzejsydor, line 7: why `public`?
Should be `private` like `startDate`

core-java-modules/core-java-lang-4/src/main/java/com/baeldung/employee/EmployeePOJO.java
```diff
@@ -0,0 +1,49 @@
+package com.baeldung.employee;
+
+import java.time.LocalDate;
+import java.util.Objects;
+
+public class EmployeePOJO {
+
+    public String firstName;
```
andrzejsydor, line 8:
```suggestion
    private String firstName;
```

andrzejsydor (commented):
could be improve a bit

core-java-modules/core-java-lang-4/src/main/java/com/baeldung/employee/EmployeeBean.java
mayank4all: > 

We are showing the conversion of POJO to a JavaBean. Using the name EmployeeBean gives more context to the readers.

PR #11966, opened 2022-03-23
Return Map from GraphQL

mabdullah12:
Three techniques have been used

1. Return as Json String
2. Return Json using GraphQL scalar type
3. Return as list of key-value pair


mabdullah12:
retest this please

PR #11975, opened 2022-03-24
Spring Boot Properties Migrator Demo

s9m33r:


spring-boot-modules/spring-boot-properties-migrator-demo/src/main/java/com/baeldung/spring/boot/properties/migrator/Main.java
```diff
@@ -0,0 +1,11 @@
+package com.baeldung.spring.boot.properties.migrator;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+@SpringBootApplication
+public class Main {
+	public static void main(String[] args) {
+		SpringApplication.run(Main.class, args);
+	}
```
KevinGilmore, line 10: Let's use spaces instead of tabs for the indentation.

s9m33r: Ahh sure. I had changed the formatter in between fixed now.

PR #11994, opened 2022-03-31
Update UserController.java

arvind-kumar-avinash:
Changed the name and email address of the second and the third user to avoid confusion for the readers.


lor6:
Thanks @arvind-kumar-avinash I've merged the PR.

PR #12039, opened 2022-04-09
[BAEL-3634] Code for Spark DataFrame article

ukhan1980:


apache-spark/docker/docker-compose.yaml
```diff
@@ -0,0 +1,33 @@
+version: "3"
```
KevinGilmore, line 1: Do we need any docker files for this?

PR #12106, opened 2022-04-21
Assertions in JUnit 4 and JUnit 5 suggestions

Alex-Golub:
- [Let's start reviewing the assertions available also in JUnit 4](https://www.baeldung.com/junit-assertions#:~:text=Let%27s%20start%20reviewing%20the%20assertions%20available%20also%20in%20JUnit%204.). Its under Junit 5 thus should be Junit **5**
- [blacklisted exception](https://www.baeldung.com/junit-assertions#:~:text=blacklisted%20exception). Is marked [deprecated](https://junit.org/junit5/docs/current/user-guide/#:~:text=7.3.-,Deprecated%20APIs,-The%20following%20table) according to Junit5 docs
- [fast-forward marker](https://www.baeldung.com/junit-assertions#:~:text=expected%20line%20is%20a-,fast%2Dforward%20marker,-.%20If%20yes%20apply). Example should be given

A valid fast-forward marker is a string that starts and ends with >> and contains at least four characters. Any character between the fast-forward literals is discarded.
```
>>>>
>> stacktrace >>
>> single line, non Integer.parse()-able comment >>
```

### DIFF
--- a/testing-modules/junit5-migration/src/test/java/com/baeldung/junit5vsjunit4assertions/Junit4AssertionsUnitTest.java
+++ b/testing-modules/junit5-migration/src/test/java/com/baeldung/junit5vsjunit4assertions/Junit4AssertionsUnitTest.java
@@ -1,5 +1,6 @@
 package com.baeldung.junit5vsjunit4assertions;
 
+import org.hamcrest.MatcherAssert;
 import org.junit.Test;
 
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


lor6:
Thanks @Alex-Golub I've merged the PR and clarified point #1 in the article. 
I'll add an internal improvement to look closer into points 2 and 3.

PR #12116, opened 2022-04-23
BAEL-5562: Check if a character is a vowel in Java

ankit0706:
https://jira.baeldung.com/browse/BAEL-5562


ashleyfrieze:
Replacing this with an alternative implementation

PR #12118, opened 2022-04-24
Bael 5481 java httpclient post

dstr89:


core-java-modules/core-java-httpclient/src/test/java/com/baeldung/httpclient/PostRequestMockServer.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.io.IOException;
+import java.net.ServerSocket;
+import java.net.URISyntaxException;
+
+import static org.mockserver.integration.ClientAndServer.startClientAndServer;
+import static org.mockserver.model.HttpRequest.request;
+import static org.mockserver.model.HttpResponse.response;
+
+public class PostRequestMockServer {
```
ashleyfrieze, line 17: I don't think that this is a great way to plug a mock server into a test. Ideally we'd make a JUnit 5 extension instead of inheriting a class like this. However if you don't want to get bogged down with that, then maybe make this class `abstract`.

ashleyfrieze (changes requested):

pom.xml
```diff
@@ -1348,6 +1348,7 @@
                  <module>core-java-modules/core-java-networking-3</module>
                  <module>core-java-modules/multimodulemavenproject</module>
                  <module>core-java-modules/core-java-strings</module>
+                 <module>core-java-modules/core-java-httpclient</module>
```
dstr89, line 1351: FYI @ashleyfrieze I just saw in the docs that there is separate profile for Java 9+ to be run with Jenkins

dstr89:
retest this please

PR #12178, opened 2022-05-07
[JAVA-11770] Added build step to copy native libraries into a specific folder

panos-kakos:


aws-modules/aws-miscellaneous/pom.xml
```diff
@@ -78,6 +78,25 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <version>3.1.1</version>
```
kwoyke, line 84: Please use a property here.

PR #12196, opened 2022-05-10
Update pom.xml

dkapil:


kvaga:
how to use it in my maven project?

kvaga:
I mean I have a maven project. How can I add github's maven project to my local maven dependencies?

PR #12201, opened 2022-05-12
Bael 5296: Added Http Request Header using the Feign Client

saikatcse03:
In this pr, I have added a few additional packages which contain feign client-specific request header implementation.
Also, looks like earlier PR for Cassandra was not merged.


feign/src/main/java/com/baeldung/feign/clients/builder/BookFeignClientBuilder.java
```diff
@@ -0,0 +1,27 @@
...
+import com.baeldung.feign.header.interceptor.AuthRequestInterceptor;
+
+import feign.Feign;
+import feign.gson.GsonDecoder;
+import feign.gson.GsonEncoder;
+
+
+public class BookFeignClientBuilder {
+
+    public <T> T createClient(Class<T> type, String uri) {
```
ashleyfrieze, line 12: These methods should probably be `static`. There's nothing `instance`ey about this code. We don't access the builder via an interface and the methods don't use any fields of the builder class.

Also, I'm surprised by the use of the Gson library here in that Jackson is more popular. However, it's not relevant to the meat of the article.

feign/src/main/java/com/baeldung/feign/header/interceptor/AuthRequestInterceptor.java
```diff
@@ -0,0 +1,21 @@
...
+
+import feign.RequestInterceptor;
+import feign.RequestTemplate;
+
+
+public class AuthRequestInterceptor implements RequestInterceptor {
+   
+    @Override
+    public void apply(RequestTemplate template) {
+        template.header("Authorisation", "Bearer " + this.getAuthorisationToken());
```
ashleyfrieze, line 13: I think the use of a getter for the auth token is misleading here. This approach would never work in practice. In practice, we'd construct the interceptor by either passing it the auth token to use, or by giving it an object that can fetch the current auth token.

The simplest approach is to make the auth token a String field of this class and to pass that token in on construction. That's real world, and a good example for the reader to follow.

feign/src/test/java/com/baeldung/feign/clients/header/dynamicheader/BookClientLiveTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * Consumes https://github.com/Baeldung/spring-hypermedia-api
+ */
+@Slf4j
```
ashleyfrieze, line 16: Let's not use Lombok to create loggers for tests.

Tests shouldn't log, and it makes the test code more oblique to add this.

feign/src/test/java/com/baeldung/feign/clients/header/dynamicheader/BookClientLiveTest.java
```diff
@@ -0,0 +1,44 @@
...
+    	
+        Book book = new Book(isbn, "Me", "It's me!", null, null);
+        
+        Map<String,Object> headerMap = new HashMap<>();
+    	
+        headerMap.put("metadata-key1", "metadata-value1");
+        headerMap.put("metadata-key2", "metadata-value2");   	
+    	
+        bookClient.create(headerMap, book);
+        log.info("{}", book);
```
ashleyfrieze, line 42: Tests should, ideally, end in an assertion.

feign/src/test/java/com/baeldung/feign/clients/header/dynamicheader/BookClientLiveTest.java
```diff
@@ -0,0 +1,44 @@
...
+public class BookClientLiveTest {
+	
+    private BookFeignClientBuilder feignClientBuilder;
+
+    private BookClient bookClient;
+
+    @Before
+    public void setup() {
+        feignClientBuilder = new BookFeignClientBuilder();
+        bookClient = feignClientBuilder.createClient(BookClient.class, "http://localhost:8081/api/books");
```
ashleyfrieze, line 26: Can we make this test run against a wiremock or mockserver, so the reader can easily run it themselves. I think it's a bad thing to expect them to start up another example app.

feign/src/test/java/com/baeldung/feign/clients/header/interceptor/BookClientLiveTest.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.UUID;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Consumes https://github.com/Baeldung/spring-hypermedia-api
+ */
+@Slf4j
```
ashleyfrieze, line 19: Remove SL4j

feign/src/test/java/com/baeldung/feign/clients/header/interceptor/BookClientLiveTest.java
```diff
@@ -0,0 +1,61 @@
...
+    public void givenBookClient_shouldPostBook() throws Exception {
+        String isbn = UUID.randomUUID()
+          .toString();
+        Book book = new Book(isbn, "Me", "It's me!", null, null);
+        bookClient.create(book);
+
+        book = bookClient.findByIsbn(isbn)
+          .getBook();
+        assertThat(book.getAuthor(), is("Me"));
+        log.info("{}", book);
```
ashleyfrieze, line 59: remove logging statements

feign/src/test/java/com/baeldung/feign/clients/header/staticheader/BookClientLiveTest.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.UUID;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Consumes https://github.com/Baeldung/spring-hypermedia-api
+ */
+@Slf4j
```
ashleyfrieze, line 18: Remove

feign/src/test/java/com/baeldung/feign/clients/header/staticheader/parameterized/BookClientLiveTest.java
```diff
@@ -0,0 +1,37 @@
...
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Consumes https://github.com/Baeldung/spring-hypermedia-api
+ */
+@Slf4j
+public class BookClientLiveTest {
+	
+	private BookFeignClientBuilder feignClientBuilder;
```
ashleyfrieze, line 18: Indentation

feign/src/test/java/com/baeldung/feign/clients/header/staticheader/parameterized/BookClientLiveTest.java
```diff
@@ -0,0 +1,37 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Consumes https://github.com/Baeldung/spring-hypermedia-api
+ */
+@Slf4j
```
ashleyfrieze, line 15: Remove

feign/src/test/java/com/baeldung/feign/clients/header/staticheader/parameterized/BookClientLiveTest.java
```diff
@@ -0,0 +1,37 @@
...
+        feignClientBuilder = new BookFeignClientBuilder();
+        bookClient = feignClientBuilder.createClient(BookClient.class, "http://localhost:8081/api/books");
+    }
+
+    @Test
+    public void givenBookClient_shouldFindOneBook() throws Exception {
+        Book book = bookClient.findByIsbn(requester, "0151072558")
+          .getBook();
+        assertThat(book.getAuthor(), containsString("Orwell"));
+        log.info("{}", book);
```
ashleyfrieze, line 35: Remove logging statements

persistence-modules/java-cassandra/src/main/java/com/baeldung/cassandra/batch/Application.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.baeldung.cassandra.batch.domain.Product;
+import com.baeldung.cassandra.batch.repository.KeyspaceRepository;
+import com.baeldung.cassandra.batch.repository.ProductRepository;
+import com.datastax.oss.driver.api.core.CqlSession;
+
+public class Application {
```
ashleyfrieze, line 13: The cassandra example files seem to be unrelated to this article.

ashleyfrieze (changes requested):

feign/src/main/java/com/baeldung/feign/clients/builder/BookFeignClientBuilder.java
saikatcse03: Yes. I have changed to static now.  for Gson, i have reused this module so Gson was already used in this module.  But yes Jackson is what i have alos used in all other instances .

feign/src/main/java/com/baeldung/feign/header/interceptor/AuthRequestInterceptor.java
saikatcse03: I thought not to make the code for AuthInterceptor too complicated with additional classes, so put one simple method in there. 
Now, I have passing an Interface to this Auth Interceptor and in runtime , i will pass one object which has simple Auth token implementation. So in nutshell, the AuthInterceptor is not generating the token rather delegating to another object instance. Definitely in actual scenario the Overriden class needs to deal with that logic in different way.

feign/src/test/java/com/baeldung/feign/clients/header/dynamicheader/BookClientLiveTest.java
saikatcse03: Removed SLF4J from these unit tests.

feign/src/test/java/com/baeldung/feign/clients/header/dynamicheader/BookClientLiveTest.java
saikatcse03: Asserts are there. have removed the logs messages

feign/src/test/java/com/baeldung/feign/clients/header/dynamicheader/BookClientLiveTest.java
saikatcse03: i feel the wiremock or mockserver will dilute the example for feign and we might need to explain the wiremock itself. So this code is running with actual endpoint which pretty easy to setup.  Also, previous article has use the same example. So, reusability was also a factor for this article.
But definitely, this is very good point to keep in mind.

feign/src/test/java/com/baeldung/feign/clients/header/staticheader/parameterized/BookClientLiveTest.java
saikatcse03: corrected

feign/src/test/java/com/baeldung/feign/clients/header/staticheader/parameterized/BookClientLiveTest.java
saikatcse03: Removed

feign/src/test/java/com/baeldung/feign/clients/header/staticheader/parameterized/BookClientLiveTest.java
saikatcse03: Logging are removed

persistence-modules/java-cassandra/src/main/java/com/baeldung/cassandra/batch/Application.java
saikatcse03: I was able to remove this. though these were all merged but still showing in my current branch. but after fetching from upstream master had fixed that.

PR #12203, opened 2022-05-12
Update : with = application.properties

selllami:


lor6:
Thanks @selllami That's a good point. I'll close the PR here because the file was moved in the meantime and now there's a conflict. But I'll add the fix in the code.

PR #12208, opened 2022-05-14
Update UserController.java

MWest2020:
updated for 2022 . Code in tutorial wasn't working


lor6:
Thanks @MWest2020 We'll take a closer look at the code.

kwoyke:
fixed in https://github.com/eugenp/tutorials/pull/12489

PR #12209, opened 2022-05-14
Update JavalinApp.java

MWest2020:
most importantly... :id changed to {id}.

exmaple above no post, in the tutorial, pint 6, there's a post.


kwoyke:
fixed in https://github.com/eugenp/tutorials/pull/12489

PR #12238, opened 2022-05-20
BAEL-5557 Split a stream into parts

mdabrowski-eu:
https://jira.baeldung.com/browse/BAEL-5557

### DIFF
--- a/core-java-modules/core-java-streams-4/src/test/java/splitting/Article.java
+++ b/core-java-modules/core-java-streams-4/src/test/java/splitting/Article.java
@@ -1,0 +1,11 @@
+package splitting;
+
+public class Article {
+    String target;
+    boolean featured;
+
+    public Article(String target, boolean featured) {
+        this.target = target;
+        this.featured = featured;
+    }
+}
--- a/core-java-modules/core-java-streams-4/src/test/java/splitting/Article.java
+++ b/core-java-modules/core-java-streams-4/src/test/java/splitting/Article.java
@@ -1,11 +1,37 @@
 package splitting;
 
-public class Article {
-    String target;
-    boolean featured;
+class Article {
+    private final String target;
+    private final boolean featured;
 
     public Article(String target, boolean featured) {
         this.target = target;
         this.featured = featured;
+    }
+
+    public String getTarget() {
+        return target;
+    }
+
+    public boolean isFeatured() {
+        return featured;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        Article article = (Article) o;
+
+        if (featured != article.featured) return false;
+        return target.equals(article.target);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = target.hashCode();
+        result = 31 * result + (featured ? 1 : 0);
+        return result;
     }
 }
--- a/core-java-modules/core-java-streams-4/src/test/java/splitting/StreamSplittingUnitTest.java
+++ b/core-java-modules/core-java-streams-4/src/test/java/splitting/StreamSplittingUnitTest.java
@@ -1,7 +1,6 @@
 package splitting;
 
 import com.google.common.collect.Lists;
-import org.junit.Assert;
 import org.junit.Test;
 import rx.Observable;
 
@@ -10,8 +9,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
 public class StreamSplittingUnitTest {
-    // given
     private final static List<Article> articles = Lists.newArrayList(
             new Article("Baeldung", true),
             new Article("Baeldung", false),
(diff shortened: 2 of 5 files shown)


core-java-modules/core-java-streams-4/src/test/java/splitting/Article.java
```diff
@@ -0,0 +1,11 @@
+package splitting;
+
+public class Article {
+    String target;
+    boolean featured;
```
ashleyfrieze, line 5: Should probably be `private`

core-java-modules/core-java-streams-4/src/test/java/splitting/StreamSplittingUnitTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.junit.Test;
+import rx.Observable;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class StreamSplittingUnitTest {
+    // given
```
ashleyfrieze, line 14: Remove these comments. Partly because Baeldung code doesn't have random comments in. Partly because these comments don't say anything. See this article - https://codingcraftsman.wordpress.com/2019/07/19/lets-add-cruft-to-our-tests/ - warning it's quite an opinionated article, written as though I'm angry... read it as though I'm smiling while I point out how we've been taught to do weird things in our code.

core-java-modules/core-java-streams-4/src/test/java/splitting/StreamSplittingUnitTest.java
```diff
@@ -0,0 +1,86 @@
...
+            new Article("The Code", false)
+    );
+
+    @Test
+    public void givenListOfArticles_shouldSplitListInTwoCategories_usingPartitioningBy() {
+        // when
+        Map<Boolean, List<Article>> groupedArticles = articles.stream().collect(Collectors.partitioningBy(a -> a.target.equals("Baeldung")));
+
+        // then
+        Assert.assertEquals(2, groupedArticles.get(true).size());
```
ashleyfrieze, line 28: Let's static import the `assertEquals` - though to be honest, I'd rather use assertJ here - `assertThat(groupedArticles.get(true)).hasSize(2)`

On top of that, there's a question around whether a map of `Boolean` to `List<Article>` is a nice thing to have or not.

And if we added getters or `equals`/`hashCode` to `Article`, we could actually compare the precise articles that were split, rather than just the count

core-java-modules/core-java-streams-4/src/test/java/splitting/Article.java
ashleyfrieze: And let's add getters to access the fields

ashleyfrieze (changes requested):

core-java-modules/core-java-streams-4/src/test/java/splitting/Article.java
mdabrowski-eu: Done.

core-java-modules/core-java-streams-4/src/test/java/splitting/StreamSplittingUnitTest.java
mdabrowski-eu: Ok, that's a take I didn't hear before. I removed given/when/then comments.

core-java-modules/core-java-streams-4/src/test/java/splitting/StreamSplittingUnitTest.java
mdabrowski-eu: Normally I also prefer AssertJ but from what I've seen in the Baeldung codebase JUnit assertions are more prevalent so I opted for consistency.
I added equals and hashCode and improved tests accordingly.

core-java-modules/core-java-streams-4/src/test/java/splitting/StreamSplittingUnitTest.java
```diff
@@ -0,0 +1,96 @@
...
+    public void givenListOfArticles_shouldSplitListInTwoOVerlappingCategories_usingTeeing() {
+        List<List<Article>> groupedArticles = articles.stream().collect(Collectors.teeing(
+                Collectors.filtering(article -> article.getTarget().equals("Baeldung"), Collectors.toList()),
+                Collectors.filtering(Article::isFeatured, Collectors.toList()),
+                List::of));
+
+        assertEquals(2, groupedArticles.get(0).size());
+        assertEquals(1, groupedArticles.get(1).size());
+
+        assertTrue(groupedArticles.get(0).containsAll(Lists.newArrayList(
```
ashleyfrieze, line 71: This use of `assertTrue` indicates a weak assertion. Rather than use an assertion that itself checks the contents of a list and reports on the differences, we've written the logic of the assertion ourselves and just done `assertTrue`. If this assertion false, you'll get `was false, should be true`, which doesn't tell you anything about the data being asserted.

I recommend switching to AssertJ for these assertions - `assertThat(groupedArticles.get(0)).containsExactly(...)`

mdabrowski-eu: Ok, I don't like mixing assertions so I migrated all of them to assertj.

PR #12253, opened 2022-05-23
BAEL-5623 reverse list article

sk1418:


core-java-modules/core-java-collections-list-4/src/main/java/com/baeldung/list/reverse/ReverseArrayList.java
```diff
@@ -0,0 +1,23 @@
+package reverse;
```
eric-martin, line 1: The package name isn't correct here

sk1418: Fixed! Thank you Eric.

PR #12262, opened 2022-05-26
JAVA-9556 added try-with-resources for resultset

keerthigadde:
JAVA-9556 added try-with-resources for resultset


lor6:
retest this please

PR #12267, opened 2022-05-26
BAEL-4469: Update Introduction to Mockito AdditionalAnswers

azhwani:
Simple PR

### DIFF
--- a/testing-modules/mockito-2/src/test/java/com/baeldung/mockito/additionalanswers/BookServiceUnitTest.java
+++ b/testing-modules/mockito-2/src/test/java/com/baeldung/mockito/additionalanswers/BookServiceUnitTest.java
@@ -7,6 +7,8 @@ import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.AdditionalAnswers.answer;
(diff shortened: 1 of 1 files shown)


testing-modules/mockito-2/src/test/java/com/baeldung/mockito/additionalanswers/BookServiceUnitTest.java
```diff
@@ -65,4 +72,27 @@ public void givenCheckifEqualsMethodMocked_whenCheckifEqualsInvoked_ThenReturnAr
...
+    }
+    
+    @Test
+    public void givenMockedMethod_whenMethodInvoked_thenReturnVoid() {
+        Long id = 2L;
+        when(bookRepository.getByBookId(anyLong())).thenAnswer(AdditionalAnswers.answerVoid(new VoidAnswer1<Long>() {
+            public void answer(Long id) {
+                System.out.println(id);
+            }
+        }));
```
ashleyfrieze, line 92: Hi - I created this feature in Mockito, and I also am an editor at Baeldung.

It's not meant to be used with anonymous inner classes. It was there to make it easy to plug a lambda into mockito.

```java
when(...)
   .thenAnswer(answerVoid((Long id) -> { 
      System.out.println(id);
   });
```

Note also the use of static imports to make the code readable.

It's worth noting that this may get a bit harder for the compiler, so it's sometimes a good idea to create a static method inside the test class with the right signature and then refer to it with a method reference.

PR #12297, opened 2022-06-01
Test if 64bit or not 

justducanhh:
Create pull request for test if it is 64 bit or not


lor6:
Hi @justducanhh We already have a test for this here: https://github.com/eugenp/tutorials/blob/master/java-native/src/test/java/com/baeldung/jvmbitversion/JVMBitVersionUnitTest.java so I don't think the one you added is needed.

Thanks.

justducanhh:
Thanks , this one just a small assignment (pull request for any repo > 500
star) that my teacher gave me.

PR #12323, opened 2022-06-07
BAEL-5404 - UUID as Entity ID in MongoDB

sebx59:
initial commit


sebx59:
cancelled, moving code to another project

PR #12405, opened 2022-06-23
Java 11501

panos-kakos:


akka-modules/pom.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>akka-modules</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
```
dkapil, line 7: no need to define version if the parent version is the same

PR #12417, opened 2022-06-26
JAVA-12730: Rename java-* modules inside core-java-modules

freelansam:


core-java-modules/pom.xml
```diff
@@ -32,13 +32,16 @@
...
+        <module>core-java-collections-conversions</module>
+        <module>core-java-collections-conversions-2</module>        
         <module>core-java-collections-list</module>
         <module>core-java-collections-list-2</module>
         <module>core-java-collections-list-3</module>
         <module>core-java-collections-list-4</module>
         <module>core-java-collections-maps</module>
         <module>core-java-collections-maps-2</module>
         <module>core-java-collections-maps-3</module>
+        <module>core-java-collections-maps-5</module>
```
kwoyke, line 44: Don't we need maps-4 here as well?

kwoyke: Alright, I realized that maps-4 is jdk9 module :)

PR #12419, opened 2022-06-27
JAVA-11500 : move quarkus related modules into quarkus-modules folder

keerthigadde:
move quarkus related modules into quarkus-modules folder


dkapil:
maven-wrapper.jar got deleted, it should be moved properly.
If git is not allowing you to add then you use the -f flag

PR #12425, opened 2022-06-29
Updated README.md

Asjad-J:
edited link back to article from https://www.baeldung.com/expose-graphql-field-with-different-name to https://www.baeldung.com/graphql-field-name


Asjad-J:
test this please

PR #12449, opened 2022-07-05
BAEL-5630: Spring Boot 3 Sample

ueberfuhr:
I included a Spring Boot 3 sample project. For this, there was a need to add a Spring Boot 3 parent pom. Unfortunately, the tests do not run, because PMD does not allow Java 17 (which is the minimum version for Spring Boot 3.


ueberfuhr:
Works fine now - incl. Spring Boot 3 Parent POM overwriting Maven PMD Plugin Version to support Java 17. (discussion about solution in JIRA)

PR #12464, opened 2022-07-07
BAEL-5545: prepared a simple demo to benchmark spring jdbc batch inserts

jrelmar:


JavierPeris (commented):
We should separate different package imports. There are a few specific comments to understand better this general comment.

jrelmar:
@JavierPeris Hi, I see there are some unresolved conversations left. Do I need to do further adjustments?

PR #12465, opened 2022-07-08
Replace old java-faker with kotlin-faker

serpro69:
Was going through some of your code and saw that you're using an unmaintained and buggy java-faker, so decided to replace it with kotlin-faker. Hope this is a good change for you :) And thanks for your awesome tutorials :)

There's also a reference to java-faker in https://github.com/eugenp/tutorials/tree/master/testing-modules/mocks , which also has an article linked https://www.baeldung.com/java-faker. So I haven't replaced that since it's an article about java-faker itself. But maybe you'll be interested in updating that as well to reference kotlin-faker instead? Since java-faker has not been maintained for awhile now and also contains bugs and CVEs.

### DIFF
--- a/spring-boot-modules/spring-boot-react/pom.xml
+++ b/spring-boot-modules/spring-boot-react/pom.xml
@@ -40,9 +40,9 @@
             <scope>runtime</scope>
         </dependency>
         <dependency>
-            <groupId>com.github.javafaker</groupId>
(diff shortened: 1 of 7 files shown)


lor6:
Thanks @serpro69 We're going to keep java-faker for now, but your library looks interesting. We'll add a topic about it specifically to our to-do list.

PR #12514, opened 2022-07-19
Add Namespace enum

doljae:
## Description
- add `Namespace` enum in [RFC4122 docs](https://datatracker.ietf.org/doc/html/rfc4122)
  - a predefined namespace to the document as an enum
  - It is written so that the namespace value can be used even in the UUID used as a standard class in [other programming languages.](https://docs.python.org/3/library/uuid.html)
- use Random object as a class property

### DIFF
--- a/core-java-modules/core-java-uuid/src/main/java/com/baeldung/uuid/UUIDGenerator.java
+++ b/core-java-modules/core-java-uuid/src/main/java/com/baeldung/uuid/UUIDGenerator.java
@@ -16,6 +16,24 @@ public final class UUIDGenerator {
 
     private UUIDGenerator() {}
 
+    // ref: https://datatracker.ietf.org/doc/html/rfc4122#appendix-C
+    public enum Namespace {
+        DNS("6ba7b810-9dad-11d1-80b4-00c04fd430c8"),
+        URL("6ba7b811-9dad-11d1-80b4-00c04fd430c8"),
+        OID("6ba7b812-9dad-11d1-80b4-00c04fd430c8"),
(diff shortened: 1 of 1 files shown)


core-java-modules/core-java-uuid/src/main/java/com/baeldung/uuid/UUIDGenerator.java
```diff
@@ -12,9 +12,28 @@
...
     private UUIDGenerator() {}
 
+    // ref: https://datatracker.ietf.org/doc/html/rfc4122#appendix-C
+    public enum Namespace {
+        DNS("6ba7b810-9dad-11d1-80b4-00c04fd430c8"),
+        URL("6ba7b811-9dad-11d1-80b4-00c04fd430c8"),
+        OID("6ba7b812-9dad-11d1-80b4-00c04fd430c8"),
+        X500("6ba7b814-9dad-11d1-80b4-00c04fd430c8");
+
+        public final String value;
```
doljae, line 26: I use `public final` because this `value` can be used in `generateType5UUID(String namespace, String name)` method in this class.

lor6:
Thanks @doljae I'll close the PR because the Namespace is not strictly needed here and we prefer to keep the code simple. We'll look into the other improvements you included.

doljae:
@lor6 ,
Thanks for the comment

> We'll look into the other improvements you included.

So you mean that you need to think about whether or not to reflect on this 67c82d7f76560cfdd82d103015d96f64aa182b69 commit?

PR #12528, opened 2022-07-25
BAEL-5682 - Atomic vs Volatile

eukovko:
An example for the article about the usage of atomic values and the volatile keyword.


core-java-modules/core-java-concurrency-advanced/src/main/java/com/baeldung/concurrent/atomicvsvolatile/SafeAtomicCounter.java
```diff
@@ -0,0 +1,16 @@
+package com.baeldung.concurrent.atomicvsvolatile;
```
andrzejsydor, line 1: maybe for better readibility:
```suggestion
package com.baeldung.concurrent.atomic_vs_volatile;
```

andrzejsydor (approved):

PR #12584, opened 2022-08-10
BAEL-5657 - Reading JSON from a URL in Java

ulisseslima:
Draft: https://drafts.baeldung.com/wp-admin/post.php?post=141246&action=edit


jackson-modules/jackson-conversions-2/pom.xml
```diff
@@ -44,4 +56,9 @@
         </resources>
     </build>
 
+    <properties>
+        <mockserver-netty.version>5.13.2</mockserver-netty.version>
+		<json.version>20220320</json.version>
```
KevinGilmore, line 61: use spaces (not tabs) to indent

PR #12608, opened 2022-08-17
[BAEL-4500] Add locked ownable synchronizers example code

kpentaris:
Title: What is “Locked ownable synchronizers” in thread dump?

JIRA ticket: https://jira.baeldung.com/browse/BAEL-4500

Draft link: https://drafts.baeldung.com/wp-admin/post.php?post=141658&action=edit

### DIFF
--- a/core-java-modules/core-java-concurrency-advanced-4/src/main/java/com/baeldung/ownablesynchronizers/Application.java
+++ b/core-java-modules/core-java-concurrency-advanced-4/src/main/java/com/baeldung/ownablesynchronizers/Application.java
@@ -4,58 +4,37 @@ import java.util.concurrent.locks.ReentrantLock;
 
 public class Application {
 
-    static ReentrantLock firstLock = new ReentrantLock(true);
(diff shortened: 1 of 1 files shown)


core-java-modules/core-java-concurrency-advanced-4/src/main/java/com/baeldung/ownablesynchronizers/Application.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.concurrent.locks.ReentrantLock;
+
+public class Application {
+
+    static ReentrantLock firstLock = new ReentrantLock(true);
+    static ReentrantLock secondLock = new ReentrantLock(true);
+
+    public static void main(String[] args) throws Exception {
+        // Create first thread that will lock on 'firstLock'
```
JavierPeris, line 11: This and the below comment about when the thread will be locked is not adding much value. Could you remove it or modify the code to express better your purpose??

PR #12612, opened 2022-08-17
JAVA-13855 Create new microservices-modules

anuragkumawat:


dkapil:
micronaut/.mvn/wrapper/maven-wrapper.jar is removed, let's add it inside microservices-modules/micronaut/.mvn/wrapper/maven-wrapper.jar

PR #12633, opened 2022-08-21
BAEL-5390: Compare Characters in Java

azhwani:
Simple PR


core-java-modules/core-java-string-operations-4/src/test/java/com/baeldung/charcomparison/CharComparisonUnitTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.Objects;
+
+import org.junit.jupiter.api.Test;
+
+public class CharComparisonUnitTest {
+
+    @Test
+    public void givenTwoPrimitiveChars_whenUsingRelationalOperators_thenCompare() {
+
```
JavierPeris, line 14: The code formatting is OK, however, there´s no need for the first line to be blank. Likewise, the last line of the method should not be a blank one.

JavierPeris (commented):
Since we are using JUnit 5, there´s no need to declare class and methods as public.

core-java-modules/core-java-string-operations-4/src/test/java/com/baeldung/charcomparison/CharComparisonUnitTest.java
```diff
@@ -0,0 +1,65 @@
...
+        assertTrue(Character.compare('Y', 'z') < 0);
+
+    }
+
+    @Test
+    public void givenTwoCharObjects_whenUsingCompareToMethod_thenCompare() {
+
+        Character chK = Character.valueOf('K');
+        Character chG = Character.valueOf('G');
+        Character chH = Character.valueOf('H');
```
JavierPeris, line 35: The boxing is not really needed, you can just do `Character chK = 'K';`

core-java-modules/core-java-string-operations-4/src/test/java/com/baeldung/charcomparison/CharComparisonUnitTest.java
azhwani: done

core-java-modules/core-java-string-operations-4/src/test/java/com/baeldung/charcomparison/CharComparisonUnitTest.java
azhwani: Yes I know, the idea here is to mention in the article that the constructor of the Character class is deprecated since Java 9
However, if you think that the information is irrelevant, please let me know ;)

JavierPeris: It´s OK to mention that info. With just one test we can show that case and having the rest with a direct initialization.

azhwani: Done

JavierPeris (approved):

PR #12646, opened 2022-08-24
List all factors of a number in Java

sk1418:


core-java-modules/core-java-numbers-5/src/main/java/com/baeldung/factors/FactorsOfInteger.java
```diff
@@ -0,0 +1,39 @@
+package com.baeldung.factors;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class FactorsOfInteger {
+    public static Set<Integer> getAllFactorsVer1(int n) {
+        Set<Integer> factors = new HashSet<>();
```
andrzejsydor, line 8: It could be with the initial size.
```suggestion
        Set<Integer> factors = new HashSet<>(n);
```

andrzejsydor (approved):

core-java-modules/core-java-numbers-5/src/main/java/com/baeldung/factors/FactorsOfInteger.java
sk1418: Thank you @andrzejsydor for the comment and the suggestion. Indeed, it's a good practice to set a reasonable init-capacity when we init a set.
However, I would leave the Set init code as it is, because:
- the suggested `new HashSet<>(n)`, n could be too large. Think about the number 100k has only 36 factors.
- If we really want to assign a possible reasonable init capacity, we should check the number `n`. 
    - if n is odd, the init-capacity could be `sqrt(abs(n))`
    - if n is even, the init-capacity could be `2*sqrt(abs(n))`
- However, the code here is to support an article. The article's focus is on the factor determination part. If we introduced the init-capacity, we need to explain much Set internal stuff, such as Set's default size (16? if I remember correctly), loadfactor, Set's resize mechanism.. and so on, it may distract readers.
Again, your comment is thoughtful, but if we consider the scope of the article, I would keep the Set initialization line as simple as possible. what do you think?

andrzejsydor: has sense, ok, thank you for the explanation

PR #12686, opened 2022-09-03
JAVA-13939 Fixed pom parsing error

dkapil:


lor6:
It seems that this fix is already in the main pom, so I'll close the PR here @dkapil

PR #12689, opened 2022-09-03
bug fix to avoid fallthrough

mariuslarsen:
Fixing bug where any value other than `3` will result in "unknown" being assigned to `output`.

https://www.baeldung.com/vavr-pattern-matching should also be updated with the same changes.


lor6:
Thanks @mariuslarsen Good catch. I've merged the PR and updated the article.

PR #12711, opened 2022-09-11
[JAVA-13966] Upgraded h2 database to 2.1.214 + clean up

panos-kakos:


web-modules/ninja/pom.xml
```diff
@@ -200,7 +200,7 @@
         <jetty.version>9.4.18.v20190429</jetty.version>
         <bootstrap.version>3.3.4</bootstrap.version>
         <jquery.version>2.1.3</jquery.version>
-        <h2.version>1.4.186</h2.version>
+        <h2.version>2.1.214</h2.version>
```
dkapil, line 203: can we make web-modules parent of this ninja module so that it can inherit the H2 version directly from the parent?

dkapil (changes requested):

PR #12740, opened 2022-09-18
BAEL-5717 - Fix spring-oauth2-resource-server Autoconfiguration failure

christian-german:


JavierPeris (commented):
There's only a quick thing. If you have a look at the static imports, you have moved them to the top in LogbackTestLogLevelIntegrationTest, however they remain in the same position in the other two files. So could you please apply the same rules to the 3 files?

christian-german:
Hi Luis,
I've updated all test files from the spring-boot-testing2 module according to your comment.

Cheers,
Christian.

JavierPeris (approved):

PR #12746, opened 2022-09-20
Enable SSL in Spring Boot with MySQL

saikatcse03:
Enable SSL in Spring Boot with MySQL


persistence-modules/spring-boot-mysql/src/test/java/com/baeldung/boot/jpa/EmployeeControllerUnitTest.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Test
+    public void givenEmployee_whenCreateEmployeeCalled_ThenReturnEmployee() throws Exception {
+        Employee employeeExpected = new Employee();
+        employeeExpected.setEmpName("Test Emp");
+        employeeExpected.setEmpDoj(LocalDate.now());
+        employeeExpected.setJobTitle("Manager");
+
+        Mockito.when(employeeRepository.save(employeeExpected)).thenReturn(employeeExpected);
```
ashleyfrieze, line 61: Let's static import `Mockito` here rather than qualify calls to `when` with the class name.

saikatcse03: Updated to static import for Mockito

PR #12803, opened 2022-10-01
pom.xml - update streamex dependency so MergeMaps compiles on java 17

press0:
update streamex dependency


lor6:
Thanks @press0 I've merged the PR.

PR #12810, opened 2022-10-03
BAEL-4038: Understanding java.lang.Thread.State: WAITING (parking)

kpentaris:
JIRA link: https://jira.baeldung.com/browse/BAEL-4038

### DIFF
--- a/core-java-modules/core-java-concurrency-advanced-4/src/main/java/com/baeldung/ownablesynchronizers/Application.java
+++ b/core-java-modules/core-java-concurrency-advanced-4/src/main/java/com/baeldung/ownablesynchronizers/Application.java
@@ -4,58 +4,37 @@ import java.util.concurrent.locks.ReentrantLock;
 
 public class Application {
 
-    static ReentrantLock firstLock = new ReentrantLock(true);
-    static ReentrantLock secondLock = new ReentrantLock(true);
+    public static Thread createThread(ReentrantLock mainLock, ReentrantLock otherLock, String threadName) {
+        return new Thread(() -> {
+            try {
+                mainLock.lock();
+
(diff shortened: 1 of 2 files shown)


core-java-modules/core-java-concurrency-advanced-4/src/main/java/com/baeldung/threadparking/Application.java
```diff
@@ -0,0 +1,25 @@
+package com.baeldung.threadparking;
+
+import java.util.concurrent.locks.LockSupport;
+
+public class Application {
+
+    public static void main(String[] args) throws InterruptedException {
```
JavierPeris, line 7: I think the same way Object.wait() can throw the InterruptedException, the park method cannot. Could you review it?

core-java-modules/core-java-concurrency-advanced-4/src/main/java/com/baeldung/ownablesynchronizers/Application.java
```diff
@@ -0,0 +1,45 @@
+package com.baeldung.ownablesynchronizers;
```
JavierPeris, line 1: We should use this code in the article or otherwise we should remove the file.

core-java-modules/core-java-concurrency-advanced-4/src/main/java/com/baeldung/threadparking/Application.java
kpentaris: The reason the exception is there is that it is mandatory for the `Thread.sleep()` method which is however, commented out. In the tutorial, where the reader will go through step by step, at some point the `main()` method will have declare the thrown exception.

PR #12819, opened 2022-10-04
BAEL-5779: assertAll vs Multiple Assertions in JUnit5

Mahn00rMalik:
BAEL-5779: assertAll vs Multiple Assertions in JUnit5


testing-modules/junit-5/src/main/java/groupedvsmultipleassertions/User.java
```diff
@@ -0,0 +1,37 @@
+package groupedvsmultipleassertions;
```
ashleyfrieze, line 1: Please can we have this in the `com.baeldung.groupedvsmultipleassertions` package?

testing-modules/junit-5/src/main/java/groupedvsmultipleassertions/User.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class User {
+    String username;
+    String email;
+    Boolean activated;
```
ashleyfrieze, line 6: These fields should be `private` and I'd suggest `boolean` rather than `Boolean`

testing-modules/junit-5/src/test/java/com/baeldung/groupedvsmultipleassertions/AssertAllUnitTest.java
```diff
@@ -0,0 +1,19 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class AssertAllUnitTest {
+    @Test
+    void givenAssertAll_whenSingleAssertStatementFails_thenRestWillStillBeExecuted() {
+        User user = new User("baeldung", "admin@baeldung.com", true);
+        assertAll(
+                "Exploring assertAll behavior",
```
ashleyfrieze, line 13: This message might be better as a description of what we're asserting.

testing-modules/junit-5/src/test/java/com/baeldung/groupedvsmultipleassertions/AssertAllUnitTest.java
```diff
@@ -0,0 +1,19 @@
...
+                () -> assertEquals("baeldung", user.getUsername(), "Username should be baeldung"),
+                () -> assertEquals("admin@baeldung.com", user.getEmail(), "Email should be admin@baeldung.com"),
+                () -> assertTrue(user.getActivated(), "User should be activated")
+        );
+    }
```
ashleyfrieze, line 18: Can we see a failing test> We can `@Disabled` it so it won't break the test pack.

testing-modules/junit-5/src/test/java/com/baeldung/groupedvsmultipleassertions/AssertAllUnitTest.java
```diff
@@ -0,0 +1,19 @@
+package com.baeldung.groupedvsmultipleassertions;
+
+import groupedvsmultipleassertions.User;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class AssertAllUnitTest {
```
ashleyfrieze, line 8: JUnit5 tests don't need to be `public`

testing-modules/junit-5/src/test/java/com/baeldung/groupedvsmultipleassertions/MultipleAssertionsUnitTest.java
```diff
@@ -0,0 +1,16 @@
+package com.baeldung.groupedvsmultipleassertions;
+
+import groupedvsmultipleassertions.User;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class MultipleAssertionsUnitTest {
+    @Test
```
ashleyfrieze, line 9: Why don't we put both versions of the unit test in the same test fixture, for ease of use? Also, let's show the failing test here for at least one use case - e.g. the first field failing... and maybe all three fields.

ashleyfrieze (changes requested):

PR #12827, opened 2022-10-06
BAEL-5670: Remove All Non Numeric Characters but Keep the Decimal Separator in Java String

rehammuzzamil08:


core-java-modules/pom.xml
```diff
@@ -129,6 +129,7 @@
         <module>core-java-regex-2</module>
         <module>core-java-uuid</module>
         <module>pre-jpms</module>
+        <module>core-java-string-apis-2</module>
```
KevinGilmore, line 132: Note that the modules are arranged alphabetically, so we should move this one up, after the core-java-string-apis module.

rehammuzzamil: Updated @KevinGilmore .
Thanks

rehammuzzamil08: Updated @KevinGilmore .
Thanks!

PR #12880, opened 2022-10-16
BAEL-5726: Class.forName() and Class.forName().newInstance

Ritzy-NCode:
user:priyaritu89

### DIFF
--- a/core-java-modules/core-java-8-2/src/test/java/com/baeldung/javaClass/ClassForNameAndInstanceUnitTest.java
+++ b/core-java-modules/core-java-8-2/src/test/java/com/baeldung/javaClass/ClassForNameAndInstanceUnitTest.java
@@ -1,0 +1,19 @@
+package com.baeldung.javaClass;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+public class ClassForNameAndInstanceUnitTest {
+    ClassForNameAndInstance tester= new ClassForNameAndInstance();
+    @Test
+    public void givenString_whenValidClass_thenReturnClass() throws ClassNotFoundException {
+        Class c= tester.createClassForName("java.lang.String");
+        assertEquals(c.getName(),"java.lang.String");
+    }
+    @Test
+    public void givenString_whenValidClass_thenReturnClassObject() throws ClassNotFoundException, InstantiationException, IllegalAccessException {
+        Object o= tester.createNewInstance("java.lang.String");
+        assertEquals(o.getClass().getName(),"java.lang.String");
+    }
+}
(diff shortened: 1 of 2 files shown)


core-java-modules/core-java-8-2/src/test/java/com/baeldung/javaClass/ClassForNameAndInstanceUnitTest.java
```diff
@@ -0,0 +1,19 @@
+package com.baeldung.javaClass;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
```
ashleyfrieze, line 5: Shall we use JUnit 5 instead?

core-java-modules/core-java-8-2/src/test/java/com/baeldung/javaClass/ClassForNameAndInstanceUnitTest.java
```diff
@@ -0,0 +1,19 @@
...
+
+public class ClassForNameAndInstanceUnitTest {
+    ClassForNameAndInstance tester= new ClassForNameAndInstance();
```
ashleyfrieze, line 8: Let's not `new` the `ClassForNameAndInstance` class as described.

core-java-modules/core-java-8-2/src/test/java/com/baeldung/javaClass/ClassForNameAndInstanceUnitTest.java
```diff
@@ -0,0 +1,19 @@
...
+    @Test
+    public void givenString_whenValidClass_thenReturnClass() throws ClassNotFoundException {
```
ashleyfrieze, line 10: Tests don't need to throw specific exceptions just `throws Exception` would be fine. This will make the second test a lot cleaner to read too.

core-java-modules/core-java-8-2/src/test/java/com/baeldung/javaClass/ClassForNameAndInstanceUnitTest.java
```diff
@@ -0,0 +1,19 @@
...
+        Class c= tester.createClassForName("java.lang.String");
+        assertEquals(c.getName(),"java.lang.String");
```
ashleyfrieze, line 12: This assertion is backwards. We should put the expected value at the beginning.

It might be better if we did `assertEquals(String.class, c)` and maybe even call the variable `stringClass`

core-java-modules/core-java-8-2/src/test/java/com/baeldung/javaClass/ClassForNameAndInstanceUnitTest.java
```diff
@@ -0,0 +1,19 @@
...
+    }
+    @Test
+    public void givenString_whenValidClass_thenReturnClassObject() throws ClassNotFoundException, InstantiationException, IllegalAccessException {
+        Object o= tester.createNewInstance("java.lang.String");
```
ashleyfrieze, line 16: Creating a new instance of `String` isn't very interesting. What if, instead, we created a custom class here in the example code and tried to use its constructor. We could even give it some default field values so we can assert that it got created properly.

ashleyfrieze (changes requested):

PR #12935, opened 2022-10-26
BAEL-5791: Convert String to Instant (creating new module)

achraftt:


core-java-modules/core-java-datetime-string-2/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>${commons-lang3.version}</version>
+        </dependency>
+        <!-- https://mvnrepository.com/artifact/commons-validator/commons-validator -->
+        <dependency>
+            <groupId>commons-validator</groupId>
+            <artifactId>commons-validator</artifactId>
+            <version>${commons-validator.version}</version>
+        </dependency>
```
KevinGilmore, line 28: These dependencies aren't used in your code. We don't need to add them unless/until they are needed in the module.

core-java-modules/core-java-datetime-string-2/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+    </dependencies>
+
+    <build>
+        <finalName>core-java-datetime-string-2</finalName>
+        <resources>
+            <resource>
+                <directory>src/main/resources</directory>
+                <filtering>true</filtering>
+            </resource>
+        </resources>
```
KevinGilmore, line 38: We can omit these lines.

core-java-modules/core-java-datetime-string-2/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+                <configuration>
+                    <source>${maven.compiler.source}</source>
+                    <target>${maven.compiler.target}</target>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+    <properties>
+        <commons-validator.version>1.6</commons-validator.version>
```
KevinGilmore, line 53: Since we don't need the dependency, we don't need this, either.

core-java-modules/core-java-datetime-string-2/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+        <maven.compiler.source>1.9</maven.compiler.source>
+        <maven.compiler.target>1.9</maven.compiler.target>
```
KevinGilmore, line 55: Any reason to target 1.9 here? Let's stick to 1.8 unless we have something that's specific to Java 9.

KevinGilmore (changes requested):

core-java-modules/core-java-datetime-string-2/pom.xml
achraftt: Done.

core-java-modules/core-java-datetime-string-2/pom.xml
achraftt: Done.

core-java-modules/core-java-datetime-string-2/pom.xml
achraftt: Done.

PR #12968, opened 2022-11-01
Feing client vs web client

VladFernoaga:


spring-reactive-modules/spring-5-reactive-3/src/main/java/com/baeldung/webclient/ProductsSlowServiceController.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.List;
+
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class ProductsSlowServiceController {
+
+    @GetMapping("/slow-service-products")
+    private List<Product> getAllProducts() throws Exception {
```
KevinGilmore, line 13: Let's have it throw InterruptedException instead since that's the only non-RuntimeException that it can throw.

PR #12988, opened 2022-11-06
Add Type Parameter vs Wildcard in Java Generics

apeterlic:


core-java-modules/core-java-lang-5/src/main/java/com/baeldung/generics/AnimalDemo.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class AnimalDemo {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AnimalDemo.class);
```
JavierPeris, line 11: Although this variable has the modifier `static final`, it is not a constant and should not appear in uppercase. In this case, we have a final reference so it should be in lowercase.

core-java-modules/core-java-lang-5/src/main/java/com/baeldung/generics/Cat.java
```diff
@@ -0,0 +1,32 @@
...
+    @Override
+    public String makeSound() {
+        return "Meow";
+    }
+
+    @Override
+    public int compareTo(Cat cat) {
+        return this.getName()
+            .length() - cat.getName()
+            .length();
```
JavierPeris, line 17: In this case, we can put together, in the same line the call to getName() and length(). In fact, it can be written in just one line and it would easier to read it.

core-java-modules/core-java-lang-5/src/main/java/com/baeldung/generics/Cat.java
```diff
@@ -0,0 +1,30 @@
...
+    public String toString() {
+        final StringBuilder sb = new StringBuilder("Cat{");
+        sb.append("type='")
+            .append(type)
+            .append('\'');
+        sb.append(", name='")
+            .append(name)
+            .append('\'');
+        sb.append('}');
+        return sb.toString();
```
JavierPeris, line 28: Since the string can be done in on line we are not gaining performance for using StringBuilder so I think just appending the literals and variables would make the code easier to read.

`return "Cat{" + "type='" + type + '\'' + ", name='" + name + '\'' + '}';`

core-java-modules/core-java-lang-5/src/main/java/com/baeldung/generics/Cat.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Override
+    public String makeSound() {
+        return "Meow";
+    }
+
+    @Override
+    public int compareTo(Cat cat) {
+        return this.getName().length() - cat.getName().length();
+    }
```
JavierPeris, line 16: Since we are overriding the compareTo method we should also override the equals method making two animals that have the same name length equal. In this way we would comply with the recommendation:
`x.compareTo(y)==0) == (x.equals(y)`

apeterlic: Thank you @JavierPeris, I've updated the code.

JavierPeris: Hi @apeterlic the equals itself doesn't have an impact on the article. However, please pay attention to the rule:
x.compareTo(y)==0) == (x.equals(y) whenever compareTo returns a 0, the equals have to return true with your definition of comparing an equals this condition is not met. It's OK as long as the consumer of the code is warned due to the impact it can have.

apeterlic: Hi @JavierPeris I understand. Would it be ok if we just add comment? Otherwise, I'd need to modify compareTo method.

JavierPeris: @apeterlic IIt's OK with a comment.

apeterlic: @JavierPeris, thanks. I added a comment.

JavierPeris (approved):

PR #12994, opened 2022-11-08
[JAVA-13960] Clean up

panos-kakos:


core-java-modules/core-java-concurrency-simple/src/test/java/com/baeldung/concurrent/synchronize/SychronizedBlockUnitTest.java
```diff
@@ -9,15 +9,15 @@
 
 import static org.junit.Assert.assertEquals;
 
-public class BaeldungSychronizedBlockUnitTest {
+public class SychronizedBlockUnitTest {
```
dkapil, line 12: Let's rename this to `SynchronizedBlocksUnitTest` just to be consistent with the original class name

panos-kakos: Both Classes renamed. Pr updated.

PR #12997, opened 2022-11-08
fix getLogger class name

press0:
fix getLogger class name


lor6:
Thanks @press0 I've merged the PR.

PR #12998, opened 2022-11-08
https://jira.baeldung.com/browse/BAEL-4630

ciphx:
Change Description : 
1. Added TestNG tests in testng-jenkins folder
2. Added XML-based configs to run the tests
3. Added a pom.xml with the surefire plugin to use the XML file during the test phase of the build.


ciphx:
not needed, will point to existing testNg code

PR #13047, opened 2022-11-18
BAEL-5951: Spring Boot 3 Sample for Native Image builds incl. Runtime Hints

ueberfuhr:
- added a module for native image builds with Spring Boot 3
 - extended parent pom for Spring Boot 3 by plugin configuration and profiles (copied from original Spring Boot 3 parent POM)


ueberfuhr:
Still in progress. because there are 3 issues to solve:
 - Swagger UI should work in Native Mode
 - GraphiQL brings error when loading scheme from server
 - Main-Class is not running in Native Mode when using Baeldung Parent POM instead of original Spring Boot POM

ueberfuhr:
FYI: There are dependencies in the `parent-modules` POM that are directly inherited in all projects but break with Spring Boot 3 because of major version changes. (Logback, SLF4J) I'm not sure, why those dependencies must be inherited into each project. I could overwrite the versions in the custom Spring Boot 3 parent POM, but it would not be possible to exclude those dependencies from the project, although not necessary.

PR #13061, opened 2022-11-19
BAEL-5978: Case Insensitive Enum Mapping in Spring Boot

azhwani:
Simple PR


spring-boot-modules/spring-boot-basic-customization-2/src/main/java/com/baeldung/enummapping/config/EnumMappingConfig.java
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.format.FormatterRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+
+@Configuration
+public class EnumMappingConfig implements WebMvcConfigurer {
+    @Override
+    public void addFormatters(FormatterRegistry registry) {
+        // ApplicationConversionService.configure(registry);
+        // registry.addConverter(new StringToLevelConverter());
```
JavierPeris, line 12: Why do we have these lines commented?

spring-boot-modules/spring-boot-basic-customization-2/src/main/java/com/baeldung/enummapping/enums/Level.java
```diff
@@ -0,0 +1,7 @@
+package com.baeldung.enummapping.enums;
+
+public enum Level {
+
+    LOW, MEDIUM, HIGH;
```
JavierPeris, line 5: We can remove the semicolon is not necessary.

spring-boot-modules/spring-boot-basic-customization-2/src/main/java/com/baeldung/enummapping/editors/LevelEditor.java
```diff
@@ -0,0 +1,16 @@
...
+
+import org.apache.commons.lang3.EnumUtils;
+
+import com.baeldung.enummapping.enums.Level;
+
+public class LevelEditor extends PropertyEditorSupport {
+
+    @Override
+    public void setAsText(String text) {
+        setValue(EnumUtils.getEnum(Level.class, text.toUpperCase()));
```
JavierPeris, line 13: Shouldn't we check the string is neither null nor empty?

JavierPeris (commented):
We should create a new module since this has already 7 articles which is the maximum for a Spring module.

spring-boot-modules/spring-boot-basic-customization-2/src/main/java/com/baeldung/enummapping/enums/Level.java
azhwani: done

spring-boot-modules/spring-boot-basic-customization-2/src/main/java/com/baeldung/enummapping/editors/LevelEditor.java
azhwani: Done

Normally, we used _required=false_, so when the parameter isn't specified, the method parameter is bound to null (No converters or property editors are called).

However, I agree. We need to protect our custom converter and property editor against any misuse or NPE

I added additional test cases to take into account _null_ and empty strings

JavierPeris (approved):

PR #13070, opened 2022-11-21
 JAVA-14897 Update spring-security-auth0 module under spring-security modules to remove usage of deprecated WebSecurityConfigurerAdapter

anuragkumawat:


spring-security-modules/spring-security-auth0/src/main/java/com/baeldung/auth0/AuthConfig.java
```diff
@@ -53,18 +52,23 @@ public AuthenticationController authenticationController() throws UnsupportedEnc
...
-        .and()
-        .formLogin()
-        .loginPage("/login")
-        .and()
-        .logout().logoutSuccessHandler(logoutSuccessHandler()).permitAll();
+    @Bean
+    public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
+        http.csrf()
+            .disable();
+        http.authorizeRequests()
```
dkapil, line 59: can these two statements be combined?

dkapil (changes requested):

spring-security-modules/spring-security-auth0/src/main/java/com/baeldung/auth0/AuthConfig.java
anuragkumawat: Yes , I have updated the file.

PR #13105, opened 2022-11-29
[BAEL-5976]: Extract Values using AssertJ in Java

cchacin:
JIRA: https://jira.baeldung.com/browse/BAEL-5976

### DIFF
--- a/testing-modules/assertion-libraries/src/main/java/com/baeldung/assertj/extracting/Address.java
+++ b/testing-modules/assertion-libraries/src/main/java/com/baeldung/assertj/extracting/Address.java
@@ -1,0 +1,25 @@
+package com.baeldung.assertj.extracting;
+
+class Address {
+    private String street;
+    private String city;
+    private ZipCode zipCode;
+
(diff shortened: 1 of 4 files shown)


testing-modules/assertion-libraries/src/main/java/com/baeldung/assertj/extracting/Address.java
```diff
@@ -0,0 +1,25 @@
+package com.baeldung.assertj.extracting;
+
+class Address {
+    private String street;
```
andrzejsydor, line 4: Couldn't be final?

andrzejsydor (changes requested):

testing-modules/assertion-libraries/src/main/java/com/baeldung/assertj/extracting/Address.java
cchacin: yes, definitely, but keeping it concise and short for the blog post snippets

PR #13110, opened 2022-11-29
BAEL-5931: Default Capacity of ArrayList

azhwani:
Simple PR

### DIFF
--- a/core-java-modules/core-java-collections-array-list/src/test/java/com/baeldung/defaultarraylistcapacity/DefaultArrayListCapacityUnitTest.java
+++ b/core-java-modules/core-java-collections-array-list/src/test/java/com/baeldung/defaultarraylistcapacity/DefaultArrayListCapacityUnitTest.java
@@ -1,0 +1,33 @@
+package com.baeldung.defaultarraylistcapacity;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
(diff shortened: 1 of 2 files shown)


core-java-modules/core-java-collections-array-list/src/test/java/com/baeldung/defaultarraylistcapacity/DefaultArrayListCapacityUnitTest.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.ArrayList;
+
+import org.junit.jupiter.api.Test;
+
+public class DefaultArrayListCapacityUnitTest {
+
+    @Test
+    void givenEmptyArrayList_whenGetDefaultCapacity_thenReturnZero() throws IllegalArgumentException, IllegalAccessException, NoSuchFieldException, SecurityException {
+
+        ArrayList<Integer> myList = new ArrayList<Integer>();
```
JavierPeris, line 14: We don't need the explicit type argument, you can remove the Integer from the object creation. And the same thing for the test below.

azhwani: Done

PR #13154, opened 2022-12-11
BAEL-5878: Getting a Value in JSONObject. Add JsonObjectValueGetter with unit tests.

rafyen:
Hi, this is the PR for the article https://jira.baeldung.com/browse/BAEL-5878, which is about getting values in JSONObject.


rafyen:
Will open a new PR with json-value-getter moved under json-2 module

PR #13166, opened 2022-12-15
[JAVA-5730] convert list of objects to list of strings

bhaskar16:
### Convert an ArrayList<Object> to ArrayList<String>

### DIFF
--- a/core-java-modules/core-java-collections-array-list/src/main/java/com/baeldung/listofobjectstolistofstring/User.java
+++ b/core-java-modules/core-java-collections-array-list/src/main/java/com/baeldung/listofobjectstolistofstring/User.java
@@ -1,22 +1,14 @@
 package com.baeldung.listofobjectstolistofstring;
 
 public class User {
-    private final String userName;
-    private final String firstName;
-    private final String lastName;
+    private final String fullName;
 
-    public User(String userName, String firstName, String lastName) {
-        this.userName = userName;
-        this.firstName = firstName;
-        this.lastName = lastName;
+    public User(String fullName) {
+        this.fullName = fullName;
     }
 
     @Override
     public String toString() {
-        return "User{" +
-                "userName='" + userName + '\'' +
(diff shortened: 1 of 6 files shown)


core-java-modules/core-java-collections-array-list/src/main/java/com/baeldung/listofobjectstolistofstring/User.java
```diff
@@ -0,0 +1,22 @@
...
+        this.lastName = lastName;
+    }
+
+    @Override
+    public String toString() {
+        return "User{" +
+                "userName='" + userName + '\'' +
+                ", firstName='" + firstName + '\'' +
+                ", lastName='" + lastName + '\'' +
+                '}';
```
Bwvolleyball, line 20: Let's follow the pattern from the `Node` class for this `toString()`

```suggestion
        return "User (" +
                "userName='" + userName + '\'' +
                ", firstName='" + firstName + '\'' +
                ", lastName='" + lastName + '\'' +
                ')';
```

core-java-modules/core-java-collections-array-list/src/main/java/com/baeldung/listofobjectstolistofstring/User.java
```diff
@@ -0,0 +1,22 @@
+package com.baeldung.listofobjectstolistofstring;
+
+public class User {
+    private final String userName;
+    private final String firstName;
+    private final String lastName;
+
+    public User(String userName, String firstName, String lastName) {
```
Bwvolleyball, line 8: The code in the article just passes one `String` to the `User` constructor, maybe we should just simplify this class to be `User(String name)` to make it easy to align to the article?

bhaskar16: Sure, makes sense.

PR #13171, opened 2022-12-18
BAEL-6044: Fix the IllegalArgumentException: No enum const class

azhwani:
Simple PR


core-java-modules/core-java-exceptions-4/src/main/java/com/baeldung/exception/noenumconst/Priority.java
```diff
@@ -0,0 +1,17 @@
...
+
+    private String name;
+
+    Priority(String name) {
+        this.name = name;
+    }
+
+    public String getPriorityName() {
+        return name;
+    }
```
JavierPeris, line 15: It seems this method is not used.

JavierPeris (approved):

PR #13195, opened 2022-12-24
Bael 5973 - RestExpress example

abh1navv:


pom.xml
```diff
@@ -7,6 +7,7 @@
     <groupId>com.baeldung</groupId>
     <artifactId>parent-modules</artifactId>
     <version>1.0.0-SNAPSHOT</version>
+
```
KevinGilmore, line 10: Let's revert the change to the root pom.xml since it's whitespace only.

microservices-modules/rest-express/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>microservices-modules</artifactId>
+        <groupId>com.baeldung</groupId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <name>RestExpress-MongoDB-Scaffold</name>
```
KevinGilmore, line 11: Can we change the name to match the artifactId as per guidelines?

KevinGilmore (commented):
Please see notes on the two pom.xml files

PR #13246, opened 2023-01-06
Check If Command Line Arguments Are Null in Java

hajarrs:


core-java-modules/core-java-lang-5/src/test/java/com/baeldung/commandLine/CommandLineWithoutErrorHandlingUnitTest.java
```diff
@@ -0,0 +1,18 @@
+package com.baeldung.commandLine;
+
+import org.junit.Test;
+import static org.junit.Assert.fail;
+
+public class CommandLineWithoutErrorHandlingUnitTest {
+
+    @Test
```
andrzejsydor, line 8: it's junit 4, why not just expected?
```suggestion
    @Test(expected=NullPointerException.class)
```

andrzejsydor (changes requested):

PR #13251, opened 2023-01-07
Feature/bael 6083 http interface

dstr89:


spring-core-6/src/test/java/com/baeldung/httpinterface/BooksServiceMockitoTest.java
```diff
@@ -0,0 +1,102 @@
...
+    @Test
+    void givenMockedWebClient_whenAllBooksAreRequested_thenTwoBooksAreReturned() {
+        when(webClient.method(HttpMethod.GET)).thenReturn(requestBodyUri);
+        when(requestBodyUri.uri(anyString(), anyMap())).thenReturn(requestBody);
+        when(requestBody.retrieve()).thenReturn(response);
+        when(response.bodyToMono(new ParameterizedTypeReference<List<Book>>(){}))
+          .thenReturn(Mono.just(List.of(
+            new Book(1,"Book_1", "Author_1", 1998),
+            new Book(2, "Book_2", "Author_2", 1999)
+          )));
```
ashleyfrieze, line 55: A couple of things here.

Firstly, it's better to use BDDMockito than Mockito. We're really setting up the givens of this, rather than the whens. In that we're saying "given response's body to mono will return X" ready for the next step of the test to go "when system under test is asked this, then the result is that".

Secondly, I think I mentioned this when we were discussing the outline, you can use deep mocks.

https://www.baeldung.com/mockito-fluent-apis

```java
@Mock(answer = Answers.RETURNS_DEEP_STUBS)
private WebClient webClient;

    @Test
    void givenMockedWebClient_whenAllBooksAreRequested_thenTwoBooksAreReturned() {
        given(webClient.method(HttpMethod.GET)
          .uri(anyString(), anyMap())
          .retrieve()
          .bodyToMono(new ParameterizedTypeReference<List<Book>>(){}))
          .willReturn(Mono.just(List.of(
            new Book(1,"Book_1", "Author_1", 1998),
            new Book(2, "Book_2", "Author_2", 1999)
          )));

```

spring-core-6/src/test/java/com/baeldung/httpinterface/BooksServiceMockitoTest.java
```diff
@@ -0,0 +1,102 @@
...
+        when(webClient.method(HttpMethod.DELETE)).thenReturn(requestBodyUri);
+        when(requestBodyUri.uri(anyString(), anyMap())).thenReturn(requestBody);
+        when(requestBody.retrieve()).thenReturn(response);
+        when(response.toBodilessEntity()).thenReturn(monoResponseEntity);
+        when(monoResponseEntity.block(any())).thenReturn(responseEntity);
+        when(responseEntity.getStatusCode()).thenReturn(HttpStatusCode.valueOf(200));
+
+        BooksService booksService = booksClient.getBooksService();
+        ResponseEntity<Void> response = booksService.deleteBook(3);
+        assertTrue(response.getStatusCode().is2xxSuccessful());
```
ashleyfrieze, line 99: Minor... this is a fair assertion, but if it fails all you get is "expected true, was false" when if you asserted the actual status code, you'd get "expected 200, was 401" or something more useful.

ashleyfrieze (changes requested):

spring-core-6/src/test/java/com/baeldung/httpinterface/BooksServiceMockitoTest.java
dstr89: Great stuff @ashleyfrieze, I wasn't aware of this feature, will change it.

PR #13270, opened 2023-01-10
JAVA-16687. Update MapStruct version.

jsgrah-spring:


mapstruct/pom.xml
```diff
@@ -76,7 +76,7 @@
     </build>
 
     <properties>
-        <org.mapstruct.version>1.4.2.Final</org.mapstruct.version>
+        <org.mapstruct.version>1.5.3.Final</org.mapstruct.version>
```
andrzejsydor, line 79: There is any breaking changes?

jsgrah-spring: @andrzejsydor No breaking change was encountered. It compiles without any issue and All tests where green.

PR #13294, opened 2023-01-15
JAVA-15980 Upgrade AWS Kinesis

anastasiosioannidis:


spring-cloud-modules/spring-cloud-stream/spring-cloud-stream-kinesis/src/main/java/com/baeldung/kclkpl/KinesisKCLApplication.java
```diff
@@ -17,32 +18,31 @@ public class KinesisKCLApplication implements ApplicationRunner {
...
-		  .config(consumerConfig)
-		  .build()
-		  .run();
-	}
-    
+    @Override
+    public void run(ApplicationArguments args) throws Exception {
+        BasicAWSCredentials awsCredentials = new BasicAWSCredentials(accessKey, secretKey);
+
+        KinesisClientLibConfiguration consumerConfig = new KinesisClientLibConfiguration("KinesisKCLConsumer", IPS_STREAM, "", "", KinesisClientLibConfiguration.DEFAULT_INITIAL_POSITION_IN_STREAM, new AWSStaticCredentialsProvider(awsCredentials),
```
dkapil, line 36: @anastasiosioannidis please do a wild card static import for `KinesisClientLibConfiguration` so that we can directly use the constants here. It will also make the code more readable here as well as in the article.

PR #13310, opened 2023-01-18
BAEL-5770 [Parse dates with multiple formats]

tgvLt:


core-java-modules/ParsingDates/src/main/java/com/parse/parsingdates/ParsingDates.java
```diff
@@ -0,0 +1,14 @@
+/*
```
Bwvolleyball, line 1: Since we can leverage the other sub-module, we shouldn't need this file.

core-java-modules/ParsingDates/src/main/java/com/parse/parsingdates/SimpleDateTimeFormater.java
```diff
@@ -0,0 +1,28 @@
...
+                .append(DateTimeFormatter.ofPattern("[MM/dd/yyyy]" + "[dd-MM-yyyy]" + "[yyyy-MM-dd]"));
+
+        DateTimeFormatter dateTimeFormatter = dateTimeFormatterBuilder.toFormatter();
+
+        try {
+            dateTimeFormatter.parse(date);
+            return true;
+        } catch (DateTimeParseException e) {
+            return false;
+        }
```
Bwvolleyball, line 25: Does it make sense to coerce this down to a boolean, or is it better to just return the parsed date so that the test can do what it needs to do?

In tests, if we expect parsing to fail, we can use `assertThatThrownBy` to watch the function under test and catch the thrown exception.

core-java-modules/ParsingDates/src/main/java/com/parse/parsingdates/SimpleDateUtils.java
```diff
@@ -0,0 +1,23 @@
...
+ * and open the template in the editor.
+ */
+package com.parse.parsingdates;
+
+import java.text.ParseException;
+import org.apache.commons.lang3.time.DateUtils;
+
+public class SimpleDateUtils {
+
+    public boolean parseDate(String date) {
```
Bwvolleyball, line 13: Same here, I'm thinking it's better to return the parsed date, and not coerce it down to a boolean.

core-java-modules/ParsingDates/src/main/java/com/parse/parsingdates/SimpleDateTimeFormat.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package com.parse.parsingdates;
+
+import java.util.Arrays;
+import java.util.List;
+import org.joda.time.format.DateTimeFormat;
+
+public class SimpleDateTimeFormat {
+
+    public boolean parseDate(String date) {
```
Bwvolleyball, line 14: Same here, let's return the parsed date instead.

Bwvolleyball (changes requested):
Let's consolidate this source / example code into the existing submodule here: https://github.com/eugenp/tutorials/tree/master/core-java-modules/core-java-8-datetime-2

We shouldn't need a new `pom.xml` at this time, and let's not commit the `target/` folder, since those are for build artifacts 😄 

With all the date parsing classes, these could probably all be consolidated into one class, and these functions could probably be changed to static, so that they are more of utility / helper functions and not tied to an object instance. Especially since they don't depend on any state in the classes.

core-java-modules/ParsingDates/src/main/java/com/parse/parsingdates/SimpleParseDate.java
```diff
@@ -0,0 +1,76 @@
...
+            + "|^((0[1-9]|[12][0-9]|3[01])-(0[13578]|10|12)-((19|2[0-9])[0-9]{2}))$"
+            + "|^((0[1-9]|[12][0-9]|30)-(0[469]|11)-((19|2[0-9])[0-9]{2}))$");
+    private static Pattern DATE_PATTERNMMDDYYYY = Pattern.compile(
+            "^(29/02/(2000|2400|2800|(19|2[0-9])(0[48]|[2468][048]|[13579][26])))$"
+            + "|^((0[1-9]|1[0-9]|2[0-8])/02/((19|2[0-9])[0-9]{2}))$"
+            + "|^((0[1-9]|[12][0-9]|3[01])/(0[13578]|10|12)/((19|2[0-9])[0-9]{2}))$"
+            + "|^((0[1-9]|[12][0-9]|30)/(0[469]|11)/((19|2[0-9])[0-9]{2}))$");
+
+    public SimpleParseDate() {
+    }
```
Bwvolleyball, line 35: I think this is extra / not needed

core-java-modules/ParsingDates/src/main/java/com/parse/parsingdates/SimpleParseDate.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
Bwvolleyball, line 5: These header comments can safely be deleted.

Bwvolleyball (commented):
I also realize that you might need some additional dependencies if you move this over to the module I mentioned.  You can just add these to the existing `pom.xml`.

PR #13317, opened 2023-01-20
Groovy module improvements

Alex-Golub:
- Upgraded Groovy to version 3.0.8
- Align tests with the Spock testing framework
- Code reformating
- Minor fixes to tests


lor6:
Thanks @Alex-Golub 
We'll update the articles to match your changes as well.

PR #13321, opened 2023-01-21
JAVA-17420 Update The transient Keyword in Java Article

timis1:
Add a new example for transient keyword for Strings when using **new** operator.

### DIFF
--- a/core-java-modules/core-java-lang-3/src/main/java/com/baeldung/transientkw/Book.java
+++ b/core-java-modules/core-java-lang-3/src/main/java/com/baeldung/transientkw/Book.java
@@ -11,6 +11,8 @@ public class Book implements Serializable {
     private transient int copies;
     private final transient String bookCategory = "Fiction";
 
+    private final transient String bookCategoryNewOperator = new String("Fiction with new Operator");
+
     public String getBookName() {
         return bookName;
     }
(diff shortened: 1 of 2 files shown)


core-java-modules/core-java-lang-3/src/main/java/com/baeldung/transientkw/Book.java
```diff
@@ -11,6 +11,8 @@ public class Book implements Serializable {
     private transient int copies;
     private final transient String bookCategory = "Fiction";
 
+    private final transient String bookCategoryNewOperator = new String("Fiction with new Operator");
```
andrzejsydor, line 14: I'm not sure that is correct.
What is the purpose for the `final` immutable object use transient?

andrzejsydor (changes requested):

core-java-modules/core-java-lang-3/src/main/java/com/baeldung/transientkw/Book.java
timis1: Hi @andrzejsydor, 

Thank you for your feedback!
You are right that Strings are by default immutable. 
The purpose of the `transient` is to not include that field in the serialization/de serialialization proces.
Could you please tell me what is your thoughts about incorrect part?

andrzejsydor: true, that's right, sorry my mistake

andrzejsydor (approved):

PR #13325, opened 2023-01-21
Feature/bael 6101 singleton

dstr89:


spring-boot-modules/spring-boot-3/src/main/java/com/baeldung/sample/singleton/SingletonPattern.java
```diff
@@ -0,0 +1,24 @@
...
+
+    private static volatile SingletonPattern instance = null;
+
+    private SingletonPattern() {}
+
+    public static SingletonPattern getInstance() {
+        if (instance == null) {
+            synchronized(SingletonPattern.class) {
+                if (instance == null) {
+                    instance = new SingletonPattern();
```
ashleyfrieze, line 13: I'd be interested in comparing this to 

```java
public final class SingletonPattern {
   private static final SingletonPattern instance = new SingletonPattern();

   public static SingletonPattern getInstance() { return instance; }
}
```

I believe that these are, functionally, the same!!!

spring-boot-modules/spring-boot-3/src/main/java/com/baeldung/sample/singleton/SingletonPattern.java
```diff
@@ -0,0 +1,24 @@
+package com.baeldung.sample.singleton;
+
+public final class SingletonPattern {
```
ashleyfrieze, line 3: The class name - `SingletonPattern` - might be worth improving:

`ThreadSafeSingleInstance`?

ashleyfrieze (changes requested):

spring-boot-modules/spring-boot-3/src/main/java/com/baeldung/sample/singleton/SingletonPattern.java
dstr89: Hi @ashleyfrieze, functionally I would say yes. But your example is not using lazy loading, which is described in the article. I would leave it like this.

spring-boot-modules/spring-boot-3/src/main/java/com/baeldung/sample/singleton/SingletonPattern.java
dstr89: Makes sense, fixed.

andrzejsydor (approved):
What about enum as a Singleton?

PR #13332, opened 2023-01-22
[int-zero-or-null] Check if an Integer Value is null or Zero in Java

sk1418:


core-java-modules/core-java-numbers-5/src/main/java/com/baeldung/IntegerNullOrZero.java
```diff
@@ -0,0 +1,26 @@
...
+    public static boolean usingStandardWay(Integer num) {
+        return num == null || num == 0;
+    }
+
+    public static boolean usingTernaryOperator(Integer num) {
+        return 0 == (num == null ? 0 : num);
+    }
+
+    public static boolean usingOptional(Integer num) {
+        return Optional.ofNullable(num).orElse(0) == 0;
```
andrzejsydor, line 20: why you return 0 instead of throw exception?
What does it mean that for null you return 0.

andrzejsydor (changes requested):

core-java-modules/core-java-numbers-5/src/main/java/com/baeldung/IntegerNullOrZero.java
lor6: Hi @andrzejsydor I noticed you've left comments on several of the PRs opened on our repository. 
However, we have an internal process and guidelines for the code we add, and most of the conversation is on our internal JIRA. So you don't have sufficient insight to provide feedback at this point, and the comments can confuse the developers. 
Please don't add comments or request changes on in-progress PRs opened here. 
If you have feedback on the published articles, please use the comment feature or email team@baeldung.com.

Thanks.

andrzejsydor: ok, no problem.

PR #13339, opened 2023-01-23
Added models for deep vs shallow copy object creation along with unit test

balasr21:
This PR contains code examples for creating deep vs shallow copy of an object

### DIFF
--- a/core-java-modules/core-java/src/main/java/com/baeldung/clone/DeepMatch.java
+++ b/core-java-modules/core-java/src/main/java/com/baeldung/clone/DeepMatch.java
@@ -1,0 +1,18 @@
+package com.baeldung.clone;
+
+import java.time.OffsetDateTime;
+
+public class DeepMatch extends ShallowMatch {
+
+    public DeepMatch(TYPE type, OffsetDateTime startTime, OffsetDateTime endTime, Integer durationInMinutes, PlayersDetail playersDetail) {
+        super(type, startTime, endTime, durationInMinutes, playersDetail);
+    }
+
+    @Override
+    protected Object clone() throws CloneNotSupportedException {
+
+        DeepMatch deepCopyMatch = (DeepMatch) super.clone();
+        deepCopyMatch.setPlayersDetail((PlayersDetail) getPlayersDetail().clone());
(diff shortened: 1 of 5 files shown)


core-java-modules/core-java/src/main/java/com/baeldung/clone/DeepMatch.java
```diff
@@ -0,0 +1,18 @@
...
+
+public class DeepMatch extends ShallowMatch {
+
+    public DeepMatch(TYPE type, OffsetDateTime startTime, OffsetDateTime endTime, Integer durationInMinutes, PlayersDetail playersDetail) {
+        super(type, startTime, endTime, durationInMinutes, playersDetail);
+    }
+
+    @Override
+    protected Object clone() throws CloneNotSupportedException {
+
```
andrzejsydor, line 13: empty line could be deleted

core-java-modules/core-java/src/main/java/com/baeldung/clone/ShallowMatch.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private OffsetDateTime startTime;
+
+    private OffsetDateTime endTime;
+
+    private Integer durationInMinutes;
+
+    private PlayersDetail playersDetail;
+
+    public enum TYPE {
```
andrzejsydor, line 22: why not just `Type` (CamelCase)
It's the standard, I think, see: https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

andrzejsydor (changes requested):

core-java-modules/core-java/src/main/java/com/baeldung/clone/DeepMatch.java
balasr21: Thanks for your comments, done

core-java-modules/core-java/src/main/java/com/baeldung/clone/ShallowMatch.java
balasr21: Thanks for your comments, done

PR #13345, opened 2023-01-25
BAEL-5990 is implemented to generate JMeter dashboard reports

muasif80:


jmeter/pom.xml
```diff
@@ -68,5 +68,91 @@
     <properties>
         <jmeter.version>2.6.0</jmeter.version>
     </properties>
+    
+    <profiles>
+		<profile>
```
andrzejsydor, line 73: the formatting is weird, please try reformat code

jmeter/src/main/java/com/baeldung/dashboard/controllers/DashBoard.java
```diff
@@ -0,0 +1,42 @@
...
+        
+        int wait = r.nextInt(6);
+        Thread.currentThread().sleep(wait);
+        model.addAttribute("quote", quote);
+        
+        return "quote";
+    }
+    
+    @GetMapping("/time")
+    public String getTime(Model model) throws InterruptedException {
```
andrzejsydor, line 36: why not use: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/ModelAndView.html

```suggestion
    public String getTime(ModelAndView model) throws InterruptedException {
```

andrzejsydor (changes requested):
comments added

jmeter/src/main/java/com/baeldung/dashboard/controllers/DashBoard.java
muasif80: Its a matter of preference. Right now its just a demo app to create some endpoints. I consider use of Model ok here. Thanks.

PR #13346, opened 2023-01-25
Update pom.xml

tgvLt:
Added dependency for Junit.


lor6:
Hi @tgvLt We already have this defined in the parent pom, so this is not necessary here.
Thanks.

PR #13366, opened 2023-01-28
BAEL-5924 Java 8 Stream with Batch Processing Support

alemoles:


core-java-modules/core-java-streams-4/src/main/java/com/baeldung/streams/processing/CustomBatchIterator.java
```diff
@@ -14,20 +14,20 @@
     private List<T> currentBatch;
     private final Iterator<T> iterator;
 
-    public CustomBatchIterator(Iterator<T> sourceIterator, int batchSize) {
+    private CustomBatchIterator(Iterator<T> sourceIterator, int batchSize) {
```
saajn, line 17: @alemoles shouldn't this be public in this solution?

PR #13373, opened 2023-01-29
Fix link to the previous article in README.md

honeykakkar:
Currently, "prev" points to "/core-java-modules/core-java-collections-1" which is not a valid relative directory in repository.


lor6:
Thanks @honeykakkar

PR #13399, opened 2023-01-31
JAVA-14723 Convert spring-mvc-basics-4 to Spring Boot project and upd…

timis1:
Migrate from spring mvc to Spring-boot


spring-web-modules/spring-mvc-basics-4/src/main/java/com/baeldung/controller/config/StudentControllerConfig.java
```diff
@@ -8,12 +8,12 @@
 import org.springframework.web.context.support.AnnotationConfigWebApplicationContext;
 import org.springframework.web.servlet.DispatcherServlet;
 
-public class StudentControllerConfig //implements WebApplicationInitializer 
+public class StudentControllerConfig //implements WebApplicationInitializer
```
dkapil, line 11: Can we remove this comment "//implements WebApplicationInitializer"

spring-web-modules/spring-mvc-basics-4/src/main/java/com/baeldung/contexts/secure/HelloWorldSecureController.java
```diff
@@ -31,19 +30,15 @@ private void processContext() {
...
-        WebApplicationContext rootContext = ContextLoader.getCurrentWebApplicationContext();
-        System.out.println("context : " + rootContext);
-        System.out.println("context Beans: " + Arrays.asList(rootContext.getBeanDefinitionNames()));
 
         System.out.println("context : " + webApplicationContext);
         System.out.println("context Beans: " + Arrays.asList(webApplicationContext.getBeanDefinitionNames()));
     }
 
-    @GetMapping(path = "/welcome")
+    @GetMapping(path = "/welcome_secure")
```
dkapil, line 38: why are we changing the endpoint here? does it makes any difference?

spring-web-modules/spring-mvc-basics-4/src/main/java/com/baeldung/controller/controller/GreetingsController.java
```diff
@@ -15,7 +15,7 @@ public class GreetingsController {
     
     @GetMapping(value = "/greetings-with-response-body", produces="application/json")
     public String getGreetingWhileReturnTypeIsString() { 
-        return "{\"test\": \"Hello using @ResponseBody\"}";
```
dkapil, line 25: Should we keep the previous body, or else we need to change the mapping as well since it mentions "-with-response-body"

spring-web-modules/spring-mvc-basics-4/src/main/java/com/baeldung/optionalpathvars/ArticleViewerController.java
```diff
@@ -1,23 +0,0 @@
...
-
-import org.springframework.web.bind.annotation.PathVariable;
-import org.springframework.web.bind.annotation.RequestMapping;
-import org.springframework.web.bind.annotation.RestController;
-
-@RestController
-public class ArticleViewerController {
-
-    @RequestMapping(value = {"/article", "/article/{id}"})
-    public Article getArticle(@PathVariable(name = "id") Integer articleId) {
```
dkapil, line 13: where have we moved this controller? was this not in use?

spring-web-modules/spring-mvc-basics-4/src/main/webapp/WEB-INF/index.jsp
```diff
@@ -1,5 +0,0 @@
-<html>
-<body>
```
dkapil, line 2: in which folder we are moving these templates, they may not be needed in the WEB-INF with boot-2 but they could still be needed and should be moved to the directory accessed by thymeleaf

spring-web-modules/spring-mvc-basics-4/src/test/java/com/baeldung/optionalpathvars/ArticleViewerControllerIntegrationTest.java
```diff
@@ -1,54 +0,0 @@
...
-import org.springframework.test.web.servlet.request.MockMvcRequestBuilders;
-import org.springframework.test.web.servlet.result.MockMvcResultMatchers;
-import org.springframework.test.web.servlet.setup.MockMvcBuilders;
-import org.springframework.web.context.WebApplicationContext;
-import com.baeldung.controller.config.WebConfig;
-
-@RunWith(SpringJUnit4ClassRunner.class)
-@WebAppConfiguration
-@ContextConfiguration(classes = { WebConfig.class })
-public class ArticleViewerControllerIntegrationTest {
```
dkapil, line 20: similar question here, do we need this test and the corresponding controller ArticleViewerController?

dkapil (changes requested):

spring-web-modules/spring-mvc-basics-4/src/main/java/com/baeldung/contexts/secure/HelloWorldSecureController.java
timis1: Because we have already a `welcome` path in the `HelloWorldController`.

spring-web-modules/spring-mvc-basics-4/src/main/java/com/baeldung/controller/controller/GreetingsController.java
timis1: I will change the body into `Hello using @ResponseBody from the annotation @RestController`.

spring-web-modules/spring-mvc-basics-4/src/main/java/com/baeldung/optionalpathvars/ArticleViewerController.java
timis1: This is basically covered in this Controller: `ArticleViewerWithRequiredAttributeController`. There are mapped two paths but in case you don't provide an `id`, you will receive an error. What do you think?

spring-web-modules/spring-mvc-basics-4/src/main/webapp/WEB-INF/index.jsp
timis1: This file is not referred anywhere, that is why I removed some of the unused files.

spring-web-modules/spring-mvc-basics-4/src/test/java/com/baeldung/optionalpathvars/ArticleViewerControllerIntegrationTest.java
timis1: Since I removed the controller I removed also the test.

dkapil: Can we rename this test to ArticleViewerWithRequiredAttributeControllerIntegrationTest since content of both the tests are same

spring-web-modules/spring-mvc-basics-4/src/main/java/com/baeldung/optionalpathvars/ArticleViewerController.java
dkapil: required is set as false, even then it will throw an error?

timis1: If I set the required to false for id it will not throw any error.

spring-web-modules/spring-mvc-basics-4/src/test/java/com/baeldung/optionalpathvars/ArticleViewerControllerIntegrationTest.java
timis1: Ok, and then should we remove this one: ArticleViewerControllerWithRequiredAttributeIntegrationTest?

dkapil: I mean yeah, we can rename it to ArticleViewerWithRequiredAttributeControllerIntegrationTest

timis1: I rename it into: ArticleViewerControllerWithRequiredAttributeIntegrationTest.

PR #13407, opened 2023-01-31
enum samples for bael6145 implementing toString() for enums

IfThen2:
enum samples for bael6145 implementing toString() for enums


core-java-modules/core-java-lang-oop-types-2/src/main/java/com/baeldung/enumtostring/FastFood1.java
```diff
@@ -0,0 +1,25 @@
...
+        case PIZZA:
+            return "Pizza Pie";
+        case BURGER:
+            return "Cheese Burger";
+        case TACO:
+            return "Crunchy Taco";
+        case CHICKEN:
+            return "Fried Chicken";
+        default:
+            return null;
```
JavierPeris, line 22: In this cases, it's better to return an empty String.

core-java-modules/core-java-lang-oop-types-2/src/main/java/com/baeldung/enumtostring/FastFood3.java
```diff
@@ -0,0 +1,29 @@
+package com.baeldung.enumtostring;
+
+public enum FastFood3 {
+    PIZZA("Pizza Pie"),
+    BURGER("Cheese Burger"),
+    TACO("Crunchy Taco"),
+    CHICKEN("Fried Chicken"),
+    ;
+
+    String prettyName;
```
JavierPeris, line 10: It's rare to provide a mutable field for an enum, so I would make it final.

IfThen2: I agree :)

I think I was avoiding using `final` because I got dinged for it on my last article.

ss12723

ss22723

But I have made the requested changes.

core-java-modules/core-java-lang-oop-types-2/src/main/java/com/baeldung/enumtostring/FastFood1.java
IfThen2: Done

core-java-modules/core-java-lang-oop-types-2/src/main/java/com/baeldung/enumtostring/FastFood3.java
JavierPeris: Yes, we have to take into account that we have this general rule of avoiding final keywords because they can distract from the main topic. However, for topics where the final keyword can have an impact in the article we can add it.

core-java-modules/core-java-lang-oop-types-2/src/main/java/com/baeldung/enumtostring/FastFood2.java
```diff
@@ -0,0 +1,28 @@
...
+        public String toString() {
+            return "Crunchy Taco";
+        }
+    },
+    CHICKEN {
+        @Override
+        public String toString() {
+            return "Fried Chicken";
+        }
+    },
```
KevinGilmore, line 27: I think we can omit the trailing comma on this line.

JavierPeris (approved):

PR #13440, opened 2023-02-07
BAEL-6046 - MongoDB - Field Level Encryption

ulisseslima:
Review 2.


persistence-modules/spring-boot-persistence-mongodb-3/src/main/java/com/baeldung/boot/csfle/config/converter/BinaryConverter.java
```diff
@@ -0,0 +1,13 @@
+package com.baeldung.boot.csfle.config.converter;
+
+import org.bson.BsonBinary;
+import org.bson.types.Binary;
+import org.springframework.core.convert.converter.Converter;
+
+public class BinaryConverter implements Converter<Binary, BsonBinary> {
```
jyemin, line 7: This converter doesn't seem to be serving any encryption-related purpose.  Why is it necessary?

jyemin: Can you respond to this question please?  I'm still unclear why this is here.  If it's not needed, let's just remove it.

ulisseslima: Hey, @jyemin, without it, I'll get an `org.springframework.core.convert.ConverterNotFoundException: No converter found capable of converting from type [org.bson.types.Binary] to type [org.bson.BsonBinary]` when retrieving documents, in case I run the code with auto decryption off.

In turn, `EncryptedCitizen` uses `BsonBinary` properties because all encryption operations do; using this converter requires less code.

jyemin: The problem won't appear in this application, but since the converter will convert every Binary to BsonBinary, it will affect all BSON binary fields, not just ones representing encrypted data.  This is not likely what a user would want.  So it's probably worth the effort to convert Binary to BsonBinary only when needed by the ClientEncryption API.

PR #13455, opened 2023-02-09
Record vs Final class in Java

drafii:
Jira: https://jira.baeldung.com/browse/BAEL-6201
code for Record vs Final class in Java article.


drafii:
Retest this please.

PR #13457, opened 2023-02-09
JAVA-13870 Move spring-caching,spring-caching-2 to spring-boot-module…

timis1:
Move spring-cashing and spring-cashing-2 to the spring-boot-modules


pom.xml
```diff
@@ -1054,7 +1052,7 @@
                 <module>spring-5-webflux-2</module>
                 <module>spring-activiti</module>
                 <module>spring-batch-2</module>
-                <module>spring-caching-2</module>
+				<module>spring-boot-modules/spring-caching-2</module>
```
dkapil, line 1055: spring-boot-modules is already included in the pom, there is no need to include the submodule explicitly in the root pom.

spring-boot-modules/pom.xml
```diff
@@ -84,6 +84,7 @@
         <module>spring-boot-data-2</module>
         <module>spring-boot-validation</module>
         <module>spring-boot-data-3</module>
+		<module>spring-caching</module>
```
dkapil, line 87: We will have to include here spring-caching and spring-caching-2 both

spring-boot-modules/spring-caching-2/pom.xml
```diff
@@ -12,7 +12,7 @@
         <groupId>com.baeldung</groupId>
         <artifactId>parent-boot-2</artifactId>
         <version>0.0.1-SNAPSHOT</version>
-        <relativePath>../parent-boot-2/pom.xml</relativePath>
+        <relativePath>../../parent-boot-2/pom.xml</relativePath>
```
dkapil, line 15: let's make the actual physical module spring-boot-modules as its parent, since spring-boot-modules is dependendent further on parent-boot-2 it should work fine.

spring-boot-modules/spring-caching/pom.xml
```diff
@@ -12,7 +12,7 @@
         <groupId>com.baeldung</groupId>
         <artifactId>parent-boot-2</artifactId>
         <version>0.0.1-SNAPSHOT</version>
-        <relativePath>../parent-boot-2</relativePath>
+        <relativePath>../../parent-boot-2</relativePath>
```
dkapil, line 15: let's make the actual physical module spring-boot-modules as its parent, since spring-boot-modules is dependendent further on parent-boot-2 it should work fine.

dkapil (changes requested):

pom.xml
timis1: I saw that spring-caching-2 needs to run with jdk9 and above. That is why I left here and I added in the sub module style and to change that location.

spring-boot-modules/pom.xml
timis1: Spring-caching-2 is included in the root project of the pom.xml in the sub module style.

spring-boot-modules/spring-caching-2/pom.xml
timis1: You mean to leave as it is here: `../parent-boot-2/pom.xml`?

spring-boot-modules/spring-caching/pom.xml
timis1: You mean to leave as it is here: `../parent-boot-2/pom.xml`?

spring-boot-modules/spring-caching-2/pom.xml
dkapil: No, instead of making parent-boot-2 as the parent, make spring-boot-modules as the parent. Spring-boot-modules further declares parent-boot-2 as its parent so we should be ok

spring-boot-modules/spring-caching/pom.xml
dkapil: No, instead of making parent-boot-2 as the parent, make spring-boot-modules as the parent. Spring-boot-modules further declares parent-boot-2 as its parent so we should be ok

timis1: Done, I did the changes suggested.

spring-boot-modules/spring-caching-2/pom.xml
timis1: Done, I did the changes suggested.

dkapil (approved):

PR #13466, opened 2023-02-11
JAVA-18264 Update article "A Guide to MongoDB with Java"

timis1:
Upgrading mongodb-driver-sync and on this way I upgraded most of the deprecated libraries and I fixed tests


persistence-modules/java-mongodb/src/main/java/com/baeldung/MongoBsonExample.java
```diff
@@ -1,14 +1,15 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import org.bson.Document;
+
 import com.mongodb.client.MongoClient;
 import com.mongodb.client.MongoClients;
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.MongoDatabase;
```
dkapil, line 12: let's revert this file, since it does not contain any changes

pom.xml
```diff
@@ -1074,6 +1074,7 @@
                 <module>tensorflow-java</module>
                 <module>xstream</module>
                 <module>webrtc</module>
+				<module>persistence-modules/java-mongodb</module> <!-- long running -->
```
dkapil, line 1077: no need to include this in a submodule fashion. persistence-modules is already present in the root pom

persistence-modules/pom.xml
```diff
@@ -43,7 +43,6 @@
         <module>java-jpa</module> <!-- long running -->
         <module>java-jpa-2</module> <!-- long running -->
         <module>java-jpa-3</module>
-        <module>java-mongodb</module> <!-- long running -->
```
dkapil, line 46: why remove from here?

persistence-modules/java-mongodb/src/test/java/com/baeldung/bsontojson/JSONDateFormatEpochTime.java
```diff
@@ -0,0 +1,18 @@
+package com.baeldung.bsontojson;
+
+import org.bson.json.Converter;
+import org.bson.json.StrictJsonWriter;
+
+/**
+ * Convertor to epoch time
+ */
+public class JSONDateFormatEpochTime implements Converter<Long> {
```
dkapil, line 9: let's add the suffix "Converter" here

dkapil: btw converting value to String, shouldn't this be the default behaviour?

dkapil (changes requested):

pom.xml
timis1: persistence-modules is in the jdk 8 profile. I migrate this project to run under jdk9 and above profile. The reason why I migrated to the jdk9 was because morphia-core requires jdk 11, that is why added in the sub-module fashion just that module and I removed the module from the pom.xml the java-mongodb.

persistence-modules/pom.xml
timis1: Because I included in the submodule fashion to run in with jdk 9 and above

persistence-modules/java-mongodb/src/test/java/com/baeldung/bsontojson/JSONDateFormatEpochTime.java
timis1: Because I saw when it serialize for that type of filed is converted into the LocalDateTime style.

pom.xml
dkapil: Ok

persistence-modules/pom.xml
dkapil: Let's comment it and add a comment

persistence-modules/java-mongodb/src/test/java/com/baeldung/bsontojson/JSONDateFormatEpochTime.java
dkapil: Is that a problem with our articles? I mean my intention is that we should have as less custom code as possible unless needed by the article.

persistence-modules/java-mongodb/src/main/java/com/baeldung/MongoBsonExample.java
timis1: I did the changes

persistence-modules/java-mongodb/src/test/java/com/baeldung/bsontojson/JSONDateFormatEpochTime.java
timis1: If we want to keep the same output in the test we should create that converter. That is because of this: 
https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#datetime

persistence-modules/pom.xml
timis1: Done

pom.xml
```diff
@@ -1055,6 +1055,7 @@
...
```
dkapil, line 1058: there are tabs in this line, please revert back to spaces and remove the comment "long running"

PR #13482, opened 2023-02-14
JAVA-18149 Upgrade libraries-2 module to JDK 11

timis1:
Migrate libraries-2 to jdk 11


pom.xml
```diff
@@ -1074,6 +1072,7 @@
                 <module>tensorflow-java</module>
                 <module>xstream</module>
                 <module>webrtc</module>
+				<module>libraries-2</module>
```
dkapil, line 1075: let's add these modules in alphabetical order. Atleast that is what we are trying to maintain :)

libraries-2/src/test/java/com/baeldung/parallel_collectors/ParallelCollectorsUnitTest.java
```diff
@@ -40,7 +40,7 @@ public void shouldProcessInParallelWithParallelCollectors() {
         List<Integer> ids = Arrays.asList(1, 2, 3);
 
         CompletableFuture<List<String>> results = ids.stream()
-          .collect(parallelToList(i -> fetchById(i), executor, 4));
+          .collect(parallelToList(ParallelCollectorsUnitTest::fetchById, executor, 4));
```
dkapil, line 43: is this an improvement or a fix?

dkapil (changes requested):

pom.xml
timis1: I will do

libraries-2/src/test/java/com/baeldung/parallel_collectors/ParallelCollectorsUnitTest.java
timis1: Yes this is an improvement to is the new way to call a static method in a lambda expression.

pom.xml
timis1: Done

libraries-2/pom.xml
```diff
@@ -16,15 +16,7 @@
...
-            <url>http://repository.jboss.org/nexus/content/groups/public/</url>
-            <releases>
-                <enabled>true</enabled>
-                <updatePolicy>never</updatePolicy>
-            </releases>
-            <snapshots>
-                <enabled>true</enabled>
-                <updatePolicy>daily</updatePolicy>
-            </snapshots>
+            <url>https://repo.maven.apache.org/maven2</url>
```
lor6, line 19: do we need to add this repository here explicitly? I think this is the default repository

libraries-2/pom.xml
```diff
@@ -119,22 +117,40 @@
...
+            <version>18.3.12</version>
+        </dependency>
     </dependencies>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>3.0.0-M7</version>
```
lor6, line 132: can we use the property from the main pom to define the version here, or is the 3.0.0-M7 required?

pom.xml
```diff
@@ -999,6 +997,7 @@
                 <module>java-native</module>
                 <module>jsoup</module>
                 <module>ksqldb</module>
+				<module>libraries-2</module>
```
lor6, line 1000: let's fix the formatting here and use spaces instead of tabs

libraries-2/pom.xml
timis1: Yes, you are right. I remove the entire block for **<repositories>**.

libraries-2/pom.xml
timis1: Done I added the variable of **maven-surefire-plugin.version**.

pom.xml
timis1: Done

PR #13485, opened 2023-02-15
JAVA-17818 Split or move spring-cloud-openfeign module (conti-2)

timis1:
Migrate articles from spring-cloud-openfeign to feign project


feign/src/main/resources/logback_spring.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
```
dkapil, line 2: do we need this file for our articles?

feign/src/main/java/com/baeldung/core/exception/NotFoundException.java
```diff
@@ -0,0 +1,18 @@
+package com.baeldung.core.exception;
+
+public class NotFoundException extends Exception {
+
```
dkapil, line 4: why we need these two exception classes?

feign/src/main/java/com/baeldung/core/ExampleApplication.java
```diff
@@ -0,0 +1,16 @@
+package com.baeldung.core;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.cloud.openfeign.EnableFeignClients;
+
+@SpringBootApplication
```
dkapil, line 7: why we need this class? how was this functioning in the spring-cloud-modules module ?

dkapil (changes requested):

feign/src/main/resources/logback_spring.xml
timis1: That file was before, I just tried to align with the latest version and I updated the file because failed at start up.

feign/src/main/java/com/baeldung/core/exception/NotFoundException.java
timis1: Because it is used here:
image

feign/src/main/java/com/baeldung/core/ExampleApplication.java
timis1: Because the code migrated from that module runs with this class.

PR #13507, opened 2023-02-19
BAEL-5915 Added different ways to create an object

ShaPhi7:


ShaPhi7:
The diff looks odd, created https://github.com/eugenp/tutorials/pull/13539 instead.

PR #13511, opened 2023-02-20
JAVA-18136 Add jackson-modules to jdk9-and-above profiles

anastasiosioannidis:


anastasiosioannidis:
retest this please

pom.xml
```diff
@@ -1134,6 +1135,7 @@
                 <module>spring-core-6</module>
                 <module>data-structures</module>
                 <module>ddd-contexts</module>
+                <module>jackson-modules</module>
```
dkapil, line 1138: we will have to remove it from default-first and integration-lite-first also

dkapil (changes requested):

pom.xml
anastasiosioannidis: Ok, done

PR #13513, opened 2023-02-20
Prevent NPE when trying to create a deep copy

chrisociepa:
`a.length` causes NPE when `null` is provided as a value


dkapil:
Thanks @chrisociepa

PR #13515, opened 2023-02-20
Custom Constructors in Records

hajarrs:


core-java-modules/core-java-14/src/main/java/com/baeldung/java14/recordsCustomConstructors/Main.java
```diff
@@ -0,0 +1,34 @@
...
+        List<StudentRecord> sortedStudentRecords = List.copyOf(mutableStudentRecords);
+        System.out.println(sortedStudentRecords);
+
+        for(StudentRecord ss: sortedStudentRecords){
+            System.out.println(ss);
+        }
+
+        StudentRecordV2 srV2 = new StudentRecordV2("Hajar", 4, 80);
+        System.out.println(srV2.grade());
+    }
```
Bwvolleyball, line 33: Would it make more sense to move what we're trying to show here into individual tests?

We don't really need a main method for these examples, so it might be clearer without this, or to have it moved into individual unit tests where what we're showing can be well-explained in the context of the test.

PR #13572, opened 2023-03-02
Configure multiple listeners on same topic

VladFernoaga:


spring-kafka-2/src/main/java/com/baeldung/spring/kafka/multiplelisteners/BookConsumer.java
```diff
@@ -0,0 +1,26 @@
+package com.baeldung.spring.kafka.multiplelisteners;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.kafka.annotation.KafkaListener;
+import org.springframework.stereotype.Service;
+
+@Service
+public class BookConsumer {
+    private static final Logger LOGGER = LoggerFactory.getLogger(BookConsumer.class);
```
JavierPeris, line 10: Although is static final in this case logger is not a constant so please use lowercase variable for non-constant variable.

spring-kafka-2/src/main/java/com/baeldung/spring/kafka/multiplelisteners/BookConsumer.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @KafkaListener(topics = "books", groupId = "books-content-search")
+    public void bookContentSearchConsumer(BookEvent event) {
+        LOGGER.info(String.format("Books event received for full-text search indexing => %s", event.toString()));
```
JavierPeris, line 14: Instead of using String.format is better for performance issues to use placeholder for variable so in this case:

`logger.info("Books event received for full-text search indexing => {}", event);`

Please do the same for the rest of log entries.

spring-kafka-2/src/main/java/com/baeldung/spring/kafka/multiplelisteners/BookEvent.java
```diff
@@ -0,0 +1,15 @@
+package com.baeldung.spring.kafka.multiplelisteners;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
```
JavierPeris, line 9: I think @Data annotation already includes @NoArgsConstructor and @AllArgsConstructor.

spring-kafka-2/src/test/java/com/baeldung/spring/kafka/multiplelisteners/KafkaMultipleListenersIntegrationTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.kafka.config.KafkaListenerEndpointRegistry;
+import org.springframework.kafka.core.KafkaTemplate;
+import org.springframework.kafka.listener.AcknowledgingConsumerAwareMessageListener;
+import org.springframework.kafka.listener.ConcurrentMessageListenerContainer;
+import org.springframework.kafka.test.context.EmbeddedKafka;
+
+@SpringBootTest(classes = MultipleListenersApplicationKafkaApp.class)
+@EmbeddedKafka(partitions = 1, brokerProperties = { "listeners=PLAINTEXT://localhost:9092", "port=9092" })
+public class KafkaMultipleListenersIntegrationTest {
```
JavierPeris, line 23: Using JUnit 5 we don't need to make the class nor the method public so you can get rid of them.

spring-kafka-2/src/main/java/com/baeldung/spring/kafka/multiplelisteners/BookEvent.java
VladFernoaga: The @Data annotation includes only the  @RequiredArgsConstructor, and for this DTO, it generates only the @NoArgsConstructor. I didn't found a way that @Data to generate both  @AllArgsConstructor and @NoArgsConstructor.

PR #13629, opened 2023-03-12
BAEL-6255 - Run a Spring Boot application in AWS Lambda

cesarevalenti90:


KevinGilmore (commented):
See notes about spelling of "John" and order of arguments to assertEquals

PR #13640, opened 2023-03-14
simplify StandardCharsets use

pjfanning:
little point in taking the StandardCharset.UTF_8 and trying to build a new Charset instance from it


lor6:
Thanks @pjfanning

PR #13668, opened 2023-03-17
Generating Time Based UUIDs

Eredisse:


core-java-modules/core-java-uuid/src/main/java/com/baeldung/timebaseduuid/UUIDCreator.java
```diff
@@ -0,0 +1,41 @@
...
+
+        int threadCount = 128;
+        int iterationCount = 100_000;
+        ConcurrentMap<UUID, Long> uuidMap = new ConcurrentHashMap<>();
+        AtomicLong collisionCount = new AtomicLong();
+        long startNanos = System.nanoTime();
+        CountDownLatch endLatch = new CountDownLatch(threadCount);
+
+        for (long i = 0; i < threadCount; i++) {
+            final long threadId = i;
```
ashleyfrieze, line 24: I'm not sure this needs to be `final`. In Java 8 it's _effectively final_ which I think means we can omit the `final` keyword here.

core-java-modules/core-java-uuid/src/main/java/com/baeldung/timebaseduuid/UUIDCreator.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class UUIDCreator {
+
+    public static void main(String[] args) throws InterruptedException {
+
+        int threadCount = 128;
+        int iterationCount = 100_000;
+        ConcurrentMap<UUID, Long> uuidMap = new ConcurrentHashMap<>();
```
ashleyfrieze, line 18: Can this LHS  just be `Map<UUID, Long>`

PR #13706, opened 2023-03-24
String to int encapsulation

ciphx:
Jira : https://jira.baeldung.com/browse/BAEL-5859
Link to article: https://drafts.baeldung.com/wp-admin/post.php?post=159699&action=edit


core-java-modules/core-java-string-conversions/src/main/java/com/baeldung/stringtoint/StringToIntEncapsulation.java
```diff
@@ -0,0 +1,53 @@
+package com.baeldung.stringtoint;
+
+import java.util.Optional;
+import org.apache.commons.lang3.math.NumberUtils;
+
+public class StringToIntEncapsulation {
```
vinipx, line 6: _StringToIntConverter_ sounds more appropriate, no?

core-java-modules/core-java-string-conversions/src/main/java/com/baeldung/stringtoint/StringToIntEncapsulation.java
```diff
@@ -0,0 +1,53 @@
...
+    public static Integer convertStringToIntUsingIntegerDecode(String input){
+        try {
+            return Integer.decode(input);
+        } catch (Exception e) {
+            // log or handle the error
+            return null; // or Integer.MIN_VALUE, or some other default value
+        }
+    }
+
+    public static Integer converStringToIntUsingOptional(String input){
```
vinipx, line 38: conver**t**StringToIntUsingOptional

core-java-modules/core-java-string-conversions/src/main/java/com/baeldung/stringtoint/StringToIntEncapsulation.java
```diff
@@ -0,0 +1,53 @@
...
+        Optional<Integer> parsedInt;
+        try {
```
vinipx, line 40: As per described in the video review, I'd rather remove this _try-catch_ block and go with this instead:

**return Optional.ofNullable(Integer.parseInt(s)).orElse(null);** 

(remember that **null** not necessarily is the best option here)

core-java-modules/core-java-string-conversions/src/main/java/com/baeldung/stringtoint/StringToIntEncapsulation.java
```diff
@@ -0,0 +1,53 @@
...
+            parsedInt = Optional.of(Integer.parseInt(input));
+        } catch (Exception e) {
+            // log or handle the error
+            parsedInt = Optional.empty();
+        }
+        return parsedInt.orElse(null);
+    }
+
+    public static int convertStringToIntUsingNumberUtils(String input){
+        //returns Integer.MIN_VALUE as the default value if conversion fails
```
vinipx, line 50: I don't think this comment is really necessary - it is explicit by the line 51 itself.

core-java-modules/core-java-string-conversions/src/test/java/com/baeldung/datetostring/DateToStringFormatterUnitTest.java
vinipx, line 1: This file have nothing to do with the article - better to remove it.

vinipx (changes requested):

core-java-modules/core-java-string-conversions/src/main/java/com/baeldung/stringtoint/StringToIntEncapsulation.java
ciphx: so orElse doesn't handle the scenario when parsing fails. It throws a number format exception. It only handles scenarios where we receive a null value.

vinipx (approved):

PR #13714, opened 2023-03-25
JAVA-18167 Upgrade persistence-modules to JDK 11

timis1:
Migrate the persistence-modules to the jdk9 and above profile


persistence-modules/pom.xml
```diff
@@ -14,12 +14,10 @@
     </parent>
 
     <modules>
-        <module>activejdbc</module>
         <module>apache-bookkeeper</module>
         <module>apache-cayenne</module>
         <module>apache-derby</module>
         <module>blaze-persistence</module>
-        <module>core-java-persistence</module>
```
dkapil, line 22: let's not do this, let's do that reverse.
Move persistence-modules into jdk9 in the main pom.xml
and in this pom.xml those that are being migrated enabled it and keep others as disabled. Gradually we will enable all of them as we make progress on this Jira

pom.xml
```diff
@@ -1232,7 +1235,6 @@
                 <module>messaging-modules</module>
 
                 <module>persistence-modules/questdb</module>
-                <module>vaadin</module>
```
dkapil, line 1235: why removing vaadin as part of this PR?

timis1: I added it back. In my commit wasn't show me that was deleted.

persistence-modules/pom.xml
timis1: I will do the changes in the next commit

mariudotsh:
Hi @timis1 , I'm adding changes in `persistence-modules/hibernate-jpa` and the integration tests aren't run by the CI job. Check my draft PR here: https://github.com/eugenp/tutorials/pull/13848. I think this might be due `persistence-modules` being in `default-jdk9-and-above` profile which excludes integration tests and/or CI misconfiguration.

You also upgraded Spring Boot to 3+, which requires at least JDK 17 https://spring.io/blog/2022/05/24/preparing-for-spring-boot-3-0#upgrade-to-java-17 so java version 17 should be explicitly set in the pom.

I don't have access to the JIRA issues you're working on, so I don't know the big picture of these changes. Letting you know and sending an email to support in hope of getting my tests to run in the CI :sweat_smile: .

timis1:
> Hi @timis1 , I'm adding changes in `persistence-modules/hibernate-jpa` and the integration tests aren't run by the CI job. Check my draft PR here: #13848. I think this might be due `persistence-modules` being in `default-jdk9-and-above` profile which excludes integration tests and/or CI misconfiguration.
> 
> You also upgraded Spring Boot to 3+, which requires at least JDK 17 https://spring.io/blog/2022/05/24/preparing-for-spring-boot-3-0#upgrade-to-java-17 so java version 17 should be explicitly set in the pom.
> 
> I don't have access to the JIRA issues you're working on, so I don't know the big picture of these changes. Letting you know and sending an email to support in hope of getting my tests to run in the CI 😅 .

Hi @mariudotsh, 
Sorry for late response. Looks like you added a new example in the hibernate-jpa. Yes, I think it will be useful to explicitly set the java version 17 since the minimum required version is 17. 
My task was related to upgrade all the articles from `persistence-modules` migrate the `default-jdk9-and-above` and on this way we migrated also to the hibernate 6.

mariudotsh:
@timis1 Thanks for the reply. I just found in the docs that "- that’s because unit tests run in the standard build; integration and live tests do not" so I'm assuming this is intended. Sorry for the trouble.

PR #13719, opened 2023-03-25
Event Sourcing - Recreate user with id from the UserCreatedEvent instead of creating a new one

FabioBentoLuiz:
As the events are geting replayed, it should recreate  the user with the ID as it is provided on the UserCreatedEvent, otherwise it will generate a new ID for each read request.


kasramp:
Thanks for the correction @FabioBentoLuiz

PR #13726, opened 2023-03-27
[JAVA-19672] Moved apache-spark module to jdk9-and-above profile

panos-kakos:


panos-kakos:
Won't be implemented in this jira task

PR #13745, opened 2023-03-31
Dev 19389

Bipinkumar27:


persistence-modules/spring-data-jpa-filtering/src/test/java/com/baeldung/repository/CustomerRepositoryIntegrationTest.java
```diff
@@ -43,6 +45,20 @@ public void givenQueryMethod_whenEmailIsNull_thenFoundByNullEmail() {
         assertEquals("D", actual.getName());
     }
 
+    @Test
+    public void givenQueryMethod_findByUUid() {
```
dkapil, line 49: @Bipinkumar27 - we will have to update the method names here. given_when_then is written in a format as "given a condition, when executing an action, then results", given clause is option in many cases.
This method should be named as
`givenUUIDIsPresent_whenQueryMethod_thenFetchedCorrectly`

second method should be called as 

`givenUUIDIsNull_whenQueryMethod_thenFetchedCorrectly`

PR #13761, opened 2023-04-03
https://jira.baeldung.com/browse/BAEL-5859

ciphx:
moved the code to a different module


vinipx:
Thanks for adjusting it.

PR #13771, opened 2023-04-04
Move lombok-module to jdk9-and-above group (excluded lombok-custom)

ehsansasanian:


lombok-modules/lombok-custom/pom.xml
```diff
@@ -9,8 +9,9 @@
 
     <parent>
         <groupId>com.baeldung</groupId>
-        <artifactId>lombok-modules</artifactId>
+        <artifactId>parent-modules</artifactId>
```
dkapil, line 12: this is not needed, the lombok-custom can still refer to parent-modules as its parent

lombok-modules/lombok-custom/pom.xml
```diff
@@ -9,8 +9,9 @@
...
         <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
```
dkapil, line 14: we usually give the module name instead of its pom file, so something like ../../parent-name could work

dkapil (changes requested):

lombok-modules/lombok-custom/pom.xml
ehsansasanian: To apply this, I had to use the maven-compiler-plugin for the submodule (lombok-custome) to override the JDK version and make it compilable.

PR #13810, opened 2023-04-10
pr: add return types for assertion; test WrongVoucher vs GoodVoucher

press0:


kasramp:
Hi @press0 - thanks for the improvement.

PR #13817, opened 2023-04-10
JAVA-20163 Upgrade hibernate specific modules to JDK 11

timis1:
Migration hibernate specific modules


persistence-modules/hibernate-enterprise/src/main/java/com/baeldung/hibernate/pojo/PolygonEntity.java
```diff
@@ -1,10 +1,10 @@
 package com.baeldung.hibernate.pojo;
 
-import com.vividsolutions.jts.geom.Polygon;
+import org.locationtech.jts.geom.Polygon;
```
dkapil, line 3: I am not sure what we are trying to do here, but we have a Polygon class in the java itself https://docs.oracle.com/javase/7/docs/api/java/awt/Polygon.html

persistence-modules/hibernate-enterprise/src/test/java/com/baeldung/hibernate/HibernateSpatialIntegrationTest.java
```diff
@@ -9,23 +9,23 @@
...
-import javax.persistence.Query;
+import jakarta.persistence.Query;
 
 import org.hibernate.Session;
 import org.hibernate.Transaction;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
+import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.jts.io.ParseException;
```
dkapil, line 20: just check if we can use the Java libraries https://docs.oracle.com/javase/8/docs/technotes/guides/2d/spec/j2d-geom.html

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/persistmaps/mapkey/User.java
```diff
@@ -1,16 +1,15 @@
...
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+import jakarta.validation.constraints.Size;
 
 import org.hibernate.validator.constraints.Length;
-import org.hibernate.validator.constraints.CreditCardNumber;
-import org.hibernate.validator.constraints.Currency;
 
 @Entity
+@Table(name="users2")
```
dkapil, line 12: any reason we can't use "users"?

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/pojo/EntityDescription.java
```diff
@@ -14,10 +17,12 @@ public class EntityDescription implements Serializable {
 
     private String description;
 
-    @Any(
-            metaDef = "EntityDescriptionMetaDef",
-            metaColumn = @Column(name = "entity_type")
-    )
+    @Any
+    @AnyDiscriminator(DiscriminatorType.STRING)
+    @AnyDiscriminatorValue(discriminator = "S", entity = Employee.class)
```
dkapil, line 22: can you let me know with which article this is linked?

pom.xml
```diff
@@ -366,6 +366,7 @@
 
                 <module>web-modules</module>
                 <module>persistence-modules/deltaspike</module> <!-- delta spike it doesn't yet the jakarta API-->
+                <module>persistence-modules/hibernate-ogm</module> <!-- hibernate-ogm wasn't updated and not support jakarta API -->
```
dkapil, line 369: we need to create a "lowest" priority jira to track this

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/pojo/inheritance/Laptop.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.hibernate.annotations.Polymorphism;
+import org.hibernate.annotations.PolymorphismType;
+
+import jakarta.persistence.Entity;
+import jakarta.persistence.Id;
+
+@Entity
+@Polymorphism(type = PolymorphismType.IMPLICIT)
+public class Laptop implements Item {
```
dkapil, line 11: Do we need this new file?

persistence-modules/hibernate-queries/src/test/java/com/baeldung/hibernate/keywords/HibernateKeywordsApplicationLiveTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.hibernate.Session;
+import org.hibernate.SessionFactory;
+import org.hibernate.cfg.Configuration;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class HibernateKeywordsApplicationLiveTest {
```
dkapil, line 16: please let me know the article with which this changes is related?

dkapil (changes requested):

persistence-modules/hibernate-annotations/src/main/java/com/baeldung/hibernate/customtypes/AddressType.java
```diff
@@ -92,57 +69,52 @@ public boolean equals(Object x, Object y) throws HibernateException {
...
+        empAdd.setAddressLine1(rs.getString(position));
 
         if (rs.wasNull())
             return null;
 
-        empAdd.setAddressLine2(rs.getString(names[1]));
-        empAdd.setCity(rs.getString(names[2]));
-        empAdd.setCountry(rs.getString(names[3]));
-        empAdd.setZipCode(rs.getInt(names[4]));
+        empAdd.setAddressLine2(rs.getString(position));
```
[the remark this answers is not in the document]
timis1: I just migrated to the new schema and the only parameter that I was able to assign was an integer instead of array. I will further investigate your question.

persistence-modules/hibernate-annotations/src/test/java/com/baeldung/hibernate/creationupdatetimestamp/HibernateCreationUpdateTimestampIntegrationTest.java
```diff
@@ -71,22 +72,22 @@ void whenCreatingEntity_ThenCreatedOnAndLastUpdatedOnAreBothSet() {
     }
 
     @Test
-    void whenCreatingEntity_ThenCreatedOnAndLastUpdatedOnAreNotEqual() {
+    void whenCreatingEntity_ThenCreatedOnAndLastUpdatedOnAreEqual() {
```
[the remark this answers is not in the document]
timis1: I would say that the test itself was wrong since the timestamp cannot be different since was initialized in just a single place and were initialized once.

persistence-modules/hibernate-enterprise/src/main/java/com/baeldung/hibernate/exception/ProductNotMapped.java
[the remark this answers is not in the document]
timis1: I was able to throw the MappingException just in this way. With the previous example of mapping to a string the exception wasn't thrown.

pom.xml
timis1: I created here a Jira task with lowest priority: https://team.baeldung.com/browse/JAVA-20557

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/pojo/inheritance/Laptop.java
timis1: I added this file because when I query by an interface if that interface doesn't have a class that implements that interface and have this annotation: `@Polymorphism(type = PolymorphismType.IMPLICIT)` it will throw an error. This way the interface Item was implemented in two classes: Laptop and Bag. The class Bag has the annotation Polymorphism with explicit annotation the other one (Laptop )has IMPLICIT Annotation.

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/pojo/EntityDescription.java
timis1: It is related to this article: https://www.baeldung.com/hibernate-dynamic-mapping

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/persistmaps/mapkey/User.java
timis1: Because we already have another entity that has the name `users`

persistence-modules/hibernate-enterprise/src/main/java/com/baeldung/hibernate/pojo/PolygonEntity.java
timis1: I will try, but I wanted to keep the same dependency. The reason here is that because the package was changed
https://stackoverflow.com/questions/70660454/migrating-com-vividsolutions-jts-to-org-locationtech-jts-still-complaining-about

persistence-modules/hibernate-enterprise/src/test/java/com/baeldung/hibernate/HibernateSpatialIntegrationTest.java
timis1: Sure, I will try.

persistence-modules/hibernate-queries/src/test/java/com/baeldung/hibernate/keywords/HibernateKeywordsApplicationLiveTest.java
timis1: I just rename the name of this class because I saw it is trying to connect a real database. 
It is related to this article: https://www.baeldung.com/java-hibernate-db-keywords-as-columns

persistence-modules/hibernate-enterprise/src/main/java/com/baeldung/hibernate/pojo/PolygonEntity.java
timis1: Polygon can be used from java but some of the methods from hibernate spatial will not be compatible with that Java Object (Polygon)

persistence-modules/hibernate-enterprise/src/test/java/com/baeldung/hibernate/HibernateSpatialIntegrationTest.java
timis1: `Coordinate` class doesn't exists in Java Library.

persistence-modules/hibernate-enterprise/src/main/java/com/baeldung/hibernate/pojo/PolygonEntity.java
dkapil: Let's not change this then

persistence-modules/hibernate-annotations/src/main/java/com/baeldung/hibernate/customtypes/AddressType.java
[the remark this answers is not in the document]
timis1: I looked over this class and it seems that the `nullSafeGet` method is not called event if one of the parameter is null (when do the save). Even if I return null from `nullSafeGet` nothing failed.
Do you have any idea or any suggestion?

persistence-modules/hibernate-annotations/src/test/java/com/baeldung/hibernate/creationupdatetimestamp/HibernateCreationUpdateTimestampIntegrationTest.java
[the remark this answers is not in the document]
dkapil: yeah that could work, actually there could be a milliseconds difference b/w creation and update timestamp, let's assume they would be equal for now

persistence-modules/hibernate-annotations/src/main/java/com/baeldung/hibernate/customtypes/AddressType.java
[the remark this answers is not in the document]
timis1: Somehow I respond to this question. I tried to set values on the `empAdd` null and store it. I expected that the after I query that record `nullSafeGet` will be triggered and call that method, this is not happened. If I make a simple a simple query to fetch that record the values are fetched correctly. 
e.g: 
`session.createQuery("FROM OfficeEmployee OE WHERE OE.empAddress.zipcode = :pinCode", OfficeEmployee.class)` 
If I ran this query the values on the `empAddress` are correctly, that is why I asked for your help/ further suggestion.

persistence-modules/hibernate-mapping/src/main/java/com/baeldung/hibernate/pojo/EntityDescription.java
timis1: Sure, I will update all the articles presented in all modules that I made changes.

persistence-modules/hibernate-queries/src/test/java/com/baeldung/hibernate/keywords/HibernateKeywordsApplicationLiveTest.java
timis1: I removed duplicated Unit test since there was already the HibernateKeywordsApplicationManualTest.

PR #13818, opened 2023-04-11
Java 18150

Bipinkumar27:


libraries-3/src/test/java/com/baeldung/immutable/ImmutablePersonUnitTest.java
```diff
@@ -7,7 +7,7 @@
 
 public class ImmutablePersonUnitTest {
 
-    @Test
+    @Test(expected = Exception.class)
```
dkapil, line 10: instead of updating the logic of this test. Add the annotation @Disabled to ignore this test and add a comment linking it to the mutability issue you have created.

Bipinkumar27: done

PR #13829, opened 2023-04-12
JAVA-18299 : Changes made for adding json view example fot hiding req…

Bipinkumar27:
…uest feilds in swagger


spring-boot-modules/spring-boot-mvc-4/src/main/java/com/baeldung/springboot/swagger/views/Views.java
```diff
@@ -0,0 +1,11 @@
+package com.baeldung.springboot.swagger.views;
+
+public class Views {
+    public static class Public{
```
dkapil, line 4: This class is not formatted

dkapil (changes requested):

spring-boot-modules/spring-boot-mvc-4/src/main/java/com/baeldung/springboot/swagger/views/Views.java
Bipinkumar27: done

PR #13857, opened 2023-04-17
BAEL-6355 - OpenAI API with Spring Boot

abh1navv:


spring-boot-modules/spring-boot-libraries-2/src/main/java/com/baeldung/openapi/controller/ChatController.java
```diff
@@ -0,0 +1,44 @@
...
+                apiUrl,
+                request,
+                ChatResponse.class);
+
+        if (response == null || response.getChoices() == null || response.getChoices().isEmpty()) {
+            return "No response";
+        }
+
+        // return the first response
+        return response.getChoices().get(0).getMessage().getContent();
```
Alexenon, line 42: Will be great to remove single comments and replace them with javadoc

spring-boot-modules/spring-boot-libraries-2/src/main/java/com/baeldung/openapi/dto/Message.java
```diff
@@ -0,0 +1,31 @@
...
+
+    private String role;
+    private String content;
+
+    Message(String role, String content) {
+        this.role = role;
+        this.content = content;
+    }
+
+    Message() {
```
Alexenon, line 13: Do you need an empty constructor ?

Alexenon (approved):

spring-boot-modules/spring-boot-libraries-2/src/main/java/com/baeldung/openapi/controller/ChatController.java
abh1navv: Updated

spring-boot-modules/spring-boot-libraries-2/src/main/java/com/baeldung/openapi/dto/Message.java
abh1navv: it was required for deserializing the response

PR #13875, opened 2023-04-21
JPA/Hibernate Associations

KingsleyAmankwah:


timis1:
Hi @KingsleyAmankwah,
The first question is, does your code work locally? :)
I looked over your changes and I give you some advice to progress in fixing your PR:
- please create each class independently (if you need to group more class together create a package and add in that package)
- hibernate 6.x is using Jakarta persistence API 3.0 you have to use the jakarta API annotation and not javax
- In your tests try to use the `EntityManager` class if you want to simulate a store.

I don't know the entire context of your task these are some few suggestion that I saw at the first sight.

KingsleyAmankwah:
@timis1  check out the code and provide some you think will help. Thank you

PR #13894, opened 2023-04-23
BAEL-6319 Hibernate 6 Boolean Converters

mariudotsh:
Mvn install log from machine with all tests passing:
[mvn-install.log](https://github.com/eugenp/tutorials/files/11304040/mvn-install.log)


persistence-modules/hibernate-mapping-2/src/main/java/com/baeldung/hibernate/HibernateUtil.java
```diff
@@ -1,4 +1,6 @@
-package com.baeldung;
+package com.baeldung.hibernate;
```
mariudotsh, line 1: I moved the main classes under "hibernate" package to be consistent with the tests which are under it as well

PR #13913, opened 2023-04-26
Java 15032

Bipinkumar27:


apache-httpclient4/src/test/java/com/baeldung/httpclient/HttpClientHeaderV4LiveTest.java
```diff
@@ -1,15 +1,23 @@
 package com.baeldung.httpclient;
 
+import static org.apache.http.HttpHeaders.USER_AGENT;
```
dkapil, line 3: revert this change, we will not be adding the v4.5 example

dkapil (changes requested):

apache-httpclient4/src/test/java/com/baeldung/httpclient/HttpClientHeaderV4LiveTest.java
Bipinkumar27: done

PR #13914, opened 2023-04-26
Java 20643

Bipinkumar27:


dkapil:
test this please

pom.xml
```diff
@@ -364,7 +364,7 @@
 
                 <module>muleesb</module>
 
-                <module>web-modules</module>
+                <!--<module>web-modules</module>-->
```
dkapil, line 367: instead of commenting, remove them

pom.xml
```diff
@@ -1195,6 +1196,8 @@
                 <module>persistence-modules/questdb</module>
                 <module>vaadin</module>
                 <module>libraries-3</module>
+                <module>web-modules/apache-tapestry</module>
+                <!--<module>web-modules/apache-tapestry</module>-->
```
dkapil, line 1200: add the entire web-modules here and then in the pom.xml of web-modules enable those that. you have upgraded for e.g. apache-tapestry and comment out the rest.

dkapil (changes requested):

pom.xml
Bipinkumar27: sure but I have divided the task into sub-tasks. Should I add all the rest modules in their jira commit?

pom.xml
Bipinkumar27: sure

dkapil:
test this please

PR #13918, opened 2023-04-27
The changes made include refactoring of the code, addition of Java Doc, and implementation of Exception Handling.

raowaqasakram:
The changes made include refactoring of the code, the addition of Java Doc, and implementation of Exception Handling.


lor6:
Thanks @raowaqasakram We usually prefer to avoid adding libraries like lombok if they're not the focus of a certain tutorial because this may make the code more complex for readers to understand. 
We'll go over the other suggested improvements and decide whether they'll be included in a new PR.

PR #13953, opened 2023-05-05
BAEL-6320: added code support for the @NaturalId article

etrandafir93:


persistence-modules/spring-data-jpa-repo-3/src/main/java/com/baeldung/spring/data/jpa/naturalid/HotelRoomsService.java
```diff
@@ -0,0 +1,35 @@
...
+		  .loadOptional(name);
+	}
+
+	public Optional<GuestRoom> guestRoom(int roomNumber, int floor) {
+		Session session = entityManager.unwrap(Session.class);
+		return session.byNaturalId(GuestRoom.class)
+		  .using("roomNumber", roomNumber)
+		  .using("floor", floor)
+		  .loadOptional();
+	}
```
KevinGilmore, line 33: Need to use spaces to indent instead of tabs in all the Java files.

persistence-modules/spring-data-jpa-repo-3/src/main/java/com/baeldung/spring/data/jpa/naturalid/entity/ConferenceRoom.java
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	protected ConferenceRoom() {
+	}
+
+	public Long getId() {
+		return id;
+	}
+
+	public void setId(Long id) {
```
KevinGilmore, line 32: Why have a setter for id but not for name, which you marked as mutable? Seems inconsistent. The other entity also does not have a setter for id.

persistence-modules/spring-data-jpa-repo-3/src/main/java/com/baeldung/spring/data/jpa/naturalid/repository/GuestRoomJpaRepo.java
```diff
@@ -0,0 +1,9 @@
+package com.baeldung.spring.data.jpa.naturalid.repository;
+
+import com.baeldung.spring.data.jpa.naturalid.entity.GuestRoom;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface GuestRoomJpaRepo extends JpaRepository<GuestRoom, Long> {
```
KevinGilmore, line 8: Let's be consistent in our repository naming -- i.e., don't shorten to "Repo".

persistence-modules/spring-data-jpa-repo-3/src/main/java/com/baeldung/spring/data/jpa/naturalid/repository/ConferenceRoomJpaRepo.java
```diff
@@ -0,0 +1,9 @@
+package com.baeldung.spring.data.jpa.naturalid.repository;
+
+import com.baeldung.spring.data.jpa.naturalid.entity.ConferenceRoom;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface ConferenceRoomJpaRepo extends JpaRepository<ConferenceRoom, Long> {
```
KevinGilmore, line 8: Let's be consistent in our repository naming -- i.e., don't shorten to "Repo".

persistence-modules/spring-data-jpa-repo-3/src/main/java/com/baeldung/spring/data/jpa/naturalid/HotelRoomsService.java
etrandafir93: hello Kevin, I have just realized that I have used the wrong IntellJ formater, so I've re-formatted everything now.

In regards to this line in particular, I had a comment for one of my previous reviews stating that I should always use 2 spaces for the line continuations (both for git and code snippets in the article), even if this would mean manually adding the 2 spaces. How should I keep it?

persistence-modules/spring-data-jpa-repo-3/src/main/java/com/baeldung/spring/data/jpa/naturalid/entity/ConferenceRoom.java
etrandafir93: removed the `(mutable = true)`. It was something I was experimenting with and forgot to remove afterward. Additionally, I have removed the setter for the id as well.

persistence-modules/spring-data-jpa-repo-3/src/main/java/com/baeldung/spring/data/jpa/naturalid/repository/GuestRoomJpaRepo.java
etrandafir93: renamed

persistence-modules/spring-data-jpa-repo-3/src/main/java/com/baeldung/spring/data/jpa/naturalid/entity/ConferenceRoom.java
KevinGilmore: Agree.

persistence-modules/spring-data-jpa-repo-3/src/main/java/com/baeldung/spring/data/jpa/naturalid/HotelRoomsService.java
KevinGilmore: Either way is fine in the case of chaining method calls, which is different from a continuation. Some editors prefer two spaces in this case, while the guideline actually calls for four. An example of a continuation is a statement with a lot of operators that you spread over multiple lines (think concatenating a bunch of strings). Another is a method call where you want to put the parameters on separate lines.

PR #13968, opened 2023-05-06
JAVA-16946: disable unit test for core-java-lang-oop-modifiers module

ehsansasanian:
To upgrade H2 version, it is necessary to remove test.mv.db from Jenkins. Therefore, unit tests shouldn't be running for core-java-lang-oop-modifiers module. After deletion, this change will be rolled back, and a .db file will be initiated.


anastasiosioannidis:
retest this please

PR #13974, opened 2023-05-08
JAVA-21106: Adding javax-servlets module to jdk9 and above

Bipinkumar27:


web-modules/javax-servlets-2/src/main/java/com/baeldung/user/check/UserCheckFilter.java
```diff
@@ -43,4 +44,7 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
 
         chain.doFilter(request, response);
     }
```
Bipinkumar27, line 46: Need to implement the default init method of the Filter class as without this method our application was failing while deploying to tomcat

PR #13989, opened 2023-05-10
JAVA-6350: Add comments for SpringDoc v2 compatible with Spring Boot 3

harry9656:


spring-boot-modules/spring-boot-springdoc/pom.xml
```diff
@@ -119,53 +118,6 @@
         </resources>
     </build>
 
-    <profiles>
-        <profile>
-            <id>integration</id>
-            <build>
```
JavierPeris, line 125: Why has this profile been eliminated?

harry9656: I migrated it to the other module.

JavierPeris: My question is: isn't any other article/code taking advantage of that profile?

harry9656: No, the integration with the plugins is part of this article. Other articles don't refer to this part.

spring-boot-modules/spring-boot-springdoc-2/src/main/java/com/baeldung/springdoc/exception/BookNotFoundException.java
```diff
@@ -0,0 +1,10 @@
+package com.baeldung.springdoc.exception;
+
+@SuppressWarnings("serial")
```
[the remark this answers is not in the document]
harry9656: We don't need it. I can't find any reason why it was there, so I removed it.

PR #14012, opened 2023-05-14
JAVA-21305 : Changes made for adding ratpack module to jdk9-and above…

Bipinkumar27:
… module


web-modules/pom.xml
```diff
@@ -25,12 +25,12 @@
         <module>javax-servlets-2</module>
         <module>jee-7</module>
         <module>jooby</module>
```
dkapil, line 27: add a comment here , add the same comment against restx

web-modules/pom.xml
```diff
@@ -25,12 +25,12 @@
...
-        <module>ninja</module>-->
+        <module>linkrest</module>-->
+        <module>ninja</module>
         <!-- <module>play-modules</module> --> <!-- Not a maven project -->
         <!-- <module>raml-modules</module> --> <!-- Not a maven project -->
-        <!--<module>ratpack</module>
-        <module>resteasy</module>
+        <module>ratpack</module>
+        <!--<module>resteasy</module>
         <module>restx</module>
```
dkapil, line 34: we need to define jooby and restx in the submodule fashion in the root pom.xml under *first and *second profile respectively so they are still executed in the jdk8 builds.

dkapil (changes requested):

web-modules/pom.xml
Bipinkumar27: done

web-modules/pom.xml
Bipinkumar27: done

PR #14048, opened 2023-05-19
BAEL-6417: MinIO Java SDK examples

michael-pratt:


libraries-data-2/pom.xml
```diff
@@ -116,6 +116,11 @@
             <artifactId>guava</artifactId>
             <version>${guava.version}</version>
         </dependency>
+        <dependency>
+            <groupId>io.minio</groupId>
+            <artifactId>minio</artifactId>
+            <version>8.5.2</version>
```
KevinGilmore, line 122: Let's move the version string to a property.

PR #14081, opened 2023-05-20
JAVA-6503: Image compression in Java

harry9656:


image-compressing/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.0</version>
+                <configuration>
+                    <source>17</source>
+                    <target>17</target>
+                </configuration>
```
JavierPeris, line 31: Unless there is a specific need, we must use version 8 as JDK.

image-compressing/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+        <artifactId>parent-modules</artifactId>
+        <groupId>com.baeldung</groupId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+
+    <dependencies>
+        <dependency>
+            <groupId>net.coobird</groupId>
+            <artifactId>thumbnailator</artifactId>
+            <version>0.4.19</version>
```
JavierPeris, line 18: We should use properties for dependency versions.

image-compressing/src/main/java/com/baeldung/image/compression/ImageCompressor.java
```diff
@@ -0,0 +1,46 @@
...
+
+            writer.write(null, new javax.imageio.IIOImage(inputImage, null, null), params);
+
+            outputStream.close();
+            writer.dispose();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        } catch (URISyntaxException e) {
+            throw new RuntimeException(e);
+        }
```
JavierPeris, line 44: If we are not going to do anything special we can collapse the Exceptions:
```
catch (IOException | URISyntaxException e) {
            throw new RuntimeException(e);
}
```

PR #14131, opened 2023-05-27
JAVA-21459 Upgrade spring-rest-http-2 to use SpringDoc in place of SpringFox

anuragkumawat:


spring-web-modules/spring-rest-http-2/pom.xml
```diff
@@ -47,11 +37,23 @@
             <artifactId>resilience4j-timelimiter</artifactId>
             <version>${resilience4j.version}</version>
         </dependency>
+		<dependency>
+			<groupId>org.springdoc</groupId>
+			<artifactId>springdoc-openapi-ui</artifactId>
```
dkapil, line 42: let's fix the formatting here

dkapil (changes requested):

spring-web-modules/spring-rest-http-2/pom.xml
anuragkumawat: done

PR #14166, opened 2023-06-01
JAVA-20167 Migrate spring-data-mongodb to JDK 17

timis1:


persistence-modules/pom.xml
```diff
@@ -40,8 +40,6 @@
         <module>java-jpa</module> <!-- long running -->
         <module>java-jpa-2</module>
         <module>java-jpa-3</module>
-        <!-- enable it when persistence-modules is migrated to JDK9+ -->
-        <!-- <module>java-mongodb</module>--> <!-- long running -->
```
lor6, line 44: any reason these rows are deleted?

pom.xml
```diff
@@ -369,6 +369,7 @@
                 <module>persistence-modules/hibernate-ogm</module> <!-- hibernate-ogm wasn't updated because it doesn't support jakarta API -->
                 <module>persistence-modules/java-cassandra</module> <!-- JAVA-21464 cassandra-unit library doesn't support to run with jdk9 and above -->
                 <module>persistence-modules/spring-data-cassandra-reactive</module> <!--JAVA-21844-->
+				<module>persistence-modules/spring-data-neo4j</module>
```
lor6, line 372: let's fix the formatting here to use 4 spaces instead of tabs

persistence-modules/pom.xml
timis1: Because it already included down: https://github.com/eugenp/tutorials/blob/720f6415dbd225125367df1881afb28adc86d737/persistence-modules/pom.xml#L111

pom.xml
timis1: I will fix the conflict and this will solve the space indentation :)

PR #14167, opened 2023-06-01
Update ImmutableCredentials.java

TheNullablePrototype:
maybe typo, added import and annotation @ConstructorBinding


lor6:
Hi @TheNullablePrototype 
According to the Spring doc: https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/context/properties/bind/ConstructorBinding.html 

"A single parameterized constructor implicitly indicates that constructor binding should be used unless the constructor is annotated with `@Autowired`." 

So this is not explicitly required if there is a single parameterized constructor which is the case here.

We'll have a closer look at the code to confirm.

PR #14191, opened 2023-06-07
BAEL-6597: How to get the last Day of the Month in Java

azhwani:
Simple PR


JavierPeris (commented):
Since the article is about getting the last day of the month I think it would be more descriptive to provide methods such as `getLastDayOfMonthUsingCalendar` and so on instead of getMaxDayUsingCalendar.

JavierPeris (approved):

PR #14282, opened 2023-06-21
JAVA-16399 Upgrade gradle-modules/gradle/maven-to-gradle to JDK 17

timis1:


pom.xml
```diff
@@ -351,7 +351,6 @@
 
                 <!-- <module>ethereum</module> --> <!-- JAVA-6001 -->
                 <!-- <module>gradle-modules</module> --> <!-- Not a maven project -->
-                <module>gradle-modules/gradle/maven-to-gradle</module>
```
dkapil, line 354: you need to remove from default-first profile too

timis1: Done

PR #14293, opened 2023-06-23
Java-15020: Changes made for adding DNS round robin test case

Bipinkumar27:


anastasiosioannidis:
Hi @Bipinkumar27 
Regarding my comment "I made a mistake about the apache-httpclient. Should have been httpclient-simple. Sorry for this. I will comment also in your new PR", in closed 14278 PR,

I will just use and adjust in article an already existing givenTimeoutIsConfigured_whenTimingOut_thenTimeoutException method in httpclient-simple.

Anyway i don't think the new class was fitting in HttpAsyncClientLiveTest (async)

So I will close this PR

PR #14317, opened 2023-06-27
Upload file and get its URL from Amazon S3

Michaelin007:


aws-modules/aws-s3/src/main/java/com/baeldung/s3/AWSS3Service.java
```diff
@@ -49,6 +50,11 @@ public void deleteBucket(String bucketName) {
     public PutObjectResult putObject(String bucketName, String key, File file) {
         return s3client.putObject(bucketName, key, file);
     }
+    //uploading object and getting url
+    public URL getObjectURL(String bucketName, String key, File file) {
+        s3client.putObject(bucketName, key, file);
+        return  s3client.getUrl(bucketName, key);
```
saajn, line 56: Remove extra space character here.

Michaelin007: done.
Thanks

PR #14359, opened 2023-07-04
JAVA-22210 Upgrade dependencies versions for spring-apache-camel module

timis1:
Upgrade to the latest camel version

### DIFF
--- a/messaging-modules/spring-apache-camel/pom.xml
+++ b/messaging-modules/spring-apache-camel/pom.xml
@@ -19,12 +19,12 @@
         <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-core</artifactId>
-            <version>${env.camel.version}</version>
+            <version>${camel.version}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-spring</artifactId>
-            <version>${env.camel.version}</version>
+            <version>${camel.version}</version>
             <exclusions>
                 <exclusion>
                     <artifactId>commons-logging</artifactId>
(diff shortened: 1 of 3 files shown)


messaging-modules/spring-apache-camel/src/main/java/com/baeldung/camel/boot/Application.java
```diff
@@ -21,7 +21,8 @@
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.stereotype.Component;
```
anastasiosioannidis, line 23: Please undo the identation

messaging-modules/spring-apache-camel/pom.xml
```diff
@@ -77,12 +77,17 @@
...
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-spring-xml</artifactId>
+            <version>${camel.version}</version>
+        </dependency>
     </dependencies>
 
     <properties>
```
anastasiosioannidis, line 87: Since now 3.12.0 is used, let's delete extra env.camel.version

timis1: env.spring.version is used here:
image

messaging-modules/spring-apache-camel/src/main/java/com/baeldung/camel/boot/Application.java
timis1: DONE

messaging-modules/spring-apache-camel/pom.xml
anastasiosioannidis: Ok, could we have <version>${spring.version}</version> in line 43?
And also adjust the <properties>

Also replace env.camel.version with camel.version (also adjust <properties>)

timis1: Just to rename that property variable ${spring.version}?
${spring.version} is not accessible from parent dependency. It is a spring-boot dependency.

messaging-modules/spring-apache-camel/pom.xml
```diff
@@ -45,7 +45,7 @@
         <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-spring-javaconfig</artifactId>
```
anastasiosioannidis, line 47: It seems we can fully remove the camel-spring-javaconfig, and instead use spring boot
https://stackoverflow.com/questions/73391193/camelconfiguration-deprecated-and-removed

So, remove this and adjust ContentBasedFileRouterConfig class (remove extends and @overrides)

timis1: I did the suggested adjustment

PR #14389, opened 2023-07-11
JAVA-22211 Upgrade dependencies versions for helidon-se module

timis1:
Migration helidon-se dependencies to the latest version

### DIFF
--- a/microservices-modules/helidon/helidon-se/pom.xml
+++ b/microservices-modules/helidon/helidon-se/pom.xml
@@ -28,13 +28,13 @@
         <dependency>
             <groupId>io.helidon.webserver</groupId>
             <artifactId>helidon-webserver-netty</artifactId>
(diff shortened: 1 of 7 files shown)


microservices-modules/helidon/helidon-se/pom.xml
```diff
@@ -28,13 +28,13 @@
         <dependency>
             <groupId>io.helidon.webserver</groupId>
             <artifactId>helidon-webserver-netty</artifactId>
-            <version>${helidon.version}</version>
```
dkapil, line 31: let's keep using property placeholders to easily control version

dkapil (changes requested):

microservices-modules/helidon/helidon-se/pom.xml
timis1: DONE

PR #14435, opened 2023-07-19
Scroll API Implementation

imsurajmishra:
JIRA : https://jira.baeldung.com/browse/BAEL-6512


persistence-modules/spring-boot-persistence-4/pom.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
imsurajmishra, line 1: Since Scroll API needs spring boot 3.1.0, I had to create a new module to avoid too many migrations in existing modules.
I am only adding bare minimum dependency needed to run Scroll API.

PR #14440, opened 2023-07-19
BAEL-6458: Guide to the yield Keyword in Java

atighe:


core-java-modules/core-java-conditionals/src/test/java/com/baeldung/conditionals/YieldTest.java
```diff
@@ -0,0 +1,25 @@
+package com.baeldung.conditionals;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+
+public class YieldTest {
```
ashleyfrieze, line 8: This class need not be `public` for JUnit 5

core-java-modules/core-java-conditionals/src/test/java/com/baeldung/conditionals/YieldTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Test
+    void testAnalyze() {
+        assertEquals("Got a 1", YieldExamples.analyze(YieldExamples.Number.ONE));
```
ashleyfrieze, line 12: I'd probably add some static imports to make this read easier. Static import `YieldExamples` and then it will read `analyze(Number.ONE)` which is clearer.

PR #14587, opened 2023-08-13
BAEL-6813: Moved Magic Square code to core-java-lang-math-3

sazzer:


core-java-modules/core-java-jar/src/test/java/com/baeldung/jar/JarAppUnitTest.java
```diff
@@ -3,10 +3,14 @@
...
+import java.util.regex.Pattern;
+
 public class JarAppUnitTest {
 
     @Test
     public void findClassTest(){
-        Assert.assertTrue(JarApp.findObjectMapperClass().endsWith("jackson-databind-2.13.3.jar"));
+        Pattern databindPattern = Pattern.compile(".*jackson-databind-(\\d)+\\.(\\d)+\\.(\\d)\\.jar$");
+
+        Assert.assertTrue(databindPattern.matcher(JarApp.findObjectMapperClass()).matches());
```
sazzer, line 14: This change is in no way related to this PR, but it's broken and causing the checks to fail.

The actual break is because Jackson is now 2.15.2 instead of 2.13.3, but I've done it this way so that it's resilliant to all future version changes.

pivovarit (approved):

PR #14700, opened 2023-09-02
BAEL-6848: Set a Parameter in a HttpServletRequest in Java

manfred106:
Fix tests in BAEL-6848

### DIFF
--- a/web-modules/javax-servlets-2/src/test/java/com/baeldung/setparam/SanitizeParametersRequestWrapperUnitTest.java
+++ b/web-modules/javax-servlets-2/src/test/java/com/baeldung/setparam/SanitizeParametersRequestWrapperUnitTest.java
@@ -1,8 +1,8 @@
 package com.baeldung.setparam;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.mockito.Mockito.when;
+import static org.assertj.core.api.Assertions.assertThat;
 
 import java.util.HashMap;
 import java.util.Map;
@@ -18,7 +18,7 @@ import org.mockito.junit.MockitoJUnitRunner;
 @RunWith(MockitoJUnitRunner.class)
 public class SanitizeParametersRequestWrapperUnitTest {
 
-    private String NEW_VALUE = "NEW VALUE";
+    private static final String NEW_VALUE = "NEW VALUE";
 
     private Map<String, String[]> parameterMap;
 
@@ -37,7 +37,7 @@ public class SanitizeParametersRequestWrapperUnitTest {
         SanitizeParametersRequestWrapper wrapper = new SanitizeParametersRequestWrapper(request);
         String actualValue = wrapper.getParameter("input");
 
-        assertEquals(actualValue, "&lt;script&gt;alert('Hello');&lt;/script&gt;");
+        assertEquals("&lt;script&gt;alert('Hello');&lt;/script&gt;", actualValue);
     }
 
     @Test(expected = UnsupportedOperationException.class)
@@ -54,7 +54,7 @@ public class SanitizeParametersRequestWrapperUnitTest {
 
         firstCallValues[0] = NEW_VALUE;
         String[] secondCallValues = wrapper.getParameterValues("input");
-        assertNotEquals(firstCallValues, secondCallValues);
+        assertThat(secondCallValues).doesNotContain(NEW_VALUE);
     }
 
 }
--- a/web-modules/javax-servlets-2/src/test/java/com/baeldung/setparam/SetParameterRequestWrapperUnitTest.java
+++ b/web-modules/javax-servlets-2/src/test/java/com/baeldung/setparam/SetParameterRequestWrapperUnitTest.java
@@ -1,7 +1,7 @@
 package com.baeldung.setparam;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.mockito.Mockito.when;
 
 import java.util.HashMap;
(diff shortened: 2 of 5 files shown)


web-modules/javax-servlets-2/src/test/java/com/baeldung/setparam/SanitizeParametersRequestWrapperUnitTest.java
```diff
@@ -18,7 +18,7 @@
 @RunWith(MockitoJUnitRunner.class)
 public class SanitizeParametersRequestWrapperUnitTest {
 
-    private String NEW_VALUE = "NEW VALUE";
+    private final String NEW_VALUE = "NEW VALUE";
```
theangrydev, line 21: private static final

web-modules/javax-servlets-2/src/test/java/com/baeldung/setparam/SetParameterRequestWrapperUnitTest.java
```diff
@@ -18,7 +18,7 @@
 @RunWith(MockitoJUnitRunner.class)
 public class SetParameterRequestWrapperUnitTest {
 
-    private String NEW_VALUE = "NEW VALUE";
+    private final String NEW_VALUE = "NEW VALUE";
```
theangrydev, line 21: private static final

web-modules/javax-servlets-2/src/test/java/com/baeldung/setparam/SanitizeParametersRequestWrapperUnitTest.java
```diff
@@ -54,7 +54,7 @@ public void whenSetValueToWrapperParametersStringArray_thenThe2ndCallShouldNotEq
 
         firstCallValues[0] = NEW_VALUE;
         String[] secondCallValues = wrapper.getParameterValues("input");
-        assertNotEquals(firstCallValues, secondCallValues);
+        assertThat(firstCallValues).isNotEqualTo(secondCallValues);
```
theangrydev, line 57: I believe @KevinGilmore might have been trying to suggest you to do something like this:
```suggestion
        assertThat(secondCallValues).doesNotContain(NEW_VALUE);
```

web-modules/javax-servlets-2/src/test/java/com/baeldung/setparam/SetParameterRequestWrapperUnitTest.java
```diff
@@ -55,7 +55,7 @@ public void whenSetValueToWrapperParametersStringArray_thenThe2ndCallShouldNotEq
 
         firstCallValues[0] = NEW_VALUE;
         String[] secondCallValues = wrapper.getParameterValues("input");
-        assertNotEquals(firstCallValues, secondCallValues);
+        assertThat(firstCallValues).isNotEqualTo(secondCallValues);
```
theangrydev, line 58: I believe @KevinGilmore might have been trying to suggest you to do something like this:
```suggestion
        assertThat(secondCallValues).doesNotContain(NEW_VALUE);
```

theangrydev (changes requested):

web-modules/javax-servlets-2/src/test/java/com/baeldung/setparam/SanitizeParametersRequestWrapperUnitTest.java
manfred106: done

web-modules/javax-servlets-2/src/test/java/com/baeldung/setparam/SetParameterRequestWrapperUnitTest.java
manfred106: done

web-modules/javax-servlets-2/src/test/java/com/baeldung/setparam/SanitizeParametersRequestWrapperUnitTest.java
manfred106: isNotEqualTo is correct.
We are asserting whether the 2nd call of getParaemterValues() is not equal to the modified value from the 1st call.

web-modules/javax-servlets-2/src/test/java/com/baeldung/setparam/SetParameterRequestWrapperUnitTest.java
manfred106: isNotEqualTo is correct.
We are asserting whether the 2nd call of getParaemterValues() is not equal to the modified value from the 1st call.

web-modules/javax-servlets-2/src/test/java/com/baeldung/setparam/SanitizeParametersRequestWrapperUnitTest.java
theangrydev: I understand that isNotEqualTo correctly asserts that the 2nd call is not equal to the modified value from the 1st call

However, this is not really using any of the power of AssertJ to make the intention of the assertion clearer, which is what I think Kevin was trying to get you to take a look at

Logically, you are modifying with NEW_VALUE and then you want to assert that this change was not effective

A buggy implementation might for example append NEW_VALUE to the end of the array, and your isNotEqualTo assertion would be green even though there is a bug

By explicitly checking for NEW_VALUE being absent, you get a bit more test coverage, and the intention of the assertion is clearer

manfred106: Assertion updated

web-modules/javax-servlets-2/src/test/java/com/baeldung/setparam/SetParameterRequestWrapperUnitTest.java
manfred106: Assertion updated

theangrydev (approved):

PR #14711, opened 2023-09-03
BAEL-6572: Modified dto to remove lombok annotation as per guidelines and review comment

balasr21:


theangrydev (approved):

mapstruct/src/main/java/com/baeldung/expression/dto/LicenseDto.java
```diff
@@ -3,13 +3,6 @@
...
-import lombok.AllArgsConstructor;
-import lombok.Builder;
-import lombok.Data;
-
-@AllArgsConstructor
-@Data
-@Builder
 public class LicenseDto {
 
     private UUID id;
```
theangrydev, line 8: The id field is actually unused and unmapped - can we update the LicenseMapper to map this field, just to have a complete example? And add one more unit test to show the mapping. We do not need to make any changes to the article, since this change is unrelated to the topic.

balasr3: sure, usually id are populated by the database when the License entity is being used persisted, hence I didn't explicitly add any mapping. But I can dd it if it will make more sense

balasr3: done

balasr3: just saw PR is already merged, let me raise another PR

PR #14712, opened 2023-09-04
BAEL-6572: Added mapping for id and test, based on review comment

balasr21:


mapstruct/src/test/java/com/baeldung/expression/mapper/LicenseMapperUnitTest.java
theangrydev, line 1: Instead of modifying each existing test, add a new test just for the id field. In that test, set the id in the LicenseDto, then assert the same id is present in the License after mapping.

theangrydev (changes requested):

mapstruct/src/test/java/com/baeldung/expression/mapper/LicenseMapperUnitTest.java
balasr3: Done but I have just validated that it's not null since the mapping is not based on licenseDto, as per my mapping example its populated with a random UUID as usually this will be decided during persistence. Please let me know if any comments

theangrydev: Which class are you suggesting would have the UUID automatically populated by the persistence layer, both the License and the LicenseDto?

In my mind, I thought that the LicenseDto was (as a Data Transfer Object) was the object that would be persisted, and the License is the domain object that is mapped to/from the LicenseDto.

The flow I would usually expect is:
1. LicenseDto is saved to persistence layer, including an automatically generated UUID
2. LicenseDto is later retrieved from the persistence layer
3. LicenseDto is mapped to domain object License, keeping the same UUID
4. License might later be modified due to some business rules
5. License is mapped back to LicenseDto to be saved, keeping the same UUID
6. LicenseDto is updated in the persistence layer, using the UUID as a key

Did you have something different in mind? What scenario is there when the License and LicenseDto would have a different automatically generated UUID?

theangrydev (changes requested):
I just noticed we did not remove Lombok for the License too - can you make this change too please? Lombok should not appear in the examples at all

mapstruct/src/test/java/com/baeldung/expression/mapper/LicenseMapperUnitTest.java
balasr3: @theangrydev Apologies for the confusion, its been a while I wrote this and I thought of something else but now realised mapper is different. Yes you are right, licenseDto is the input and we map it to domain object License so we will get Id from the licenseDto. I have modified the PR accordingly

theangrydev (changes requested):
I just noticed we did not remove Lombok for the License too - can you make this change too please? Lombok should not appear in the examples at all

theangrydev (approved):

PR #14756, opened 2023-09-12
[JAVA-19576] Initial commit

panos-kakos:


persistence-modules/spring-hibernate-6/src/test/resources/.gitignore
```diff
@@ -0,0 +1,13 @@
+*.class
+
+#folders#
```
dkapil, line 3: I don't think we need multiple gitignores, we already have one that is defined at the repo root. Let's remove these

persistence-modules/spring-hibernate-6/pom.xml
```diff
@@ -0,0 +1,149 @@
...
+            <artifactId>spring-data-jpa</artifactId>
+            <version>${org.springframework.data.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate.orm</groupId>
+            <artifactId>hibernate-core</artifactId>
+            <version>${hibernate.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate.orm</groupId>
```
dkapil, line 51: Please ensure we are only adding the bare minimum dependencies needed for these 3 articles

dkapil (changes requested):

PR #14769, opened 2023-09-14
JAVA-23380: Changes made for upgrading sprint-boot-autoconfiguration …

Bipinkumar27:
…to spring-boot-3

### DIFF
--- a/spring-boot-modules/spring-boot-autoconfiguration/src/main/java/com/baeldung/autoconfiguration/MySQLAutoconfiguration.java
+++ b/spring-boot-modules/spring-boot-autoconfiguration/src/main/java/com/baeldung/autoconfiguration/MySQLAutoconfiguration.java
@@ -3,10 +3,10 @@ package com.baeldung.autoconfiguration;
 import java.util.Arrays;
 import java.util.Properties;
 
-import javax.persistence.EntityManagerFactory;
 import javax.sql.DataSource;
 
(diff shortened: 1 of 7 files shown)


spring-boot-modules/spring-boot-autoconfiguration/src/main/java/com/baeldung/autoconfiguration/MySQLAutoconfiguration.java
```diff
@@ -31,7 +31,13 @@
 import org.springframework.orm.jpa.vendor.HibernateJpaVendorAdapter;
 import org.springframework.util.ClassUtils;
 
-@Configuration
+import jakarta.persistence.EntityManagerFactory;
+
+/*@Configuration
```
dkapil, line 36: let's not forget to remove comment code from here

dkapil (changes requested):

spring-boot-modules/spring-boot-autoconfiguration/src/main/java/com/baeldung/autoconfiguration/MySQLAutoconfiguration.java
anastasiosioannidis: @Bipinkumar27, indeed, are these comments needed here?

PR #14786, opened 2023-09-18
BAEL-5755: Convert an XML file to CSV file

ChernoVl:


vinipx:
@ChernoVl general - please, remove all comments from code samples.

PR #14816, opened 2023-09-23
BAEL-6964: Convert Positive Int to Negative and Vice Versa in Java

PaulGreer1:


PaulGreer1:
Hi, Luis Javier Peris. Nothing has changed since the first PR. I needed to organize my branches so I could have multiple concurrent PRs.

PR #14840, opened 2023-09-26
BAEL-6835, PrintWriter vs FileWriter in Java

nadiahylary:


core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
```diff
@@ -0,0 +1,144 @@
...
+import java.nio.file.Paths;
+import java.time.LocalDateTime;
+
+
+import org.apache.commons.io.FileUtils;
+import org.junit.jupiter.api.*;
+
+public class CharStreamsUnitTest {
+    public String todoListPwFile = "src/main/resources/pw_todos.txt";
+    public String todoListFwFile = "src/main/resources/fw_todos.txt";
```
theangrydev, line 16: We might as well also define these as new File(..) instead of Strings

theangrydev (changes requested):

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
theangrydev, line 1: We can also remove the assertTrue(file.exists()); assertions, since the following assertion on the file contents is stronger and will only pass if the file exists. Just another small thing to keep the test code minimal,

theangrydev: And you can use AssertJ assertThat(file).hasContent("..."); to avoid having to introduce the FileUtils dependency

nadiahylary:
1- The test class, CharsTreamsUnitTest.java doesn't pass the checks when the FileNotFoundExceptions aren't handled manually .
2- The BeforeEach method doesn't pass the checks either when the IOException is not manually handled.
3- Using the try-with-resources syntax doesn't work either.
4- Passing the file path strings directly in the File constructors generates errors because I am using the files created in the BeforeEach method as containers for the return values of my methods: ` pwFile = CharStreams.writingStringToFilePrintWriter(TODO_LIST_PW_FILE , TODO.toString());`
5- For consistency purposes, I prefer to pass file paths as Strings to all my writer methods' arguments, due to their constructors' second parameters used in some of the CharStreams.java methods.

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
```diff
@@ -0,0 +1,114 @@
...
+    private static File pwFile = new File(TODO_LIST_PW_FILE); 
+    private static File fwFile =  new File(TODO_LIST_FW_FILE);
+
+    @BeforeEach
+    public void createFile() throws IOException {
+        try{
+            pwFile.mkdirs();
+            pwFile.createNewFile();
+            fwFile.mkdirs();
+            fwFile.createNewFile();
```
theangrydev, line 29: As discussed, fix the bug in this test where you are accidentally creating a directory for the actual pwFile/fwFile path, instead of creating a directory for the containing folder

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
```diff
@@ -0,0 +1,114 @@
...
+        } catch (FileNotFoundException fnfe) {
+            fnfe.printStackTrace();
+        }
+    }
+    
+
+    @AfterEach
+    public void clean() throws IOException {
+        Files.deleteIfExists(Paths.get(TODO_LIST_PW_FILE));
+        Files.deleteIfExists(Paths.get(TODO_LIST_FW_FILE));
```
theangrydev, line 111: To avoid yet another conversion to/from strings and files, please use the existing file variables here:

```suggestion
        Files.deleteIfExists(pwFile.toPath());
        Files.deleteIfExists(fwFile.toPath());

```

core-java-modules/core-java-io-5/pom.xml
```diff
@@ -14,6 +14,18 @@
     </parent>
 
     <dependencies>
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <version>2.13.0</version>
+        </dependency>
```
theangrydev, line 21: Remove unused dependency

core-java-modules/core-java-io-5/pom.xml
```diff
@@ -14,6 +14,18 @@
...
+        <!-- https://mvnrepository.com/artifact/org.assertj/assertj-core -->
+        <dependency>
+             <groupId>org.assertj</groupId>
+             <artifactId>assertj-core</artifactId>
+             <version>3.24.2</version>
```
theangrydev, line 26: Introduce a property for the dependency version ${assertj.version}

theangrydev (changes requested):

theangrydev:
> 1- The test class, CharsTreamsUnitTest.java doesn't pass the checks when the FileNotFoundExceptions aren't handled manually .
> 2- The BeforeEach method doesn't pass the checks either when the IOException is not manually handled.

As discussed, there is a bug in the test, fix the bug and then the extra try-catch blocks can be removed.
 
> 3- Using the try-with-resources syntax doesn't work either.

Please give an example of when this does not work for you and I can help you to fix it.

> 4- Passing the file path strings directly in the File constructors generates errors because I am using the files created in the BeforeEach method as containers for the return values of my methods: ` pwFile = CharStreams.writingStringToFilePrintWriter(TODO_LIST_PW_FILE , TODO.toString());`

I do not understand what you mean here, but one overarching issue is that you should not have these methods at all. Please inline the methods into the main test body. Tests are written with the code under test clearly visible in the test method. Having a layer of indirection with the CharStreams methods makes it much harder to see what is going on in these tests.

In addition, you should not be assigning to a field that is supposed to be a constant. We can use the same constant for both the constructor and for the assertion. This is preferred to having multiple variables that do the same job (single responsibility principle).

> 5- For consistency purposes, I prefer to pass file paths as Strings to all my writer methods' arguments, due to their constructors' second parameters used in some of the CharStreams.java methods.

Once you have inlined the CharStreams.java methods, it will become more clear that there are redundant new File creations that can be replaced by the constant fields you have defined. 

> 6- I use the flush() methods in the article intentionally for teaching purposes, since the PrintWriter can automatically flush as opposed to the FileWriter, it'll be more appropriate to omit the flush() method only when demonstrating the auto-flushing feature of the PrintWriter.

Makes sense, understood.

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+public class CharStreamsUnitTest {
+
+    private static DailyTodo TODO = new DailyTodo("Code", LocalDateTime.now(), true);
+    private static DailyTodo TODO_1 = new DailyTodo("Sleep", LocalDateTime.now(), false);
+    private static DailyTodo TODO_2 = new DailyTodo("Read", LocalDateTime.now(), false);
+    public static String TODO_LIST_PW_FILE = "src/main/resources/pw_todos.txt";
+    public static String TODO_LIST_FW_FILE = "src/main/resources/fw_todos.txt";
+    private static File pwFile = new File(TODO_LIST_PW_FILE); 
+    private static File fwFile =  new File(TODO_LIST_FW_FILE);
```
theangrydev, line 21: Let's delete these `pwFile` and `fwFile` fields altogether

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
theangrydev, line 29: Let's use the Files API instead here, and avoid the explicit creation of the new file. The `PrintWriter` and `FileWriter` will already take care of file creation for us.

```suggestion
        Files.createDirectories(Paths.get(TODO_LIST_PW_FILE).getParent());
        Files.createDirectories(Paths.get(TODO_LIST_FW_FILE).getParent());

```

theangrydev (changes requested):

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
```diff
@@ -0,0 +1,116 @@
...
+import java.time.LocalDateTime;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import org.junit.jupiter.api.*;
+
+public class CharStreamsUnitTest {
+
+    private static final DailyTodo TODO = new DailyTodo("Code", LocalDateTime.now(), true);
+    private static final DailyTodo TODO_1 = new DailyTodo("Sleep", LocalDateTime.now(), false);
+    private static final DailyTodo TODO_2 = new DailyTodo("Read", LocalDateTime.now(), false);
```
theangrydev, line 17: I also just realised that there is no reason for us to have multiple objects here, since you only ever use one object per test. So, you can delete TODO_1 and TODO_2 and just use TODO. The reason for this change is that it can confuse the reader as they search to understand why there are 3 different variables used in the tests.

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
```diff
@@ -0,0 +1,116 @@
...
+    @Test
+    public void whenUsingTextData_thenFileWriterWritesToFile() throws IOException {
+        try {
+            CharStreams.writingStringToFileFileWriter(TODO_LIST_FW_FILE, TODO.toString());
+            assertThat(new File(TODO_LIST_FW_FILE)).hasContent(TODO.toString());
+        } catch (FileNotFoundException fnfe) {
+            fnfe.printStackTrace();
+        }
+        
+    }
```
theangrydev, line 47: For these two tests where we are using a String directly, I think it would be best to avoid using the TODO object and instead use a String directly. The reason for this change is that it can confuse the reader as they try to understand the purpose of the TODO object and why a String was not used directly.

nadiahylary:
I made the changes as you required but now the checks are failing.

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Test
+    public void whenWritingToExixtingFile_thenFileWriterAppendsToFile() throws IOException {
+        try {
+            CharStreams.appendingToFileFileWriter(TODO_LIST_FW_FILE, TODO_2);
+            assertThat(new File(TODO_LIST_FW_FILE)).hasContent(TODO_2.toString());
+        } catch (FileNotFoundException fnfe) {
+            fnfe.printStackTrace();
+        }
+    }
```
theangrydev, line 87: For these two append tests, to truly test append you need to first begin the test with an existing file that has existing content. Then, after the append, we should assert that both the previous content and the appended content is present.

For example:
```
        String existingText = "existing text";
        Files.write(Paths.get(TODO_LIST_FW_FILE), existingText.getBytes(StandardCharsets.UTF_8), StandardOpenOption.CREATE_NEW);

        try (FileWriter fileWriter = new FileWriter(new File(TODO_LIST_FW_FILE), StandardCharsets.UTF_8, true)) {
            fileWriter.append(TODO.toString());
        }

        assertThat(new File(TODO_LIST_FW_FILE)).hasContent(existingText + TODO.toString());
```

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Test
+    public void whenWritingToExixtingFile_thenPrintWriterAppendsToFile() throws IOException {
+        try {
+            CharStreams.appendingToFilePrintWriter(TODO_LIST_PW_FILE, TODO_2);
+            assertThat(new File(TODO_LIST_PW_FILE)).hasContent(TODO_2.toString());
+        } catch (FileNotFoundException fnfe) {
+            fnfe.printStackTrace();
+        }
+    }
```
theangrydev, line 77: For these two append tests, to truly test append you need to first begin the test with an existing file that has existing content. Then, after the append, we should assert that both the previous content and the appended content is present.

NOTE: when I tested this with PrintWriter, it does not append in the way you assumed it worked. In the following code, the original file is overwritten:
```
        String existingText = "existing text";
        Files.write(Paths.get(TODO_LIST_PW_FILE), existingText.getBytes(StandardCharsets.UTF_8), StandardOpenOption.CREATE_NEW);

        try (PrintWriter printWriter = new PrintWriter(new File(TODO_LIST_PW_FILE), StandardCharsets.UTF_8)) {
            printWriter.append(TODO.toString());
        }

        assertThat(new File(TODO_LIST_PW_FILE)).hasContent(existingText + TODO.toString());
```

One correct way to append to an existing file using a PrintWriter is:
```
        String existingText = "existing text";
        Files.write(Paths.get(TODO_LIST_PW_FILE), existingText.getBytes(StandardCharsets.UTF_8), StandardOpenOption.CREATE_NEW);

        try (PrintWriter printWriter = new PrintWriter(new FileOutputStream(TODO_LIST_PW_FILE, true), false, StandardCharsets.UTF_8)) {
            printWriter.append(TODO.toString());
        }

        assertThat(new File(TODO_LIST_PW_FILE)).hasContent(existingText + TODO.toString());
```

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
```diff
@@ -0,0 +1,116 @@
...
+        try {
+            CharStreams.writingNonStringToFileFileWriter(TODO_LIST_FW_FILE, TODO_1);
+            assertThat(new File(TODO_LIST_FW_FILE)).hasContent(TODO_1.toString());
+        } catch (FileNotFoundException fnfe) {
+            fnfe.printStackTrace();
+        }
+    }
+
+    @Test
+    public void whenWritingToExixtingFile_thenPrintWriterAppendsToFile() throws IOException {
```
theangrydev, line 70: Typo in test name
```suggestion
    public void whenWritingToExistingFile_thenPrintWriterAppendsToFile() throws IOException {

```

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
```diff
@@ -0,0 +1,116 @@
...
+        try {
+            CharStreams.appendingToFilePrintWriter(TODO_LIST_PW_FILE, TODO_2);
+            assertThat(new File(TODO_LIST_PW_FILE)).hasContent(TODO_2.toString());
+        } catch (FileNotFoundException fnfe) {
+            fnfe.printStackTrace();
+        }
+    }
+
+    @Test
+    public void whenWritingToExixtingFile_thenFileWriterAppendsToFile() throws IOException {
```
theangrydev, line 80: Typo in test name
```suggestion
    public void whenWritingToExistingFile_thenFileWriterAppendsToFile() throws IOException {

```

theangrydev (changes requested):

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
```diff
@@ -0,0 +1,116 @@
...
+    @BeforeEach
+    public void createFiles() throws IOException {
+        Files.createDirectories(Paths.get(TODO_LIST_PW_FILE).getParent());
+        Files.createDirectories(Paths.get(TODO_LIST_FW_FILE).getParent());
+    }
+
+    @Test
+    public void whenUsingTextData_thenPrintWriterWritesToFile() throws IOException {
+        try {
+            pwFile = CharStreams.writingStringToFilePrintWriter(TODO_LIST_PW_FILE , TODO.toString());
```
theangrydev, line 31: This is why your current branch is failing to compile at the moment

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Test
+    public void whenHandlingError_thenFileWriterHandlesError() {
+        try (FileWriter fileWriter = new FileWriter(TODO_LIST_FW_FILE, true)) {
+        fileWriter.write(TODO_2.toString());
+        } catch (IOException e) {
+            fail("error not expected", e);
+        }
+        assertThat(new File(TODO_LIST_PW_FILE)).hasContent(TODO_2.toString());
```
theangrydev, line 106: The assertion was made on the wrong file and that is why the test is failing

```suggestion
        assertThat(new File(TODO_LIST_FW_FILE)).hasContent(TODO_2.toString());

```

core-java-modules/core-java-io-5/src/main/java/com/baeldung/charstreams/CharStreams.java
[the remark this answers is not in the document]
theangrydev: This has still not been addressed. Can you please confirm you have read this comment and understand what you need to do?

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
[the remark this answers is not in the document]
theangrydev: This has still not been addressed. Can you please confirm you have read this comment and understand what you need to do?

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
nadiahylary: The TODO object is been converted into String. I use it to avoid hard coding the String every time I need it and to avoid creating an extra variable.

nadiahylary:
All the checks have passed.

theangrydev:
> All the checks have passed.

Please scroll up and review all the unresolved issues, I think you might have accidentally missed them by only seeing the most recent ones. Thank you.

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
theangrydev: I think this is going to be confusing for the reader. I can leave it as an outstanding point that we can get a 2nd opinion on from a senior editor when they review it.

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
nadiahylary: I added a new file creation with the first two lines in CharStreams.java: 
`public static void appendingToFilePrintWriter(String fileName, DailyTodo content) throws IOException {
        String existingText = "existing text";
        Files.write(Paths.get(fileName), existingText.getBytes(StandardCharsets.UTF_8), StandardOpenOption.CREATE_NEW);
        PrintWriter printWriter = new PrintWriter(new File(fileName), StandardCharsets.UTF_8);
        printWriter.append(content.toString());`

but it now fails.

I did something similar for the FileWriter append method.

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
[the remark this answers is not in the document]
nadiahylary: Understood and done!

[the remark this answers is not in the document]
nadiahylary: Understood and done!

core-java-modules/core-java-io-5/src/main/java/com/baeldung/charstreams/CharStreams.java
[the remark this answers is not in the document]
nadiahylary: Understood and done!

nadiahylary:
Resolved all the issues, but the checks are failing.

theangrydev:
> Resolved all the issues, but the checks are failing.

Can you please confirm that you are able to compile and run your code locally? From your commit history, it looks as if you might be using the GitHub Web UI to make changes, rather than actually running the code on your local machine

Which IDE are you using?

nadiahylary:
I use IntelliJ IDEA.
And yes, the recent edits from 5 hours ago were made with GitHub web, before that I was working in my IDE and all the tests were passing.

theangrydev:
> I use IntelliJ IDEA. And yes, the recent edits from 5 hours ago were made with GitHub web, before that I was working in my IDE and all the tests were passing.

Please use your IDE to compile the code and run the tests. It will tell you where there are compilation errors, and if a test is failing, you can use the debugger to investigate why.

Do you need any help with either of these things?

nadiahylary:
Okay, no thanks.

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
```diff
@@ -0,0 +1,117 @@
+package com.baeldung.charstreams;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.io.*;
+import java.nio.file.*;
+import java.nio.file.Paths;
+import java.time.LocalDateTime;
```
theangrydev, line 8: @nadiahylary I spent a few minutes to show you what I would do if I was writing these tests

Feel free to keep or adjust this as you wish

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
```diff
@@ -0,0 +1,117 @@
+package com.baeldung.charstreams;
```
theangrydev, line 1:
```suggestion
package com.baeldung.charstreams;

import static org.junit.jupiter.api.Assertions.*;

import java.io.*;
import java.nio.charset.StandardCharsets;
import java.nio.file.*;
import java.nio.file.Paths;
import java.time.LocalDateTime;

import static org.assertj.core.api.Assertions.assertThat;
import org.junit.jupiter.api.*;

public class CharStreamsUnitTest {

    private static final DailyTodo DAILY_TODO = new DailyTodo("Code", LocalDateTime.now(), true);
    private static final String EXAMPLE_TEXT = "example";

    private static final String TODO_LIST_PW_FILE = "src/test/resources/pw_todos.txt";
    private static final String TODO_LIST_FW_FILE = "src/test/resources/fw_todos.txt";
    

    @BeforeEach
    public void createFileDirectories() throws IOException {
        Files.createDirectories(Paths.get(TODO_LIST_PW_FILE).getParent());
        Files.createDirectories(Paths.get(TODO_LIST_FW_FILE).getParent());
    }

    @Test
    public void whenUsingTextData_thenPrintWriterWritesToFile() throws IOException {
        try (PrintWriter printWriter = new PrintWriter(new File(TODO_LIST_PW_FILE))) {
            printWriter.print(EXAMPLE_TEXT);
            printWriter.flush();
            assertThat(new File(TODO_LIST_PW_FILE)).hasContent(EXAMPLE_TEXT);
        }
    }

    @Test
    public void whenUsingTextData_thenFileWriterWritesToFile() throws IOException {
        try (FileWriter fileWriter = new FileWriter(new File(TODO_LIST_FW_FILE))) {
            fileWriter.write(EXAMPLE_TEXT);
            fileWriter.flush();
            assertThat(new File(TODO_LIST_FW_FILE)).hasContent(EXAMPLE_TEXT);
        }
    }

    @Test
    public void whenUsingFormattedData_thenPrintWriterWritesToFile() throws IOException {
        try (PrintWriter printWriter = new PrintWriter(new File(TODO_LIST_PW_FILE), StandardCharsets.UTF_8)) {
            printWriter.print(DAILY_TODO);
            printWriter.flush();
            assertThat(new File(TODO_LIST_PW_FILE)).hasContent(DAILY_TODO.toString());
        }
    }

    @Test
    public void whenUsingFormattedData_thenFileWriterWritesToFile() throws IOException {
        try (FileWriter fileWriter = new FileWriter(new File(TODO_LIST_FW_FILE))) {
            fileWriter.write(DAILY_TODO.toString());
            fileWriter.flush();
            assertThat(new File(TODO_LIST_FW_FILE)).hasContent(DAILY_TODO.toString());
        }
    }

    @Test
    public void whenWritingToExistingFile_thenPrintWriterAppendsToFile() throws IOException {
        String existingText = "existing text";
        Files.write(Paths.get(TODO_LIST_PW_FILE), existingText.getBytes(StandardCharsets.UTF_8), StandardOpenOption.CREATE_NEW);

        try (PrintWriter printWriter = new PrintWriter(new FileOutputStream(TODO_LIST_PW_FILE, true), false, StandardCharsets.UTF_8)) {
            printWriter.append(DAILY_TODO.toString());
            printWriter.flush();
            assertThat(new File(TODO_LIST_PW_FILE)).hasContent(existingText + DAILY_TODO.toString());
        }
    }

    @Test
    public void whenWritingToExistingFile_thenFileWriterAppendsToFile() throws IOException {
        String existingText = "existing text";
        Files.write(Paths.get(TODO_LIST_FW_FILE), existingText.getBytes(StandardCharsets.UTF_8), StandardOpenOption.CREATE_NEW);

        try (FileWriter fileWriter = new FileWriter(new File(TODO_LIST_FW_FILE), StandardCharsets.UTF_8, true)) {
            fileWriter.append(DAILY_TODO.toString());
            fileWriter.flush();
            assertThat(new File(TODO_LIST_FW_FILE)).hasContent(existingText + DAILY_TODO.toString());
        }
    }

    @Test
    public void whenCheckingError_thenPrintWriterReturnsFalse() throws IOException {
        try (PrintWriter printWriter = new PrintWriter(new File(TODO_LIST_PW_FILE), StandardCharsets.UTF_8)) {
            printWriter.print(DAILY_TODO);
            printWriter.flush();
            assertFalse(printWriter.checkError());
        }
    }

    @Test
    public void whenHandlingError_thenFileWriterHandlesError() {
        try (FileWriter fileWriter = new FileWriter(TODO_LIST_FW_FILE)) {
            fileWriter.write(DAILY_TODO.toString());
        } catch (IOException e) {
            fail("error not expected", e);
        }
    }

    @Test
    public void whenAutoFlushEqualTrue_thenPrintWriterFlushesAutomatically() throws FileNotFoundException {
        try (PrintWriter printWriter = new PrintWriter(new FileOutputStream(TODO_LIST_PW_FILE), true)) {
            printWriter.println(DAILY_TODO);
            assertThat(new File(TODO_LIST_PW_FILE)).hasContent(DAILY_TODO.toString() + System.lineSeparator());
        }
    }
    
    @AfterEach
    public void clean() throws IOException {
        Files.deleteIfExists(Paths.get(TODO_LIST_PW_FILE));
        Files.deleteIfExists(Paths.get(TODO_LIST_FW_FILE));
    }
}
```

@nadiahylary I spent a few minutes to show you what I would do if I was writing these tests

Feel free to keep or adjust this as you wish

nadiahylary: Thank you.
6 on 7 tests are passing locally. The one faililng is the:
`@Test
    public void whenWritingToExistingFile_thenPrintWriterAppendsToFile() throws IOException {
        String existingText = "existing text";
        Files.write(Paths.get(TODO_LIST_PW_FILE), existingText.getBytes(StandardCharsets.UTF_8), StandardOpenOption.CREATE_NEW);

        try (PrintWriter printWriter = new PrintWriter(new File(TODO_LIST_PW_FILE), StandardCharsets.UTF_8)) {
            printWriter.append(TODO.toString());
        }

        assertThat(new File(TODO_LIST_PW_FILE)).hasContent(existingText + TODO.toString());
    }`

theangrydev: Yes, this is the bug that I told you about here:
https://github.com/eugenp/tutorials/pull/14840#discussion_r1343180479

The bug is caused by the PrintWriter not being constructed in append mode. To fix the problem, you can do this:
```
PrintWriter printWriter = new PrintWriter(new FileOutputStream(TODO_LIST_PW_FILE, true), false, StandardCharsets.UTF_8);
```

Note the `FileOutputStream` which has a `true` append parameter.

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
theangrydev: As mentioned above, the cause of the issue is constructing the writer not in append mode, see the second example above for how to specify append mode

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
nadiahylary: I fixed it but now it is given a different error: _FileAlreadyExistsException_

theangrydev: This is probably because you have a leftover file from a previous partial test run. The simplest solution would be to move the cleanup logic into the `@BeforeEach` and remove the `@AfterEach`

```
   @BeforeEach
    public void createFiles() throws IOException {
        Files.createDirectories(Paths.get(TODO_LIST_PW_FILE).getParent());
        Files.deleteIfExists(Paths.get(TODO_LIST_PW_FILE));
        Files.createDirectories(Paths.get(TODO_LIST_FW_FILE).getParent());
        Files.deleteIfExists(Paths.get(TODO_LIST_FW_FILE));
    }
```

theangrydev (changes requested):
Outstanding issues from my perspective:
1. use try-with-resources
2. flush in every test apart from the auto flush test
3. implement the whenAutoFlushEqualTrue_thenPrintWriterFlushesAutomatically
4. move cleanup file delete logic into before method
5. use a string constant for the first two tests about text data

See https://github.com/eugenp/tutorials/pull/14840#discussion_r1343301800 for my suggestions on how to address these, you are welcome to use this code directly, it already works well

core-java-modules/core-java-io-5/src/test/java/com/baeldung/charstreams/CharStreamsUnitTest.java
nadiahylary: if I create a new file and instantly delete it what's the point of creating it at all? And this will generate FileNotFoundException at every test because tge file is deleted immediately after creation.

theangrydev: First of all, the `@BeforeEach` suggested does not create any files. It ensures the parent directory of the expected file path is created, so that the actual tests are able to make a strong assumption about that location existing. There is no file creation here, just a parent directory creation.

Secondly, the purpose of deleting in the  `@BeforeEach`, is to cope with cases like the one you may have encountered where, during development, the filesystem is left in an intermediate state (e.g. a test fails during a debugging session and no cleanup code was run). By deleting before the test runs, we are enforcing a precondition for the test, allowing it to assume that the file does not exist at the point in time the test is executed.

nadiahylary: Okay, everything now works.

nadiahylary:
Please I need some assistance with this: implement the whenAutoFlushEqualTrue_thenPrintWriterFlushesAutomatically

nadiahylary:
> Outstanding issues from my perspective:
> 
> 1. use try-with-resources
> 2. flush in every test apart from the auto flush test
> 3. implement the whenAutoFlushEqualTrue_thenPrintWriterFlushesAutomatically
> 4. move cleanup file delete logic into before method
> 5. use a string constant for the first two tests about text data
> 
> See [#14840 (comment)](https://github.com/eugenp/tutorials/pull/14840#discussion_r1343301800) for my suggestions on how to address these, you are welcome to use this code directly, it already works well

I did all these but the checks are still failing.

nadiahylary:
Everything works well locally.

PR #14852, opened 2023-09-28
JAVA-24378 Fix references to parents - Week 35 - 2023

anuragkumawat:


spring-boot-modules/spring-boot-ssl-bundles/pom.xml
```diff
@@ -4,10 +4,10 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
-        <groupId>org.springframework.boot</groupId>
-        <artifactId>spring-boot-starter-parent</artifactId>
-        <version>3.1.3</version>
-        <relativePath/>
+        <groupId>com.baeldung</groupId>
```
anastasiosioannidis, line 7: I guess the reason you didn't use the spring-boot-modules as parent, is because spring-boot-ssl-bundles is upgraded to Spring Boot 3, and spring-boot-modules is still on "parent-boot-2", correct?

anuragkumawat: Yes, you are correct.

PR #14867, opened 2023-09-28
BAEL-7026: Check If a String Contains a Number Value in Java

azhwani:
+ Create new methods for checking if a string contains a digit
+ Create unit test cases for the methods


core-java-modules/core-java-string-operations-6/src/main/java/com/baeldung/strcontainsnumber/StrContainsNumberUtils.java
```diff
@@ -0,0 +1,77 @@
...
+
+    static boolean checkUsingGuava(String input) {
+        if (input == null || input.isEmpty()) {
+            return false;
+        }
+
+        String result = CharMatcher.forPredicate(Character::isDigit)
+          .retainFrom(input);
+
+        return result.length() > 0;
```
JavierPeris, line 74: We can replace `result.length() > 0` with `!result.isEmpty()`

core-java-modules/core-java-string-operations-6/src/main/java/com/baeldung/strcontainsnumber/StrContainsNumberUtils.java
```diff
@@ -0,0 +1,77 @@
...
+            return false;
+        }
+
+        return input.chars()
+          .anyMatch(Character::isDigit);
+    }
+
+    static boolean checkUsingApacheCommonsLang(String input) {
+        String result = StringUtils.getDigits(input);
+        return result != null && result.length() > 0;
```
JavierPeris, line 63: We forgot to replace this length check.
```suggestion
        return result != null && !result.isEmpty();
```

azhwani: Good catch ;) done

core-java-modules/core-java-string-operations-6/src/main/java/com/baeldung/strcontainsnumber/StrContainsNumberUtils.java
azhwani: done

JavierPeris (approved):

PR #14889, opened 2023-10-01
[JAVA-12051] Moved code to core-java-streams-simple for ebook

panos-kakos:


core-java-modules/core-java-streams-simple/src/main/java/com/baeldung/streams/forEach/ReverseList.java
anastasiosioannidis, line 1: This code seems to belong to article:
https://www.baeldung.com/java-collection-stream-foreach

which isn't included in our list.
Is there a reason to copy this class?

core-java-modules/core-java-streams-simple/src/main/java/com/baeldung/streams/filter/Customer.java
anastasiosioannidis, line 1: This code belongs to article: 
https://www.baeldung.com/java-stream-filter-lambda
which is indeed included in our list.

Why is this class copied and not moved?
 Is this class used in other articles?

panos-kakos: Yes it is. 
Article :  https://www.baeldung.com/java-stream-filter-count
Class : StreamCountUnitTest

PR #14896, opened 2023-10-03
JAVA-25517 Upgrade spring-boot-bootstrap to Spring Boot 3

timis1:


spring-boot-modules/spring-boot-bootstrap/pom.xml
```diff
@@ -47,13 +48,12 @@
         <dependency>
             <groupId>io.rest-assured</groupId>
             <artifactId>rest-assured</artifactId>
-            <version>${rest-assured.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>javax.servlet</groupId>
-            <artifactId>javax.servlet-api</artifactId>
-            <version>${servlet.version}</version>
```
anastasiosioannidis, line 56: Is that needed?
After removing it, module still builds fine

timis1: That dependency was removed. This appear in the removed items.

PR #14948, opened 2023-10-11
To change default redirect

SamsterZero:
Previously after logging in we used to get the desired index page, but the address bar still had the "/login" in the address bar, this code fixes that


SamsterZero:
Previously after logging in we used to get the desired index page, but the address bar still had the "/login" in the address bar, this code fixes that

SamsterZero:
# Chaning Redirect after logging in

lor6:
Thanks @SamsterZero

PR #14958, opened 2023-10-12
JAVA-26001_D_M: Fix formatting of POMs for modules D to M

Bipinkumar27:


jhipster-modules/jhipster-microservice/car-app/pom.xml
anastasiosioannidis, line 1: All jhipster modules can be ignored in this task.

PR #14961, opened 2023-10-12
JAVA-26001_A_C: Fix formatting of POMs for modules A to C

Bipinkumar27:


core-java-modules/core-java-21/pom.xml
anastasiosioannidis, line 1: "build" should be before "properties", as per jira description

core-java-modules/core-java-collections-array-list-2/pom.xml
anastasiosioannidis, line 1: Attribute "xmlns="http://maven.apache.org/POM/4.0.0"" should be in the first line

"parent" should have one line spacing above.

core-java-modules/core-java-collections-list-6/pom.xml
anastasiosioannidis, line 1: Add an one line spacing between the "parent" end and "project" end.

PR #14994, opened 2023-10-15
[JAVA-26049] Upgraded spring-boot-environment module to spring-boot 3…

panos-kakos:
… version


spring-boot-modules/spring-boot-environment/pom.xml
```diff
@@ -151,7 +147,7 @@
     </profiles>
 
     <properties>
-        <jpa.version>2.2</jpa.version>
```
anastasiosioannidis, line 154: Is this needed to run the PriceCalculationApplication with Spring Boot 3?

PR #14997, opened 2023-10-15
[JAVA-26243] Upgraded spring-di-3 module to spring-boot 3

panos-kakos:


spring-di-3/pom.xml
```diff
@@ -42,11 +42,33 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-test</artifactId>
         </dependency>
```
anastasiosioannidis, line 44: Shouldn't these just come as transitive dependencies from spring-boot-dependencies?

dkapil: @anastasiosioannidis i think this is ok, the spring-boot-dependencies are declared in the dependency Manangement tag in the parent-spring-6, so inherited modules will have to pull the dependency they want.

PR #15089, opened 2023-10-28
BAEL-7086 - Convert ResultSet Into Map

niket17590:


persistence-modules/core-java-persistence-3/src/test/java/com/baeldung/resultsettomap/ResultSetToMapUnitTest.java
```diff
@@ -0,0 +1,105 @@
...
+    public void whenUsingNativeJava_thenConvertResultSetToMap() throws SQLException {
+        ResultSet resultSet = connection.prepareStatement("SELECT * FROM EMPLOYEE")
+            .executeQuery();
+        Map<String, List<String>> valueMap = new HashMap<>();
+
+        while (resultSet.next()) {
+            String empCity = resultSet.getString("empCity");
+            String empName = resultSet.getString("empName");
+            if (!valueMap.containsKey(empCity)) {
+                valueMap.putIfAbsent(empCity, new ArrayList<>());
```
Maiklins, line 55: You can simply call `valueMap.put()`, then it also works with Java 7 and earlier versions. I assume that's what you mean by "native" Java.

PR #15099, opened 2023-10-30
BAEL-7088: Code for Connect 4 Article

sazzer:


algorithms-modules/algorithms-miscellaneous-7/src/test/java/com/baeldung/algorithms/connect4/GameLiveTest.java
```diff
@@ -0,0 +1,40 @@
+package com.baeldung.algorithms.connect4;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class GameLiveTest {
```
KevinGilmore, line 8: This could simply run as a unit test with the regular builds - no need to make it a "LiveTest".

sazzer: Well dang - it was meant to be and I apparently just used the wrong word!

PR #15106, opened 2023-10-31
BAEL-6430 : Generate unique long using UUID | Hangga Aji Sayekti

hangga:


core-java-modules/core-java-uuid/src/test/java/com/baeldung/uuid/UUIDLongGeneratorUnitTest.java
```diff
@@ -0,0 +1,63 @@
...
+import java.text.DecimalFormat;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.IntStream;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class UUIDLongGeneratorUnitTest {
+    final static int n = 1000000;
```
theangrydev, line 16: `private static final` for constant values

core-java-modules/core-java-uuid/src/test/java/com/baeldung/uuid/UUIDLongGeneratorUnitTest.java
```diff
@@ -0,0 +1,63 @@
...
+    UUIDLongGenerator uuidLongGenerator = new UUIDLongGenerator();
```
theangrydev, line 17: `private` for fields

core-java-modules/core-java-uuid/src/test/java/com/baeldung/uuid/UUIDLongGeneratorUnitTest.java
```diff
@@ -0,0 +1,63 @@
...
+                if (!uniqueValues.add(uniqueValue)) {
+                    collisions.incrementAndGet();
+                }
+                if (uniqueValue < 0) {
+                    negative.incrementAndGet();
+                }
+            } catch (IllegalAccessException | InvocationTargetException e) {
+                throw new RuntimeException(e);
+            }
+        });
```
theangrydev, line 55: I think this would be cleaner using a standard for loop instead of `IntStream`, for example:
```suggestion
        int collisions = 0;
        int negative = 0;

        for (int i = 0; i < n; i++) {
            long uniqueValue = (long) method.invoke(uuidLongGenerator);
            if (!uniqueValues.add(uniqueValue)) {
                collisions++;
            }
            if (uniqueValue < 0) {
                negative++;
            }
        });
```

core-java-modules/core-java-uuid/src/test/java/com/baeldung/uuid/UUIDLongGeneratorUnitTest.java
```diff
@@ -0,0 +1,63 @@
...
+        IntStream.range(0, n).forEach(i -> {
+            try {
+                long uniqueValue = (long) method.invoke(uuidLongGenerator);
+                if (!uniqueValues.add(uniqueValue)) {
+                    collisions.incrementAndGet();
+                }
+                if (uniqueValue < 0) {
+                    negative.incrementAndGet();
+                }
+            } catch (IllegalAccessException | InvocationTargetException e) {
```
theangrydev, line 52: Since exceptions are not the focus of this article, let's just have them rethrown with `throws Exception` on all the test methods, so we do not have to think about them

core-java-modules/core-java-uuid/src/test/java/com/baeldung/uuid/UUIDLongGeneratorUnitTest.java
```diff
@@ -0,0 +1,63 @@
...
+    private void printTableHeader() {
+        System.out.format("%-30s %-15s %-15s %-15s %-15s%n", "Approach", "collisions", "negatives", "collision", "negative");
+        System.out.format("%-30s %-15s %-15s %-15s %-15s%n", "(method name)", "count", "count", "probability", "probability");
+        System.out.println("--------------------------------------------------------------------------------------------");
+    }
+
+    private void printOutput(String method, int collisionsCount, int negativeCount, double collisionsProbability, double negativeProbability) {
+        DecimalFormat decimalFormat = new DecimalFormat("#.#####");
+        System.out.format("%-30s %-15s %-15s %-15s %-15s%n", method, collisionsCount, negativeCount, decimalFormat.format(collisionsProbability), decimalFormat.format(negativeProbability));
+    }
```
theangrydev, line 36: We use SLF4J for logging rather than sysout

core-java-modules/core-java-uuid/src/test/java/com/baeldung/uuid/UUIDLongGeneratorUnitTest.java
```diff
@@ -0,0 +1,63 @@
...
+            } catch (IllegalAccessException | InvocationTargetException e) {
+                throw new RuntimeException(e);
+            }
+        });
+
+        double collisionsProbability = (double) collisions.get() / n;
+        double negativeProbability = (double) negative.get() / n;
+        printOutput(method.getName(), collisions.get(), negative.get(), collisionsProbability, negativeProbability);
+
+        assertTrue(collisionsProbability <= 0.001); // threshold = 0.001
```
theangrydev, line 61: We recommend to use AssertJ, for example:
```suggestion
        assertThat(collisionsProbability).isLessThan(COLLISION_THRESHOLD);
```

Also, prefer to extract a constant like `COLLISION_THRESHOLD` instead of the comment.

core-java-modules/core-java-uuid/src/test/java/com/baeldung/uuid/UUIDLongGeneratorUnitTest.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    private void collisionAndNegativeCheck(Method method) {
+        Set<Long> uniqueValues = new HashSet<>();
+        AtomicInteger collisions = new AtomicInteger(0);
+        AtomicInteger negative = new AtomicInteger(0);
+
+        IntStream.range(0, n).forEach(i -> {
+            try {
+                long uniqueValue = (long) method.invoke(uuidLongGenerator);
```
theangrydev, line 45: I'm a bit wary about the use of reflection since it is not the focus of the article, but I do understand why you have used it since it provides the simplest way to reference each method name and implementation together. Let's see if the senior editor has any issues with it when we come to final review.

core-java-modules/core-java-uuid/src/main/java/com/baeldung/uuid/UUIDLongGenerator.java
```diff
@@ -0,0 +1,58 @@
+package com.baeldung.uuid;
+
+import java.nio.ByteBuffer;
+import java.util.UUID;
+
+public class UUIDLongGenerator {
```
theangrydev, line 6: You have found that the probability of collisions with these approaches is between 50% and 100% by experimentation. The title of the article is about generating a unique positive long, so I think it's important that we do our best to generate a positive value, since the negative rate is so high.

So how about, instead of allowing for negative values, this could be a `UUIDPositiveLongGenerator` and you could simply call `Math.abs` on each of the results to ensure a positive value? The test can then be changed to assert that all the values generated are positive.

Then the statistics you generate in the test can just focus on collisions.

theangrydev, line 6: If the reflection approach stays, let's make it clear that these methods are used:

```suggestion
/**
 * Methods are called by reflection in the unit test
 */
@SuppressWarnings("unused")
public class UUIDLongGenerator {
```

core-java-modules/core-java-uuid/src/main/java/com/baeldung/uuid/UUIDLongGenerator.java
```diff
@@ -0,0 +1,58 @@
...
+    public long gethashCode(){
+        return UUID.randomUUID().toString().hashCode();
+    }
+
+    public long combineByteBuffer(){
+        UUID uuid = UUID.randomUUID();
+        ByteBuffer bb = ByteBuffer.wrap(new byte[16]);
+        bb.putLong(uuid.getMostSignificantBits());
+        bb.putLong(uuid.getLeastSignificantBits());
+        bb.rewind(); // Kembalikan posisi buffer ke awal
```
theangrydev, line 24: Translate the comment to English, or remove it

theangrydev (changes requested):

core-java-modules/core-java-uuid/src/test/java/com/baeldung/uuid/UUIDPositiveLongGeneratorUnitTest.java
```diff
@@ -0,0 +1,54 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class UUIDPositiveLongGeneratorUnitTest {
+    private final static int n = 1000000;
+    private final static double COLLISION_THRESHOLD = 0.001;
+    private final UUIDPositiveLongGenerator uuidLongGenerator = new UUIDPositiveLongGenerator();
+    private final Logger logger = LoggerFactory.getLogger(UUIDPositiveLongGeneratorUnitTest.class);
+
+    @Test
+    void whenForeachGenerateLongValue_thenCollisionsCheck() throws InvocationTargetException, IllegalAccessException {
```
theangrydev, line 22: Simplify the throws clause, since these exceptions are not the focus of the test:
```suggestion
    void whenForeachGenerateLongValue_thenCollisionsCheck() throws Exception {
```

core-java-modules/core-java-uuid/src/test/java/com/baeldung/uuid/UUIDPositiveLongGeneratorUnitTest.java
```diff
@@ -0,0 +1,54 @@
...
+        logger.info("-----------------------------------------------------------------------");
+    }
+
+    private void printOutput(String method, int collisionsCount, double collisionsProbability) {
+        DecimalFormat decimalFormat = new DecimalFormat("#.#####");
+        logger.info(String.format("%-30s %-15s %-15s", method, collisionsCount, decimalFormat.format(collisionsProbability)));
+    }
+
+
+    private void collisionCheck(Method method) throws InvocationTargetException, IllegalAccessException {
```
theangrydev, line 41: Simplify the throws clause, since these exceptions are not the focus of the test:
```suggestion
    private void collisionCheck(Method method) throws Exception {
```

core-java-modules/core-java-uuid/src/test/java/com/baeldung/uuid/UUIDPositiveLongGeneratorUnitTest.java
```diff
@@ -0,0 +1,54 @@
...
+        Set<Long> uniqueValues = new HashSet<>();
+        int collisions = 0;
+        for (int i = 0; i < n; i++) {
+            long uniqueValue = (long) method.invoke(uuidLongGenerator);
```
theangrydev, line 45: Let's also assert that the value is positive, for example:
```suggestion
            long uniqueValue = (long) method.invoke(uuidLongGenerator);
            assertThat(uniqueValue).isPositive();
```

core-java-modules/core-java-uuid/src/test/java/com/baeldung/uuid/UUIDPositiveLongGeneratorUnitTest.java
```diff
@@ -0,0 +1,54 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.text.DecimalFormat;
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class UUIDPositiveLongGeneratorUnitTest {
+    private final static int n = 1000000;
```
theangrydev, line 16: Constants should be UPPER_CASE style, and let's name this something more descriptive e.g. NUMBER_OF_CHECKS

theangrydev (changes requested):

core-java-modules/core-java-uuid/src/test/resources/logback.xml
```diff
@@ -2,8 +2,8 @@
 <configuration>
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
-            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n
-            </pattern>
+            <pattern>%d{HH:mm:ss} - %msg%n</pattern>
+            <!--<pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>-->
```
theangrydev, line 6: I'm checking with the editorial team to see if we have a standard pattern that is preferred to be used, or if it is flexible. Note that this change affects every other article that may be put in this module in future.

I believe in the article itself we can truncate the log output to omit the timestamp/thread/level/class/etc and just keep the message itself, checking this point with the team too.

core-java-modules/core-java-uuid/src/test/resources/logback.xml
```diff
@@ -12,6 +12,7 @@
 
     <!-- in order to debug some marshalling issues, this needs to be TRACE -->
     <logger name="org.springframework.web.servlet.mvc" level="WARN" />
+    <logger name="c.b.u.UUIDPositiveLongGeneratorUnitTest" level="INFO"/>
```
theangrydev, line 15: This is redundant because the test only uses INFO logging, and INFO is the default in this file too.

When we set INFO level here, it means "show logs from this class for ERROR, WARN, INFO but not for DEBUG and TRACE".
```suggestion
```

theangrydev (changes requested):
Looks good, just two points about logging, one of which I am checking with the editorial team to understand what is typical here

core-java-modules/core-java-uuid/src/main/java/com/baeldung/uuid/UUIDPositiveLongGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+public class UUIDPositiveLongGenerator {
+    public long getLeastSignificantBits() {
+        return Math.abs(UUID.randomUUID().getLeastSignificantBits());
+    }
+
+    public long getMostSignificantBits() {
+        return Math.abs(UUID.randomUUID().getMostSignificantBits());
+    }
+
+    public long gethashCode() {
```
theangrydev, line 19:
```suggestion
    public long getHashCode() {
```

theangrydev (approved):

core-java-modules/core-java-uuid/src/main/java/com/baeldung/uuid/UUIDPositiveLongGenerator.java
hangga: If so, all the gethashCode() words in the article must be updated too

theangrydev (approved):

core-java-modules/core-java-uuid/src/test/resources/logback.xml
```diff
@@ -2,8 +2,7 @@
 <configuration>
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
-            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n
-            </pattern>
+            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
```
KevinGilmore, line 5: No reason to make this change as it is a whitespace-only change and not a functional change.

PR #15142, opened 2023-11-06
Bael 6576: Enable async with retry in Spring application

saikatcse03:


spring-scheduling/src/test/java/com/baeldung/asyncwithretry/EventServiceIntegrationTest.java
```diff
@@ -0,0 +1,90 @@
...
+    private DownstreamService downstreamService;
+
+    @Test
+    void givenAsyncMethodHasNoException_whenAsyncMethodIscalled_thenReturnSuccess_WithoutAnyRetry() throws InterruptedException, ExecutionException {
+        LOGGER.info("Testing for async with retry execution with thread " + Thread.currentThread().getName());
+        when(downstreamService.publishEvents(anyList())).thenReturn(true);
+        Future<String> resultFuture = asyncEventService.processEvents(List.of("test1"));
+
+        while (!resultFuture.isDone() && !resultFuture.isCancelled()) {
+        }
```
ashleyfrieze, line 41: I suggest putting a `Thread.sleep(5)` in this loop, otherwise the test's own thread will be whizzing round burning up CPU units and slowing down the system. (in all of these `whiles`)

spring-scheduling/src/test/java/com/baeldung/asyncwithretry/EventServiceIntegrationTest.java
```diff
@@ -0,0 +1,90 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(EventServiceIntegrationTest.class);
+
+    @Autowired
+    private EventService asyncEventService;
+
+    @MockBean
+    private DownstreamService downstreamService;
+
+    @Test
+    void givenAsyncMethodHasNoException_whenAsyncMethodIscalled_thenReturnSuccess_WithoutAnyRetry() throws InterruptedException, ExecutionException {
```
ashleyfrieze, line 35: We don't need to throw multiple checked exceptions here - just use `throws Exception` from any test

spring-scheduling/src/test/java/com/baeldung/asyncwithretry/EventServiceIntegrationTest.java
saikatcse03: Updated that with the more readable version of Thread.sleep() -> TimeUnit.SECONDS.sleep() which does the same thing.

spring-scheduling/src/test/java/com/baeldung/asyncwithretry/EventServiceIntegrationTest.java
saikatcse03: sure, i've updaed with Exception where required multiples exception type.

spring-scheduling/src/main/java/com/baeldung/asyncwithretry/DownstreamService.java
```diff
@@ -0,0 +1,10 @@
+package com.baeldung.asyncwithretry;
+
+import org.springframework.stereotype.Service;
+
+import java.util.List;
+
+@Service
+public interface DownstreamService{
```
ashleyfrieze, line 8: Missing ` ` before `{`

saikatcse03: fixed

spring-scheduling/src/test/java/com/baeldung/asyncwithretry/EventServiceIntegrationTest.java
```diff
@@ -0,0 +1,94 @@
...
+    private DownstreamService downstreamService;
+
+    @Test
+    void givenAsyncMethodHasNoException_whenAsyncMethodIscalled_thenReturnSuccess_WithoutAnyRetry() throws Exception {
+        LOGGER.info("Testing for async with retry execution with thread " + Thread.currentThread().getName());
+        when(downstreamService.publishEvents(anyList())).thenReturn(true);
+        Future<String> resultFuture = asyncEventService.processEvents(List.of("test1"));
+
+        while (!resultFuture.isDone() && !resultFuture.isCancelled()) {
+            TimeUnit.SECONDS.sleep(5);
```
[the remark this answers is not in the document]
saikatcse03: Ok, update to millis

PR #15154, opened 2023-11-06
BAEL-6600 - limit max size by LinkedHashMap and custom HashMap

mcasari:
Add a class and two test use cases for limiting HashMap max size, by LinkedHashMap  and HashMap


core-java-modules/core-java-collections-maps-7/src/main/java/com/baeldung/map/HashMapWithMaxSizeLimit.java
```diff
@@ -0,0 +1,25 @@
...
+    private int maxSize = -1;
+    
+   
+    public HashMapWithMaxSizeLimit(int maxSize) {
+        super();
+        this.maxSize = maxSize;
+    }
+
+
+    public V putWithLimit(K key, V value) throws Exception {
```
Maiklins, line 18: It's better to call this method `put` and then `return super.put(key, value);`. Otherwise, one can still call the "normal" put method of the super call and circumvent the limit.

core-java-modules/core-java-collections-maps-7/src/main/java/com/baeldung/map/HashMapWithMaxSizeLimit.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.HashMap;
+
+public class HashMapWithMaxSizeLimit<K, V> extends HashMap<K, V> {
+
+    private static final long serialVersionUID = 1L;
+
+    private int maxSize = -1;
+    
+   
+    public HashMapWithMaxSizeLimit(int maxSize) {
```
Maiklins, line 12: Should values smaller than -1 and 0 be allowed here?

mcasari: No, the -1 is used to define an implicit "no limit" default. I forgot to write some explanation in the article, I did it now.

core-java-modules/core-java-collections-maps-7/src/main/java/com/baeldung/map/HashMapWithMaxSizeLimit.java
mcasari: I chose a pushWihLimit because I couldn't make put to throw Exception. Now I realized that I can use unchecked exceptions instead. So, in the new commit, I simply overrided put and as you said, call super.put(key, value).

mcasari:
I fixed the logic of put implementation. I grouped all the conditions for default behavior in a single if and then raise exception if those are not met.

mcasari:
Added some test cases

core-java-modules/core-java-collections-maps-7/src/test/java/com/baeldung/map/LimitMaxSizeHashMapByLinkedHashMapUnitTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+class LimitMaxSizeHashMapByLinkedHashMapUnitTest {
+
+    private final int MAX_SIZE = 4;
+    private LinkedHashMap<Integer, String> linkedHashMap;
+
+    @BeforeEach
+    void setUp() {
+        linkedHashMap = new LinkedHashMap<Integer, String>() {
+            private static final long serialVersionUID = 1L;
```
KevinGilmore, line 19: This can be omitted, as per guidelines.

core-java-modules/core-java-collections-maps-7/src/test/java/com/baeldung/map/LimitMaxSizeHashMapByLinkedHashMapUnitTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class LimitMaxSizeHashMapByLinkedHashMapUnitTest {
+
+    private final int MAX_SIZE = 4;
+    private LinkedHashMap<Integer, String> linkedHashMap;
+
+    @BeforeEach
```
KevinGilmore, line 16: Since there's only one test method, we could just move all this logic into the test.

core-java-modules/core-java-collections-maps-7/src/test/java/com/baeldung/map/LimitMaxSizeHashMapByLinkedHashMapUnitTest.java
```diff
@@ -0,0 +1,43 @@
...
+    void setUp() {
+        linkedHashMap = new LinkedHashMap<Integer, String>() {
+            private static final long serialVersionUID = 1L;
+
+            protected boolean removeEldestEntry(Map.Entry<Integer, String> eldest) {
```
KevinGilmore, line 21: Add the @Override annotation.

core-java-modules/core-java-collections-maps-7/src/main/java/com/baeldung/map/HashMapWithMaxSizeLimit.java
```diff
@@ -0,0 +1,28 @@
+package com.baeldung.map;
+
+import java.util.HashMap;
+
+public class HashMapWithMaxSizeLimit<K, V> extends HashMap<K, V> {
+
+    private static final long serialVersionUID = 1L;
```
KevinGilmore, line 7: Omit as per guidelines

core-java-modules/core-java-collections-maps-7/src/test/java/com/baeldung/map/LimitMaxSizeHashMapByLinkedHashMapUnitTest.java
mcasari: Ok, I fixed this in a new commit

core-java-modules/core-java-collections-maps-7/src/test/java/com/baeldung/map/LimitMaxSizeHashMapByLinkedHashMapUnitTest.java
mcasari: Ok, I fixed this in a new commit

core-java-modules/core-java-collections-maps-7/src/test/java/com/baeldung/map/LimitMaxSizeHashMapByLinkedHashMapUnitTest.java
mcasari: Ok, I fixed this in a new commit

core-java-modules/core-java-collections-maps-7/src/main/java/com/baeldung/map/HashMapWithMaxSizeLimit.java
mcasari: Ok, I fixed this in a new commit

mcasari:
I pushed a commit but for some reason one of the Jenkins' jobs got stuck.

Maiklins:
retest this please

PR #15206, opened 2023-11-12
JAVA-26733 Upgrade wiremock to the latest version

timis1:


apache-httpclient/pom.xml
anastasiosioannidis, line 1: On another issue, we configured this module to have ONLY apache-httpclient 5 version.

Could you leave the version 5 and only upgrade wiremock?

apache-httpclient/pom.xml
```diff
@@ -63,6 +63,11 @@
             <version>${wiremock.version}</version>
             <scope>test</scope>
         </dependency>
```
anastasiosioannidis, line 65: So, remove this

apache-httpclient/pom.xml
```diff
@@ -77,11 +82,12 @@
 
     <properties>
         <mockserver.version>5.6.1</mockserver.version>
-        <wiremock.version>2.5.1</wiremock.version>
+        <wiremock.version>3.0.1</wiremock.version>
         <!-- http client & core 5 -->
         <httpcore5.version>5.2</httpcore5.version>
         <httpclient5.version>5.2</httpclient5.version>
         <httpclient5-fluent.version>5.2</httpclient5-fluent.version>
```
anastasiosioannidis, line 89: And remove that also

apache-httpclient/pom.xml
timis1: I added `httpclient` in the pom.xml because was needed for this test: HttpAsyncClientLiveTest#whenUseSSLWithHttpAsyncClient_thenCorrect to set `setHostnameVerifier`

apache-httpclient/pom.xml
timis1: Do you refer to remove this dependency?
`<dependency>
            <groupId>org.apache.httpcomponents</groupId>
            <artifactId>httpclient</artifactId>
            <version>${httpclient.version}</version>
        </dependency>`

apache-httpclient/pom.xml
anastasiosioannidis: Ok I understand that here the task is to upgrade wiremock, not handle httpclient upgrade (should be done in past issue)
Please leave it there - I will open another jira to fix upgrading apache-httpclient to v5 (removing v4 imports)

apache-httpclient/pom.xml
anastasiosioannidis: Yes - But you can leave it

PR #15225, opened 2023-11-16
[BAEL-7067] Difference Between Entity and DTO

ovidiumihaitacu:


spring-boot-modules/spring-boot-crud/src/main/java/com/baeldung/entity_dto_differences/Application.java
```diff
@@ -0,0 +1,13 @@
+package com.baeldung.entity_dto_differences;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+@SpringBootApplication
+public class Application {
+
+    public static void main(String[] args) {
+
```
theangrydev, line 10: Can we please drop these extra blank lines at the start of methods, and same thing in all the other classes. If you use the standard Baeldung formatter, I believe it should fix code style issues like this automatically, to match the Baeldung guidelines.

spring-boot-modules/spring-boot-crud/src/main/java/com/baeldung/entity_dto_differences/Application.java
```diff
@@ -0,0 +1,13 @@
+package com.baeldung.entity_dto_differences;
```
theangrydev, line 1: Java package names should not have underscores in them. We follow these guidelines for naming:
https://google.github.io/styleguide/javaguide.html#s5.2.1-package-names

spring-boot-modules/spring-boot-crud/src/main/java/com/baeldung/entity_dto_differences/dto/BookDto.java
```diff
@@ -0,0 +1,35 @@
+package com.baeldung.entity_dto_differences.dto;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import javax.validation.constraints.NotNull;
+
+public class BookDto {
+
+    @NotNull
+    @JsonProperty("NAME")
```
theangrydev, line 10: These custom JSON field names are untested. In your integration test, you should assert on the string and not reuse an ObjectMapper. I can edit this to @JsonProperty("WRONG") and none of the tests fail.

spring-boot-modules/spring-boot-crud/src/main/java/com/baeldung/entity_dto_differences/dto/BookDto.java
```diff
@@ -0,0 +1,35 @@
...
+    private final String name;
+
+    @NotNull
+    @JsonProperty("AUTHOR")
+    private final String author;
+
+    // Default constructor for Jackson deserialization
```
theangrydev, line 17: If you use a `@JsonCreator` constructor and move the `@JsonProperty` annotations into that constructor, you can avoid the ugly default constructor. Just an FYI, you don't have to do this if you don't feel like it.

spring-boot-modules/spring-boot-crud/src/main/java/com/baeldung/entity_dto_differences/dto/BookDto.java
```diff
@@ -0,0 +1,35 @@
...
+    public BookDto() {
+        this.name = null;
+        this.author = null;
```
theangrydev, line 20: Typically I would expect a constructor like this to call `this(null, null)`. Just an FYI, you don't have to do this if you don't feel like it.

spring-boot-modules/spring-boot-crud/src/main/java/com/baeldung/entity_dto_differences/dto/UserCreationDto.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public String getFirstName() {
+        return firstName;
+    }
+
+    public String getLastName() {
+        return lastName;
+    }
+
+    public String getAddress() { return address; }
```
theangrydev, line 49: Fix the formatting. Again, if you use the standard formatter you can just auto format not have to worry about the code guidelines.

spring-boot-modules/spring-boot-crud/src/main/java/com/baeldung/entity_dto_differences/entity/Book.java
```diff
@@ -0,0 +1,38 @@
...
+
+@Entity
+@Table(name = "books")
+public class Book {
+
+    @Id
+    private String name;
+    private String author;
+
+    public Book() {
```
theangrydev, line 15: Since you left a comment about the reason for the default constructor in the other classes, I think it makes sense to do the same here too.

spring-boot-modules/spring-boot-crud/src/test/java/com/baeldung/entity_dto_differences/controller/UserControllerIntegrationTest.java
```diff
@@ -0,0 +1,92 @@
...
+    private ObjectMapper objectMapper;
+
+    @BeforeEach
+    public void beforeEach() {
+
+        userRepository.deleteAll();
+    }
+
+    @Test
+    public void whenGetUsers_thenCorrect() throws Exception {
```
theangrydev, line 44: Your test names do not match what is going on in the tests. If you have a given/when/then section in the test body, I would expect the test name to reflect those parts too.

spring-boot-modules/spring-boot-crud/src/test/java/com/baeldung/entity_dto_differences/controller/UserControllerIntegrationTest.java
```diff
@@ -0,0 +1,92 @@
...
+                .andReturn();
+        UserResponseDto userResponseDto = objectMapper.readValue(result.getResponse().getContentAsString(), UserResponseDto.class);
+
+        //then
+        assertEquals(userCreationDto.getFirstName(), userResponseDto.getFirstName());
+        assertEquals(userCreationDto.getLastName(), userResponseDto.getLastName());
+    }
+
+    // Helper method to convert objects to JSON
+    private String asJsonString(final Object obj) {
```
theangrydev, line 85: Tests like this should not reuse the object mapper. We need hard coded assertions in order to actually cover e.g. the JSON field names.

spring-boot-modules/spring-boot-crud/src/test/java/com/baeldung/entity_dto_differences/entity/BookUnitTest.java
```diff
@@ -0,0 +1,30 @@
+package com.baeldung.entity_dto_differences.entity;
+
+import org.junit.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+public class BookUnitTest {
```
theangrydev, line 8: I would not typically expect to see unit tests for getters and setters. Since you already wrote them, we can keep them if you want, your choice. Typically getters and setters would already be indirectly tested via their usage in other tests.

spring-boot-modules/spring-boot-crud/src/test/java/com/baeldung/entity_dto_differences/entity/UserUnitTest.java
```diff
@@ -0,0 +1,65 @@
...
+        Book book2 = new Book("Book2", "Author2");
+        Book book3 = new Book("Book2", "Author3");
+
+        User user = new User("John", "Doe", "123 Main St", Arrays.asList(book1, book2, book3));
+
+        assertEquals("Book2", user.getNameOfMostOwnedBook());
+    }
+
+    @Test
+    public void whenGetNameOfMostOwnedBook_withNoBooks_thenCorrect() {
```
theangrydev, line 59: The format we require for test names is given_when_then if there are givens, or when_then if there are no givens

This for example should be named something like `givenUserWithNoBooks_whenGetNameOfMostOwnedBook_thenReturnNull`

I am also not a fan of this `thenCorrect` naming, it does not make it clear what "correct" means. Let's be specific in what our "then" expectations are please. Same in all your other tests.

theangrydev (changes requested):

spring-boot-modules/spring-boot-crud/src/main/java/com/baeldung/entity_dto_differences/Application.java
ovidiumihaitacu: I'm using the formatter, but it doesn't delete these newlines. I've removed them manually

spring-boot-modules/spring-boot-crud/src/test/java/com/baeldung/entityDtoDifferences/controller/UserControllerIntegrationTest.java
```diff
@@ -0,0 +1,81 @@
...
+    @Test
+    public void givenValidUser_whenPostUser_thenUserIsCreated() throws Exception {
+        //given
+        String user = "{\"FIRST_NAME\":\"John\",\"LAST_NAME\":\"Doe\",\"BOOKS\":[{\"NAME\":\"Book1\",\"AUTHOR\":\"Author1\"},{\"NAME\":\"Book2\",\"AUTHOR\":\"Author2\"}]}";
+        //when
+        MvcResult result = mockMvc.perform(MockMvcRequestBuilders.post("/users")
+                        .content(user)
+                        .contentType(MediaType.APPLICATION_JSON)
+                        .accept(MediaType.APPLICATION_JSON))
+                .andReturn();
```
theangrydev, line 68: This still does not look like our standard formatter, which you can find here:

theangrydev (changes requested):

spring-boot-modules/spring-boot-crud/src/test/java/com/baeldung/entityDtoDifferences/controller/UserControllerIntegrationTest.java
ovidiumihaitacu: I'm using the formatter ... I've modified the indentation for this specific code block and formatted the code ... so this is the output of our standard formatter

theangrydev: @ovidiumihaitacu which IDE are you using? Here is what I see when I apply the Baeldung formatter to this class:

Screenshot 2023-11-30 at 09 36 47

Screenshot 2023-11-30 at 09 38 42

ovidiumihaitacu: @theangrydev Same as you, Intellij. This is my config:
image

I reran the format and this is the outcome (now it's different than before, it formatted the code to have one-liners):
image

It's strange...

theangrydev (approved):

spring-boot-modules/spring-boot-crud/src/main/java/com/baeldung/entitydtodifferences/dto/BookDto.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @JsonProperty("NAME")
+    private final String name;
+
+    @JsonProperty("AUTHOR")
+    private final String author;
+
+    // Default constructor for Jackson deserialization
+    public BookDto() {
+        this(null, null);
```
KevinGilmore, line 15: Let's omit this line. It is the default behavior of Java objects to initialize their Object fields to null.

spring-boot-modules/spring-boot-crud/src/main/java/com/baeldung/entitydtodifferences/dto/UserCreationDto.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @JsonProperty("ADDRESS")
+    private final String address;
+
+    @JsonProperty("BOOKS")
+    private final List<BookDto> books;
+
+    // Default constructor for Jackson deserialization
+    public UserCreationDto() {
+        this(null, null, null, null);
```
KevinGilmore, line 23: Omit this line.

PR #15270, opened 2023-11-22
[GST-4] Timefold Solver article

ge0ffrey:


timefold-solver/src/main/java/com/baeldung/timefoldsolver/Employee.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public String toString() {
+        return name;
+    }
+
+    // ************************************************************************
+    // Getters and setters
+    // ************************************************************************
+
```
theangrydev, line 23: Remove these comment blocks, I don't remember seeing this style of comments in other articles so it would be a bit out of place
```suggestion
```

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftSchedule.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public ShiftSchedule(List<Employee> employees, List<Shift> shifts) {
+        this.employees = employees;
+        this.shifts = shifts;
+    }
+
+    // ************************************************************************
+    // Getters and setters
+    // ************************************************************************
+
```
theangrydev, line 34: Remove these comment blocks, I don't remember seeing this style of comments in other articles so it would be a bit out of place
```suggestion
```

timefold-solver/src/main/java/com/baeldung/timefoldsolver/Shift.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public String toString() {
+        return start + " - " + end;
+    }
+
+    // ************************************************************************
+    // Getters and setters
+    // ************************************************************************
+
```
theangrydev, line 40: Remove these comment blocks, I don't remember seeing this style of comments in other articles so it would be a bit out of place
```suggestion
```

timefold-solver/src/main/java/com/baeldung/timefoldsolver/Shift.java
```diff
@@ -0,0 +1,57 @@
...
+public class Shift {
+
+    private LocalDateTime start;
+    private LocalDateTime end;
+    private String requiredSkill;
+
+    @PlanningVariable
+    private Employee employee;
+
+    // A no-arg constructor is required for @PlanningEntity annotated classes
```
theangrydev, line 18: Out of curiousity, why is it that @PlanningEntity classes need a no-arg constructor, but @PlanningVariable types e.g. Employee seem to be fine without one?

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleApp.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.List;
+import java.util.Set;
+
+import ai.timefold.solver.core.api.solver.Solver;
+import ai.timefold.solver.core.api.solver.SolverFactory;
+import ai.timefold.solver.core.config.solver.SolverConfig;
+
+public class ShiftScheduleApp {
+
+    public static void main(String[] args) {
```
theangrydev, line 14: Let's replace this main method with a unit test, I don't think we encourage main methods in examples, and prefer tests so we have a way to detect regressions in future

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProvider.java
```diff
@@ -0,0 +1,37 @@
...
+import ai.timefold.solver.core.api.score.stream.Constraint;
+import ai.timefold.solver.core.api.score.stream.ConstraintFactory;
+import ai.timefold.solver.core.api.score.stream.ConstraintProvider;
+
+import static ai.timefold.solver.core.api.score.stream.Joiners.equal;
+
+public class ShiftScheduleConstraintProvider implements ConstraintProvider {
+
+    @Override
+    public Constraint[] defineConstraints(ConstraintFactory constraintFactory) {
```
theangrydev, line 13: Some feedback on the ConstraintProvider API, it's unusual to use array types in a modern library API, I would have expected a List<Constraint> instead. But I appreciate this is something that is not an easy thing to change if you have an existing userbase.

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleApp.java
```diff
@@ -0,0 +1,54 @@
...
+                new Shift(monday.atTime(6, 0), monday.atTime(14, 0), "Waiter"),
+                new Shift(monday.atTime(9, 0), monday.atTime(17, 0), "Bartender"),
+                new Shift(monday.atTime(14, 0), monday.atTime(22, 0), "Bartender"),
+                new Shift(tuesday.atTime(6, 0), tuesday.atTime(14, 0), "Waiter"),
+                new Shift(tuesday.atTime(14, 0), tuesday.atTime(22, 0), "Bartender")
+        ));
+    }
+
+    private static void printSolution(ShiftSchedule solution) {
+        System.out.println("Shift assignments");
```
theangrydev, line 46: If you keep print statements, there is a guideline to use SLF4J with logback, rather than System.out

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProvider.java
```diff
@@ -0,0 +1,37 @@
...
+                requiredSkill(constraintFactory)
+        };
+    }
+
+    public Constraint atMostOneShiftPerDay(ConstraintFactory constraintFactory) {
+        return constraintFactory.forEach(Shift.class)
+                .join(Shift.class,
+                        equal(shift -> shift.getStart().toLocalDate()),
+                        equal(Shift::getEmployee))
+                .filter((shift1, shift2) -> shift1 != shift2)
```
theangrydev, line 25: Is it safe to use reference equality? I'm guessing this means that ConstraintFactory guarantees it will use the exact same instance of a particular shift repeatedly

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Test
+    void atMostOneShiftPerDay() {
+        Employee ann = new Employee("Ann", null);
+        constraintVerifier.verifyThat(ShiftScheduleConstraintProvider::atMostOneShiftPerDay)
+                .given(
+                        ann,
+                        new Shift(MONDAY.atTime(6, 0), MONDAY.atTime(14, 0), null, ann),
+                        new Shift(MONDAY.atTime(14, 0), MONDAY.atTime(22, 0), null, ann))
+                // Penalizes both A-B and B-A. To avoid that, use forEachUniquePair() in the constraint instead.
```
theangrydev, line 25: As someone new to this topic, I find this a bit cryptic. Is there a simpler way to explain what you were communicating here?

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
```diff
@@ -0,0 +1,50 @@
...
+class ShiftScheduleConstraintProviderTest {
+
+    private static final LocalDate MONDAY = LocalDate.of(2030, 4, 1);
+    private static final LocalDate TUESDAY = LocalDate.of(2030, 4, 2);
+
+    ConstraintVerifier<ShiftScheduleConstraintProvider, ShiftSchedule> constraintVerifier = ConstraintVerifier.build(
+            new ShiftScheduleConstraintProvider(), ShiftSchedule.class, Shift.class);
+
+    @Test
+    void atMostOneShiftPerDay() {
```
theangrydev, line 18: We have a convention to use the given_when_then naming convention for test names, I think you can find some examples in the style guide

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
```diff
@@ -0,0 +1,50 @@
...
+        Employee ann = new Employee("Ann", null);
+        constraintVerifier.verifyThat(ShiftScheduleConstraintProvider::atMostOneShiftPerDay)
+                .given(
```
theangrydev, line 21: The formatting/indentation looks like you are not using the formatter profile from the style guide, let me know if you need any help setting that up

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
```diff
@@ -0,0 +1,50 @@
+package com.baeldung.timefoldsolver;
+
+import java.time.LocalDate;
+import java.util.Set;
+
+import ai.timefold.solver.test.api.score.stream.ConstraintVerifier;
```
theangrydev, line 6: Could you explain a bit, or provide a documentation link to the purpose of the ai.timefold.solver.test package helpers?

What does the typical intented development process look like for someone writing a solver with corresponding tests?

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleApp.java
```diff
@@ -0,0 +1,54 @@
...
+public class ShiftScheduleApp {
+
+    public static void main(String[] args) {
+        SolverFactory<ShiftSchedule> solverFactory = SolverFactory.create(new SolverConfig()
+                .withSolutionClass(ShiftSchedule.class)
+                .withEntityClasses(Shift.class)
+                .withConstraintProviderClass(ShiftScheduleConstraintProvider.class)
+                // The solver runs only for 5 seconds on this small dataset.
+                // It's recommended to run for at least 5 minutes ("5m") otherwise.
+                .withTerminationSpentLimit(Duration.ofSeconds(5)));
```
theangrydev, line 21: Given a long enough time spent, is the solution deterministic?

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftSchedule.java
```diff
@@ -0,0 +1,47 @@
...
+@PlanningSolution
+public class ShiftSchedule {
+
+    @ValueRangeProvider
+    private List<Employee> employees;
+    @PlanningEntityCollectionProperty
+    private List<Shift> shifts;
+
+    @PlanningScore
+    private HardSoftScore score = null;
```
theangrydev, line 20:
```suggestion
    private HardSoftScore score;
```

theangrydev (changes requested):

timefold-solver/src/main/java/com/baeldung/timefoldsolver/Shift.java
ge0ffrey: Because internally Timefold Solver needs to "planning clone" the entity and solution classes to keep track of the best solution (so far) while the working solution changes to find a better one.
The other classes (Employee) don't change during planning, so there is no need to do such magic on them.

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleApp.java
ge0ffrey: Yes, absolutely.
2 runs for the same input will yield the same best score at the same iteration (which is at roughly the same time)

Reproducibility is worth it's weight in gold when debugging NullPointerExceptions in a user's filter predicate :)

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
ge0ffrey: Here's the documentation: https://timefold.ai/docs/timefold-solver/latest/constraints-and-score/score-calculation#constraintStreamsTestingIsolatedConstraints

Would you like that in this PR somehow? Or in the article?

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProvider.java
ge0ffrey: Thanks for the feedback.
It's intentional design, at least at the time, because the array of constraints never changes size. And now we're stuck with it for Backwards Compatibility for now.

It looks old indeed :/

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleApp.java
ge0ffrey: With pleasure.

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProvider.java
ge0ffrey: Timefold Solver internally guarantees that 1 solution with one set of shifts will be used in 1 score director (= repeated score calculations). This mean that something like `same(Shift::getEmployee)` would work (if such an operation would exist). But `toLocalDate()` doesn't guarantee the same instance on each call, so it's better to use equals semantics.

ge0ffrey: It's unsafe to use equality that depends on the entire state of the object if the object is a planning entity.
So something like `equal(Shift::getDependentShift)` with Shift doing equality based on `Shift.employee` too (which changes during planning) would cause havoc. But Shift doing equality based on a ID, or even start/end datetimes is fine.

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
ge0ffrey: Yea, I struggle with explaining this one :(
Basically, the score penalty is 2 instead of 1 because it penalizes both the tuple {shift A, shift B} and the tuple {shift B, shift A}. The fix entails adding PlanningId annotations which I didn't want to bring into scope.

Let me try rewording that.

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProvider.java
ge0ffrey: Generally speaking, this just works. Almost no users run into trouble with how it works out of the box. So I wouldn't go into trying to explain the nitty gritty in the article here.

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
ge0ffrey: What do you think of this?

```
                // Penalizes by 2 because both {shiftA, shiftB} and {shiftB, shiftA} match.
                // To avoid that, use forEachUniquePair() in the constraint instead.
```

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
ge0ffrey: Changed article text to

The test-scoped timefold-solver-test dependency contains ConstraintVerifier, a helper to test each constraint in isolation. This improves maintenance: we can refactor a single constraint without breaking tests of other constraints.

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
ge0ffrey: Done.

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
ge0ffrey: No problem:

whenTwoShiftsOnOneDay_thenPenalize
whenTwoShiftsOnDifferentDays_thenDoNotPenalize
whenEmployeeLacksRequiredSkill_thenPenalize
whenEmployeeHasRequiredSkill_thenDoNotPenalize

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleApp.java
ge0ffrey: If we replace this with a unit test as it, it will take 5 seconds from every CI build.

We should probably avoid that by changing

`.withTerminationSpentLimit(Duration.ofSeconds(5))`

to

`.withTerminationConfig(new TerminationConfig().withBestScoreLimit("0hard/0soft"))`

that will bring it down to 74milliseconds on my machine.

ge0ffrey: In the article, I recommend against making that change. This only works if we know the score that the dataset will converge too (which is normally a catch 22 situation).

ge0ffrey: I have not made this change yet. Let me know what you prefer :)

ge0ffrey:
Ready for review again.

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleApp.java
theangrydev: > In the article, I recommend against making that change. 

Where is that in the article sorry? I missed it

And for my understanding, are we saying that `withTerminationSpentLimit` is more like "runs for exactly this amount of time" rather than "runs for at most this amount of time, exiting early if the most optimal solution has been found"?

It's not clear to me why it should take the full 5 seconds for such a small dataset, surely even a brute force approach would complete in less than this time? What am I missing here, does the solver do something monte-carlo like and not realise if it has exhausted all the possibilities already?

Also can you explain what `withBestScoreLimit` means, is this like putting an "exit early" condition in, when we reach some acceptable target metric threshold? 

Sorry for all the questions! I intentionally did not look through the documentation, I wanted fresh eyes to help come up with the kinds of questions a reader new to this might too :)

timefold-solver/src/main/java/com/baeldung/timefoldsolver/Shift.java
theangrydev: PS: you might be able to use bytecode manipulation tools such as https://bytebuddy.net/ to remove the need for a no-arg constructor. Out of the scope of this article, but could help make your user facing APIs a bit more concise.

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProvider.java
theangrydev: So to summarise, are we saying that:
1. We can rely on the base PlanningEntity (Shift in this case) as being reference-stable for the purpose of these APIs
2. But we cannot rely on any PlanningEntity fields (e.g. nested Shift, or reference to another PlanningEntity) to be reference-stable

The reason that 1. was confusing to me is that, in the wider Java world, it's very unusual to see an intentional use of reference equality for non primitive types. Usually a line like this would be an unintentional bug, and it's only because you explicitly designed for it that it's ok.

2. Makes sense, I suppose this means that one limitation of the solver is that you can't have constraints that create cyclic dependencies with the rest of the solver state, if I'm understanding correctly. And instead we impose constraints that partition the solution space, but do not introduce cycles. Let me know if I'm misunderstanding anything here.

> Generally speaking, this just works. Almost no users run into trouble with how it works out of the box. So I wouldn't go into trying to explain the nitty gritty in the article here.

Yes I agree not to clutter the article with these details, just making sure I'm understanding what's going on exactly.

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
theangrydev: >  // Penalizes by 2 because both {shiftA, shiftB} and {shiftB, shiftA} match.
Makes sense

> // To avoid that, use forEachUniquePair() in the constraint instead.
Meaning, forEachUniquePair will consider {shiftA, shiftB} as equivalent to {shiftB, shiftA}?

It would help me if we could say "To avoid that, use forEachUniquePair() in the constraint instead of ___ in the constraint instead", it wasn't clear which part of the constraint forEachUniquePair would replace

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
theangrydev: My recommendation:

whenTwoShiftsOnOneDay_thenPenalize
-> givenTwoShiftsOnOneDay_whenApplyingAtMostOneShiftPerDayConstraint_thenPenalizeTwice

whenTwoShiftsOnDifferentDays_thenDoNotPenalize
-> givenTwoShiftsOnDifferentDays_whenApplyingAtMostOneShiftPerDayConstraint_thenDoNotPenalize

whenEmployeeLacksRequiredSkill_thenPenalize
-> givenEmployeeLacksRequiredSkill_whenApplyingRequiredSkillConstraint_thenPenalize

whenEmployeeHasRequiredSkill_thenDoNotPenalize
-> givenEmployeeHasRequiredSkill_whenApplyingRequiredSkillConstraint_thenDoNotPenalize

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
theangrydev: >Would you like that in this PR somehow? Or in the article?

Sure let's leave a link in the article if you feel it makes sense.

> The test-scoped timefold-solver-test dependency contains ConstraintVerifier, a helper to test each constraint in isolation. This improves maintenance: we can refactor a single constraint without breaking tests of other constraints.

I feel like this is a pretty useful feature to point out from a design perspective. Constraint design seems like the a critical part of the configuration, and I'm glad to see there's a supported way to help assist that design process by testing in isolation, good stuff!

theangrydev:
I did not look into why, but the CI check failed here

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleApp.java
ge0ffrey: Very good point, I 've brought it down to 2 second and added this in the article which is a practical get started approach to that complexity:

> We also need to configure how long we want a solver to run. For large datasets, with thousands of shifts and far more constraints, it's impossible to find the optimal solution in reasonable time (due to the exponential nature of [NP-hard](https://en.wikipedia.org/wiki/NP-hardness) problems). Instead, we want to find the best possible solution in the amount of time available. Let's limit that to 2 seconds for now:

I think a discussion of brute force scalability, NP-hard theory, etc is out of scope for this article.

ge0ffrey: We do have a 3 hour video explaining that, in case we want to link that in the article :)
  https://youtu.be/FLaYKlLWKW4?si=GavSGoPIZ_dl91Vy&t=617
(currently not linked)

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
ge0ffrey: Thanks, I am happy to see appreciation and understanding of the grander design :) Article adjusted.

I 've added a general link to the documention at the bottom of the article:
  For more information, see [the Timefold Solver documentation](https://timefold.ai/docs/timefold-solver/latest/).
That link will never go stale.

ge0ffrey:
CI failed due to PMD and I fixed that.

ge0ffrey:
As far as I can tell, the only open item is to change ShiftScheduleApp into a unit test.

Proposal A) Leave as is.

Proposal B) Change to UnitTest, but leave it to take 2 seconds.
  More in sync with article.
  
Proposal C) Change to UnitTest and use best score limit, so it takes a few milliseconds.
  Not in sync with article.
  
Proposal D) Change to UnitTest and use best score limit, and adjust article to do that too.

My opinion:

- Hard -1 on D). I don't want to teach users something that will bite them when they use a real dataset.

- No preference on A) vs B) vs C). Happy to adjust the PR to what you prefer.

timefold-solver/src/main/java/com/baeldung/timefoldsolver/ShiftScheduleApp.java
theangrydev: > Very good point, I 've brought it down to 2 second and added this in the article which is a practical get started approach to that complexity:

Makes sense, looks good

> I think a discussion of brute force scalability, NP-hard theory, etc is out of scope for this article.

Agreed!

> We do have a 3 hour video explaining that, in case we want to link that in the article :)

Got it, will keep it in mind and let you know

theangrydev:
> As far as I can tell, the only open item is to change ShiftScheduleApp into a unit test.
> 
> Proposal A) Leave as is.
> 
> Proposal B) Change to UnitTest, but leave it to take 2 seconds. More in sync with article.
> 
> Proposal C) Change to UnitTest and use best score limit, so it takes a few milliseconds. Not in sync with article.
> 
> Proposal D) Change to UnitTest and use best score limit, and adjust article to do that too.
> 
> My opinion:
> 
> * Hard -1 on D). I don't want to teach users something that will bite them when they use a real dataset.
> * +0 on A), B) and C). Happy to adjust the PR to the proposal you believe is best.

😆 I like how you are now tuned to think in terms constraint numbers now!

I'm inclined to go with B) since (as far as I am aware) we have guidelines against having main methods in PRs

ge0ffrey:
Done :)

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
theangrydev: @ge0ffrey this one I'd like clarified please

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderUnitTest.java
```diff
@@ -0,0 +1,52 @@
...
+    ConstraintVerifier<ShiftScheduleConstraintProvider, ShiftSchedule> constraintVerifier = ConstraintVerifier.build(new ShiftScheduleConstraintProvider(),
+        ShiftSchedule.class, Shift.class);
+
+    @Test
+    void whenTwoShiftsOnOneDay_thenPenalize() {
+        Employee ann = new Employee("Ann", null);
+        constraintVerifier.verifyThat(ShiftScheduleConstraintProvider::atMostOneShiftPerDay)
+            .given(ann, new Shift(MONDAY.atTime(6, 0), MONDAY.atTime(14, 0), null, ann), new Shift(MONDAY.atTime(14, 0), MONDAY.atTime(22, 0), null, ann))
+            // Penalizes by 2 because both {shiftA, shiftB} and {shiftB, shiftA} match.
+            // To avoid that, use forEachUniquePair() in the constraint instead.
```
theangrydev, line 24: Instead of ..?

theangrydev (changes requested):
Just one point about the forEachUniquePair comment to clarify please, about what it is used instead of exactly

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleSolverUnitTest.java
```diff
@@ -0,0 +1,57 @@
...
+        SolverFactory<ShiftSchedule> solverFactory = SolverFactory.create(new SolverConfig().withSolutionClass(ShiftSchedule.class)
+            .withEntityClasses(Shift.class)
+            .withConstraintProviderClass(ShiftScheduleConstraintProvider.class)
+            // The solver runs only for 2 seconds on this tiny dataset.
+            // It's recommended to run for at least 5 minutes ("5m") on large datasets.
+            .withTerminationSpentLimit(Duration.ofSeconds(2)));
+        Solver<ShiftSchedule> solver = solverFactory.buildSolver();
+
+        ShiftSchedule problem = loadProblem();
+        ShiftSchedule solution = solver.solve(problem);
```
theangrydev, line 32: Sorry I missed this too, can we actually assert on the solution, rather than just printing it, if it is deterministic?

ge0ffrey: If we keep `.withTerminationSpentLimit(Duration.ofSeconds(2)));` it is possible that extremely slow working nodes with a large stop-the-world GC might **randomly fail** the unit test. But in the unit test we know the optimal score, so we can safely change it to this:

```
// For this dataset, we know the optimal score in advance (which is normally not the case).
            // So we can use .withBestScoreLimit() instead of .withTerminationSpentLimit().
            .withTerminationConfig(new TerminationConfig().withBestScoreLimit("0hard/0soft")));
            ```

File: timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleSolverUnitTest.java [hunk 17]
```diff
@@ -0,0 +1,63 @@
...
+            // So we can use .withBestScoreLimit() instead of .withTerminationSpentLimit().
+            .withTerminationConfig(new TerminationConfig().withBestScoreLimit("0hard/0soft")));
+        Solver<ShiftSchedule> solver = solverFactory.buildSolver();
+
+        ShiftSchedule problem = loadProblem();
+        ShiftSchedule solution = solver.solve(problem);
+        assertThat(solution.getShifts().size()).isNotZero();
+        for (Shift shift : solution.getShifts()) {
+            assertThat(shift.getEmployee()).isNotNull();
+        }
```
theangrydev, line 38: I was thinking more about asserting on the specific hard coded optimal result data that is expected, rather than just a structural assertion

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleSolverUnitTest.java
```diff
@@ -0,0 +1,63 @@
...
+import ai.timefold.solver.core.api.solver.SolverFactory;
+import ai.timefold.solver.core.config.solver.SolverConfig;
+import ai.timefold.solver.core.config.solver.termination.TerminationConfig;
+
+public class ShiftScheduleSolverUnitTest {
+
+    private static final Logger logger = LoggerFactory.getLogger(ShiftScheduleSolverUnitTest.class);
+
+    @Test
+    public void solve() {
```
theangrydev, line 24: We need to rename this in the given/when/then style to meet the guidelines

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
theangrydev: @ge0ffrey ping here, the current names are not in BDD style, you currently have bits of the given merged into the when

theangrydev (changes requested):
+ see previous comment on BDD test naming

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleSolverUnitTest.java
ge0ffrey: There are multiple optimal solutions, so that would be too strict.
I'll assert the optimal score to unsure it's an optimal result.

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleSolverUnitTest.java
ge0ffrey: Fixed

ge0ffrey:
> Just one point about the forEachUniquePair comment to clarify please, about what it is used instead of exactly

Improved comment

timefold-solver/src/test/java/com/baeldung/timefoldsolver/ShiftScheduleConstraintProviderTest.java
ge0ffrey: Changing
  givenTwoShiftsOnOneDay_whenApplyingAtMostOneShiftPerDayConstraint_thenPenalizeTwice
to
   givenTwoShiftsOnOneDay_whenApplyingAtMostOneShiftPerDayConstraint_thenPenalize

The penalize twice shouldn't be made into the spec. It penalizes, so that's fine. It doesn't matter by how much.

The current impl is good enough: a simple impl that works (even in production), but the unique pair approach (which requires a bunch more code due to the @PlanningId requirement) is more perfect. Perfect is the enemy of good here, for a tutorial.

ge0ffrey: Fixed

theangrydev (approved):

timefold-solver/README.md
```diff
@@ -0,0 +1,7 @@
+## OptaPlanner
+
+This module contains articles about OptaPlanner.
+
+### Relevant articles
+
+- [A Guide to OptaPlanner](https://www.baeldung.com/opta-planner)
```
KevinGilmore, line 7: We can omit this line for now. The editorial team will add it after publishing once they have the final URL.

ge0ffrey: Yes - and the readme needs to talk about Timefold instead of OptaPlanner (copy-paste error). Fixed.

PR #15279, opened 2023-11-23
[negate-int] negate an int in java

sk1418:


core-java-modules/core-java-numbers-conversions/src/test/java/com/baeldung/negate/NegateIntUnitTest.java
```diff
@@ -0,0 +1,62 @@
+package com.baeldung.negate;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class NegateIntUnitTest {
+    @Test
```
acerasoni, line 8: Could we add a space here?

PR #15330, opened 2023-11-30
JAVA-27019 Simplify spring-di modules' poms

timis1:


spring-di-2/pom.xml
anastasiosioannidis, line 1: All good here. Since there's some time left. let's try this:

-Replace parent-spring-6 with parent-boot-3
-Delete spring.boot-version references

spring-di-3/pom.xml
anastasiosioannidis, line 1: All good here. Since there's some time left. let's try this:

-Replace parent-spring-6 with parent-boot-3
-Delete spring.boot-version references

spring-di/pom.xml
anastasiosioannidis, line 1: All good here. Since there's some time left. let's try this:

-Replace parent-spring-6 with parent-boot-3
-Delete spring-context and dependencies like that. These may be brought by spring boot.

We want only spring boot 3 parent to bring all needed dependencies.

Please try leaving as little as possible here

spring-di-2/pom.xml
timis1: No parent-boot-3 is needed.

spring-di-3/pom.xml
timis1: No parent-boot-3 is needed.

spring-di/pom.xml
timis1: We will not use spring-boot-3 anymore for now.

PR #15347, opened 2023-12-02
BAEL-7283: Convert a page of objects to list in Spring Data

azhwani:
Add a method/test case to demonstrate how to convert a page to a list in Spring Data

### DIFF
--- a/persistence-modules/spring-data-jpa-query-3/src/test/java/com/baeldung/spring/data/jpa/paging/CustomerServiceUnitTest.java
+++ b/persistence-modules/spring-data-jpa-query-3/src/test/java/com/baeldung/spring/data/jpa/paging/CustomerServiceUnitTest.java
@@ -1,14 +1,17 @@
 package com.baeldung.spring.data.jpa.paging;
 
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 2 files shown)


persistence-modules/spring-data-jpa-query-3/src/test/java/com/baeldung/spring/data/jpa/paging/CustomerServiceUnitTest.java
```diff
@@ -77,4 +77,14 @@ void givenAListOfCustomers_whenGetCustomers_thenReturnsDesiredDataAlongWithPagin
         assertEquals(expectedTotalElements, customers.getTotalElements());
         assertEquals(expectedTotalPages, customers.getTotalPages());
     }
+
+    void givenAPageOfCustomers_whenGetCustomerList_thenReturnsDesiredListData() {
```
JavierPeris, line 81: We are missing the @Test annotation otherwise the test is not executed.

azhwani: Oops, done

PR #15350, opened 2023-12-03
Final Static Variables in Java

Michaelin007:


core-java-modules/core-java-lang-6/src/main/java/com/baeldung/finalstatic/Bike.java
```diff
@@ -0,0 +1,14 @@
+package com.baeldung.finalstatic;
```
Maiklins, line 1: Let's rename the package to `staticfinal`

Michaelin007: Thanks @Maiklins, resolved.

PR #15391, opened 2023-12-10
BAEL-7280: How to get last record ordered by date in Spring Data

azhwani:
+ Create Post entity class
+ Create Post JPA repository
+ Create test class


dmelikhov:
hi @azhwani, the 3rd method in PostRepository won’t work as JPQL doesn’t support “LIMIT”

> org.hibernate.hql.internal.ast.QuerySyntaxException: unexpected token: LIMIT near line 1

PR #15404, opened 2023-12-12
BAEL-7278 first draft

wynnteo:


pdf-2/src/main/java/com/baeldung/exceltopdf/ExcelToPDFConverter.java
```diff
@@ -0,0 +1,167 @@
...
+        addTableHeader(worksheet, table);
+        addTableData(worksheet, table);
+        document.add(table);
+    }
+
+    private static void addTableHeader(XSSFSheet worksheet, PdfPTable table) throws DocumentException, IOException {
+        Row headerRow = worksheet.getRow(0);
+        for (int i = 0; i < headerRow.getPhysicalNumberOfCells(); i++) {
+            Cell cell = headerRow.getCell(i);
+            String headerText = cell.getCellType() == CellType.STRING ? cell.getStringCellValue() : String.valueOf(cell.getNumericCellValue());
```
Maiklins, line 77: Does this work if the cell type isn't String or Numeric, for example Blank?

pdf-2/src/main/java/com/baeldung/exceltopdf/ExcelToPDFConverter.java
```diff
@@ -0,0 +1,167 @@
...
+                    }
+                }
+
+                cellPdf.setHorizontalAlignment(Element.ALIGN_CENTER);
+                table.addCell(cellPdf);
+            }
+        }
+    }
+
+    private static Font getCellStyle(Cell cell) throws DocumentException, IOException {
```
Maiklins, line 126: The article says that we can preserve the cell style including the font, borders, colors, etc. However, the code has the font and font size hardcoded. Can we either change the code or the description in the article? Otherwise, it's confusing to the reader.

pdf-2/src/main/java/com/baeldung/exceltopdf/ExcelToPDFConverter.java
```diff
@@ -0,0 +1,167 @@
...
+        Iterator<Row> rowIterator = worksheet.iterator();
+        while (rowIterator.hasNext()) {
+            Row row = rowIterator.next();
+            if (row.getRowNum() == 0) {
+                continue;
+            }
+            for (int i = 0; i < row.getPhysicalNumberOfCells(); i++) {
+                Cell cell = row.getCell(i);
+                String cellValue;
+                if (cell != null) {
```
Maiklins, line 94: Can this be null if the for loop has the condition `i < row.getPhysicalNumberOfCells()`?

pdf-2/src/main/java/com/baeldung/exceltopdf/ExcelToPDFConverter.java
wynnteo: Thanks for pointing this out, will fix it to handle blank and other types.

pdf-2/src/main/java/com/baeldung/exceltopdf/ExcelToPDFConverter.java
wynnteo: You're correct, the cell variable within the loop shouldn't be null if the condition i < row.getPhysicalNumberOfCells() is met. I'll remove the unnecessary null check.

pdf-2/src/main/java/com/baeldung/exceltopdf/ExcelToPDFConverter.java
wynnteo: Noted on this, I'll update the code, and in the description will also highlight the limited font family iText support.

pdf-2/src/main/java/com/baeldung/exceltopdf/ExcelToPDFConverter.java
```diff
@@ -74,9 +88,27 @@ private static void addTableHeader(XSSFSheet worksheet, PdfPTable table) throws
...
+                headerText = cell.getStringCellValue();
+                break;
+            case NUMERIC:
+                headerText = String.valueOf(BigDecimal.valueOf(cell.getNumericCellValue()));
+                break;
+            case BLANK:
+                headerText = ""; // or null
+                break;
+            default:
+                logger.warn("Unsupported cell type: {}", cell.getCellType());
```
Maiklins, line 104: Maybe it's worth mentioning in the article that an Excel file that contains formulae isn't fully supported.

wynnteo: Okay noted.

PR #15405, opened 2023-12-12
BAEL-7349 - Introduction to Spring Cloud AWS SQS.

tomazfernandes:
First rough draft of tutorial code.


spring-cloud-modules/spring-cloud-aws-v3/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+    <name>spring-cloud-aws-v3</name>
+    <packaging>jar</packaging>
+    <description>Spring Cloud AWS Examples</description>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>io.awspring.cloud</groupId>
+                <artifactId>spring-cloud-aws</artifactId>
+                <version>3.1.0</version>
```
theangrydev, line 18: We have a convention to keep all dependency versions in a property, can you extract a property for this version number

spring-cloud-modules/spring-cloud-aws-v3/src/main/java/com/baeldung/spring/cloud/aws/sqs/BaeldungListeners.java
```diff
@@ -0,0 +1,34 @@
...
+
+@Component
+public class BaeldungListeners {
+
+    private static final Logger logger = LoggerFactory.getLogger(BaeldungListeners.class);
+
+    public static final String STRING_PAYLOAD_QUEUE_NAME = "string-test-queue";
+    public static final String RECORD_PAYLOAD_QUEUE_NAME = "record-test-queue";
+
+    final CountDownLatchContainer latchContainer;
```
theangrydev, line 16: Keep this private if possible

spring-cloud-modules/spring-cloud-aws-v3/src/test/java/com/baeldung/spring/cloud/aws/sqs/SpringCloudAwsSQSLiveTest.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Test
+    void givenARecordPayload_whenSend_shouldReceive() throws Exception {
+        sqsTemplate.send(to -> to
+                .queue(RECORD_PAYLOAD_QUEUE_NAME)
+                .payload(new BaeldungRecord(1, "My Record")));
+        logger.info("Sent message");
+        assertThat(latchContainer.recordPayloadLatch.await(3, TimeUnit.SECONDS)).isTrue();
+    }
```
theangrydev, line 39: Currently, we are not proving that the expected message payload is received, only that "some message" is received, and doing so indirectly via the knowledge about what these latches are responsible for. 

Can we find a way to test the actual message payload too?

And can you think of a more direct way to test this scenario? I currently have to know about the relationship between CountDownLatchContainer and BaeldungListeners in order to understand what the test is proving. Ideally, it should be clear from looking at the test class alone what exactly is being tested.

spring-cloud-modules/spring-cloud-aws-v3/src/test/java/com/baeldung/spring/cloud/aws/sqs/SpringCloudAwsSQSLiveTest.java
```diff
@@ -0,0 +1,41 @@
...
+        sqsTemplate.send(to -> to.queue(STRING_PAYLOAD_QUEUE_NAME).payload("My String"));
+        logger.info("Sent message");
+        assertThat(latchContainer.stringPayloadLatch.await(3, TimeUnit.SECONDS)).isTrue();
+    }
+
+    @Test
+    void givenARecordPayload_whenSend_shouldReceive() throws Exception {
+        sqsTemplate.send(to -> to
+                .queue(RECORD_PAYLOAD_QUEUE_NAME)
+                .payload(new BaeldungRecord(1, "My Record")));
```
theangrydev, line 36: Based on the way this is currently indented, it looks as if you might not be using the Baeldung formatter, which you can find here:

theangrydev (changes requested):

spring-cloud-modules/spring-cloud-aws-v3/src/test/java/com/baeldung/spring/cloud/aws/sqs/SpringCloudAwsSQSLiveTest.java
tomazfernandes: I added a note on the scenario in the JIRA ticket, please let me know what you think.

If we use a DB approach, we could definitely check what was persisted in the database.

spring-cloud-modules/spring-cloud-aws-v3/pom.xml
tomazfernandes: Just a note - I was having lots of troubles with dependency versions when inheriting the parent project's pom since it uses fairly outdated dependencies, so I let go of that inheritance and created an independent pom.xml. It still compiles as part of the same project though.

Perhaps we could consider having a separate spring cloud module for Spring Boot 3 and adding new tutorials there.

theangrydev: Hmm, yes I'll ask internally about what the best option would be here

spring-cloud-modules/spring-cloud-aws-v3/src/main/java/com/baeldung/spring/cloud/aws/sqs/BaeldungListeners.java
theangrydev: New PR supersedes this discussion I think

spring-cloud-modules/spring-cloud-aws-v3/src/test/java/com/baeldung/spring/cloud/aws/sqs/SpringCloudAwsSQSLiveTest.java
theangrydev: New PR supersedes this discussion I think

spring-cloud-modules/spring-cloud-aws-v3/src/test/java/com/baeldung/spring/cloud/aws/sqs/SpringCloudAwsSQSLiveTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Test
+    void givenARecordPayload_whenSend_shouldReceive() {
+        // given
+        String userId = UUID.randomUUID()
+            .toString();
+        var payload = new UserCreatedEvent(userId, "John", "john@baeldung.com");
+
+        // when
+        sqsTemplate.send(to -> to.queue(eventQueuesProperties.getUserCreatedFifoQueue())
```
theangrydev, line 54: This test does not seem to do anything FIFO specific, but I'm guessing the point was just to show "it's possible to use a FIFO queue" in one of the tests

tomazfernandes: Yeah, that was a last minute addition to this PR but I think it's worth mentioning it.

I plan to explain some basic things about what gets configured automatically for FIFO queues, such as acknowledgement.

theangrydev (approved):

PR #15406, opened 2023-12-12
[BAEL-6064] Enable SSL Debug Logging

anujgaud:


core-java-modules/core-java-security-4/src/main/java/com/baeldung/enablessldebug/SSLDebugLogger.java
```diff
@@ -0,0 +1,7 @@
+package com.baeldung.enablessldebug;
+
+public class SSLDebugLogger {
+    public static void enableSSLDebugUsingSystemProperties() {
+        System.setProperty("javax.net.debug", "ssl:debug");
```
acerasoni, line 5: Would your tests still work if we moved this snippet of code into a private method within the Unit test class? I'm thinking that in that case we could remove this source code class altogether.

At the same time we want to replicate the setup someone would have in their project.

What are your thoughts?

core-java-modules/core-java-security-4/src/test/java/com/baeldung/enablessldebug/SSLDebugLoggerUnitTest.java
```diff
@@ -0,0 +1,31 @@
...
+import java.io.InputStream;
+import java.util.logging.ConsoleHandler;
+import java.util.logging.Level;
+import java.util.logging.LogManager;
+import java.util.logging.Logger;
+
+import org.junit.Test;
+
+public class SSLDebugLoggerUnitTest {
+    @Test
```
acerasoni, line 15: Could we add a space

- Between the top of the class and the first line of code in the class (helps readability)
i.e.
public class SSLDebugLoggerUnitTest {

    @Test
...
}
- Between the test setup and assertions
i.e. 
```
...
 SSLDebugLogger.enableSSLDebugUsingSystemProperties();

 assertEquals("ssl:debug", System.getProperty("javax.net.debug"));
...
```

core-java-modules/core-java-security-4/src/test/java/com/baeldung/enablessldebug/SSLDebugLoggerUnitTest.java
```diff
@@ -0,0 +1,31 @@
...
+    public void givenDebugging_whenSystemProperties_thenEnableSSL() {
+        SSLDebugLogger.enableSSLDebugUsingSystemProperties();
```
acerasoni, line 17: To replicate the dynamic on-off behaviour of this approach, can you modify this test to

- Enable SSL debugging
- Assert it's enabled
- Disable SSL debugging
- Assert it's disabled

core-java-modules/core-java-security-4/src/main/java/com/baeldung/enablessldebug/SSLDebugLogger.java
anujgaud: "At the same time we want to replicate the setup someone would have in their project."
Didn't get this, could you please clarify?

core-java-modules/core-java-security-4/src/test/java/com/baeldung/enablessldebug/SSLDebugLoggerUnitTest.java
anujgaud: Added a new line between class and test. for asserts statement I've grouped them into a logical way followed by a new line. Please review.

core-java-modules/core-java-security-4/src/test/java/com/baeldung/enablessldebug/SSLDebugLoggerUnitTest.java
anujgaud: Done

PR #15465, opened 2023-12-21
https://jira.baeldung.com/browse/BAEL-7376

Michaelin007:


spring-boot-modules/spring-boot-3/src/test/java/com/baeldung/springjunitconfiguration/Person.java
```diff
@@ -0,0 +1,15 @@
+package com.baeldung.springjunitconfiguration;
+
+public class Person {
```
Maiklins, line 3: Can you move this class to the src/main folder instead of src/test folder?

Michaelin007: Hi @Maiklins, I've moved it.

PR #15475, opened 2023-12-24
First version

wynnteo:


spring-kafka/src/main/java/com/baeldung/partitioningstrategy/KafkaApplication.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Bean
+    public ProducerFactory<String, String> producerFactory()
+    {
+        Map<String, Object> configProps = new HashMap<>();
+        configProps.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9092");
+        configProps.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, StringSerializer.class);
+        configProps.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, StringSerializer.class);
+        return new DefaultKafkaProducerFactory<>(configProps);
+    }
```
theangrydev, line 34: This looks as if you have not used the standard Baeldung formatter, can you please apply it to all of the files in the PR:

spring-kafka/src/main/java/com/baeldung/partitioningstrategy/KafkaMessageConsumer.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.kafka.annotation.KafkaListener;
+import org.springframework.messaging.handler.annotation.Header;
+import org.springframework.messaging.handler.annotation.Payload;
+import org.springframework.stereotype.Service;
+import org.springframework.kafka.support.KafkaHeaders;
+import jakarta.annotation.Nullable;
+
+@Service
+public class KafkaMessageConsumer {
+    private final List<ReceivedMessage> receivedMessages = new ArrayList<>();
```
theangrydev, line 14: This is not thread safe, can we use a CopyOnWriteArrayList or similar instead?

Or, do you know if @KafkaListener is guaranteed to be single threaded?

spring-kafka/src/main/java/com/baeldung/partitioningstrategy/KafkaMessageConsumer.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @KafkaListener(topics = {"order-topic", "default-topic"}, groupId = "test-group")
+    public void listen(@Payload String message,
+        @Header(KafkaHeaders.RECEIVED_PARTITION) int partition,
+        @Header(KafkaHeaders.RECEIVED_KEY) @Nullable String key) {
+        ReceivedMessage receivedMessage = new ReceivedMessage(key, message, partition);
+        System.out.println("Received message: " + receivedMessage);
```
theangrydev, line 21: We have a guideline to use SLF4J for all logging

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
```diff
@@ -0,0 +1,164 @@
...
+        kafkaMessageConsumer.clearReceivedMessages();
+    }
+
+    @Test
+    public void givenDefaultPartitioner_whenSendingMessagesWithoutKey_shouldUseStickyDistribution() throws InterruptedException {
+        kafkaProducer.send("default-topic", "message1");
+        kafkaProducer.send("default-topic","message2");
+        kafkaProducer.send("default-topic","message3");
+
+        Thread.sleep(2000);
```
theangrydev, line 52: It's almost never a good idea to use Thead.sleep in a test

If you need to wait for an async condition, use something like:
https://www.baeldung.com/awaitility-testing

To wait on a condition (e.g. the number of messages received), so the test can be deterministic

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
```diff
@@ -0,0 +1,164 @@
...
+    @BeforeEach
+    public void clearMessages() {
+        kafkaMessageConsumer.clearReceivedMessages();
+    }
+
+    @Test
+    public void givenDefaultPartitioner_whenSendingMessagesWithoutKey_shouldUseStickyDistribution() throws InterruptedException {
+        kafkaProducer.send("default-topic", "message1");
+        kafkaProducer.send("default-topic","message2");
+        kafkaProducer.send("default-topic","message3");
```
theangrydev, line 50: Fix formatting

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
```diff
@@ -0,0 +1,164 @@
...
+        Thread.sleep(2000);
+
+        List<ReceivedMessage> records = kafkaMessageConsumer.getReceivedMessages();
+        int expectedPartition = records.get(0).getPartition();
+
+        for (ReceivedMessage record : records) {
+            if (record.getKey() == null) {
+                Assert.assertEquals("Message without key should be in the same partition", expectedPartition, record.getPartition());
+            }
+        }
```
theangrydev, line 61: What is the purpose of the if statement here? Isn't every message sent without a key in this test?

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @Test
+    void givenProducerWithSameKeyMessages_whenSendingMessages_shouldRouteToSamePartition() throws InterruptedException {
+        kafkaProducer.send("order-topic", "partitionA", "critical data");
+        kafkaProducer.send("order-topic", "partitionA", "more critical data");
+        kafkaProducer.send("order-topic", "partitionB", "another critical message");
+        kafkaProducer.send("order-topic", "partitionA", "another more critical data");
+
+        Thread.sleep(1000);
```
theangrydev, line 71: Same point about Thread.sleep

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
```diff
@@ -0,0 +1,164 @@
...
+        });
+    }
+
+    @Test
+    public void givenProducerWithSameKeyMessages_whenSendingMessages_shouldReceiveInProducedOrder() throws InterruptedException {
+        kafkaProducer.send("order-topic", "partitionA", "message1");
+        kafkaProducer.send("order-topic", "partitionA", "message3");
+        kafkaProducer.send("order-topic", "partitionA", "message4");
+
+        Thread.sleep(1000);
```
theangrydev, line 92: Same point about Thread.sleep

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
```diff
@@ -0,0 +1,164 @@
...
+
+    @Test
+    public void givenCustomPartitioner_whenSendingMessages_shouldRouteToCorrectPartition() throws InterruptedException {
+        // Configure the producer with the custom partitioner
+        KafkaTemplate<String, String> kafkaTemplate = setProducerToUseCustomPartitioner();
+
+        kafkaTemplate.send("order-topic", "123_premium", "Order 123, Premium order message");
+        kafkaTemplate.send("order-topic", "456_normal", "Normal order message");
+
+        Thread.sleep(1000);
```
theangrydev, line 115: Same point about Thread.sleep

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @Test
+    public void givenCustomPartitioner_whenSendingMessages_shouldConsumeOnlyFromSpecificPartition() throws InterruptedException {
+        KafkaTemplate<String, String> kafkaTemplate = setProducerToUseCustomPartitioner();
+
+        kafkaTemplate.send("order-topic", "123_premium", "Order 123, Premium order message");
+        kafkaTemplate.send("order-topic", "456_normal", "Normal order message");
+
+        Thread.sleep(1000);
```
theangrydev, line 136: Same point about Thread.sleep

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
```diff
@@ -0,0 +1,164 @@
...
+
+        consumer.assign(Collections.singletonList(new TopicPartition("order-topic", 0)));
+        ConsumerRecords<String, String> records = consumer.poll(Duration.ofMillis(100));
+
+        for (ConsumerRecord<String, String> record : records) {
+            assertEquals("Premium order message should be in partition 0", 0, record.partition());
+            assertEquals("123_premium", record.key());
+        }
+
+        consumer.close();
```
theangrydev, line 146: This close should be done after every test in an `@AfterEeach` - or it might actually already be handled by the `@SpringBootTest` - you can set a breakpoint inside the close method and see if it is already called per test

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
```diff
@@ -0,0 +1,164 @@
...
+    @Autowired
+    private EmbeddedKafkaBroker embeddedKafkaBroker;
+
+    @Autowired
+    private Consumer<String, String> consumer;
+
+    @BeforeEach
+    public void clearMessages() {
+        kafkaMessageConsumer.clearReceivedMessages();
+    }
```
theangrydev, line 44: What is the purpose of this? Isn't the KafkaMessageConsumer already initialized with zero messages on every individual test run?

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
```diff
@@ -0,0 +1,164 @@
...
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.kafka.core.DefaultKafkaProducerFactory;
+import org.springframework.kafka.core.KafkaTemplate;
+import org.springframework.kafka.test.EmbeddedKafkaBroker;
+import org.springframework.kafka.test.context.EmbeddedKafka;
+import org.springframework.kafka.test.utils.KafkaTestUtils;
+
+@SpringBootTest
+@EmbeddedKafka(partitions = 3, brokerProperties = {"listeners=PLAINTEXT://localhost:9092"})
+public class KafkaApplicationUnitTesting {
```
theangrydev, line 28: Rename to match the Baeldung test naming convention + I think this should be labelled as an integration test rather than a unit test

spring-kafka/src/main/java/com/baeldung/partitioningstrategy/KafkaMessageConsumer.java
```diff
@@ -0,0 +1,30 @@
...
+        receivedMessages.add(receivedMessage);
+    }
+
+    public List<ReceivedMessage> getReceivedMessages() {
+        return receivedMessages;
+    }
+
+    public void clearReceivedMessages() { receivedMessages.clear(); }
```
theangrydev, line 29: I don't think this is necessary, your tests should be instantiating a fresh copy of this on every test run

theangrydev (changes requested):
Makes sense overall, just a few details to take care of

spring-kafka/src/main/java/com/baeldung/partitioningstrategy/KafkaApplication.java
wynnteo: Noted.

spring-kafka/src/main/java/com/baeldung/partitioningstrategy/KafkaMessageConsumer.java
wynnteo: I've checked, the @KafkaListener is not inherently thread-safe. Thanks for pointing out this, will change to CopyOnWriteArrayList instead.

spring-kafka/src/main/java/com/baeldung/partitioningstrategy/KafkaMessageConsumer.java
wynnteo: The line was removed.

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
wynnteo: I realised the messages accumulated there is why I do this to reinitiliaze the message list.

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
wynnteo: Noted with this.

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
wynnteo: Updated

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
wynnteo: Updated

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
wynnteo: I've checked, it actually handled by the @SpringBootTest, therefore I've removed it.

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
wynnteo: Updated

spring-kafka/src/main/java/com/baeldung/partitioningstrategy/KafkaMessageConsumer.java
wynnteo: I think because the receivedMessages list is a member of the KafkaMessageConsumer class, instantiated as a Spring bean. It retains its state throughout the test suite.

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationUnitTesting.java
theangrydev: Sorry, I did not realise that the Spring context was static and cached between test runs

spring-kafka/src/main/java/com/baeldung/partitioningstrategy/KafkaMessageConsumer.java
theangrydev: Sorry, I did not realise that the Spring context was static and cached between test runs

spring-kafka/src/main/java/com/baeldung/partitioningstrategy/KafkaProducer.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.springframework.kafka.core.KafkaTemplate;
+import org.springframework.stereotype.Service;
+
+@Service
+public class KafkaProducer {
+
+    private final KafkaTemplate<String, String> kafkaTemplate;
+
+    public KafkaProducer(KafkaTemplate<String, String> kafkaTemplate) {
```
theangrydev, line 11: For clarity, do you think we should annotate this constructor with `@AutoWired` even though it is not strictly necessary?

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationIntegrationTesting.java
```diff
@@ -0,0 +1,193 @@
...
+            .until(() -> kafkaMessageConsumer.getReceivedMessages()
+                .size() >= 3);
+
+        List<ReceivedMessage> records = kafkaMessageConsumer.getReceivedMessages();
+
+        Set<Integer> uniquePartitions = records.stream()
+            .map(ReceivedMessage::getPartition)
+            .collect(Collectors.toSet());
+
+        Assert.assertEquals(1, uniquePartitions.size());
```
theangrydev, line 69:
```suggestion
        assertEquals(1, uniquePartitions.size());
```

theangrydev (changes requested):

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationIntegrationTesting.java
```diff
@@ -0,0 +1,193 @@
...
+import org.springframework.kafka.test.EmbeddedKafkaBroker;
+import org.springframework.kafka.test.context.EmbeddedKafka;
+import org.springframework.kafka.test.utils.KafkaTestUtils;
+
+import static org.awaitility.Awaitility.await;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+@SpringBootTest
+@EmbeddedKafka(partitions = 3, brokerProperties = { "listeners=PLAINTEXT://localhost:9092" })
+public class KafkaApplicationIntegrationTesting {
```
theangrydev, line 35: Sorry, this still does not match the test naming convention, please see:
https://github.com/eugenp/tutorials/blob/5a3511a0a937b9816988191cf628f05423a31091/custom-pmd/src/main/java/com/baeldung/pmd/UnitTestNamingConventionRule.java#L17

theangrydev:
```suggestion
public class KafkaApplicationIntegrationTest {
```

spring-kafka/src/main/java/com/baeldung/partitioningstrategy/KafkaProducer.java
```diff
@@ -0,0 +1,26 @@
+package com.baeldung.partitioningstrategy;
+
+import org.springframework.kafka.core.KafkaTemplate;
+import org.springframework.stereotype.Service;
+
+@Service
+public class KafkaProducer {
```
theangrydev, line 7: What is the purpose of this wrapper? Can we remove it and use the KafkaTemplate directly? I'm worried that in the article we refer to kafkaProducer.send rather than the standard  kafkaTemplate.send and so it might not be clear that we are using standard Kafka library features

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationIntegrationTesting.java
wynnteo: Updated.

spring-kafka/src/main/java/com/baeldung/partitioningstrategy/KafkaProducer.java
wynnteo: I've removed the wrapper.

spring-kafka/src/test/java/com/baeldung/partitioningstrategy/KafkaApplicationIntegrationTesting.java
wynnteo: Updated.

spring-kafka/src/main/java/com/baeldung/partitioningstrategy/KafkaProducer.java
wynnteo: At first, I was using it to do some object transforming. But I find it defeats the purpose and will make the example complicated.  I agree with you that should use the standard kafkaTemplate instead. Updated.

theangrydev (approved):

PR #15513, opened 2023-12-29
BAEL-7368 - String to Long 

danielmcnally285:
String to long


Maiklins:
retest this please

danielmcnally285:
@Maiklins It looks to me that master for core-java-modules submodule doesn't compile as it fails to find the 'parent-modules' module. It looks like some work has been done recently to move about the modules as I noticed that the module 'parent-java' doesn't exist anymore. The latest commit in the PR seems to rectify the issue whereby the Jenkins build ci now passes (this was based on other PRs which seem to be having to do the same commit to get around this). However, I am unsure as to whether this is the 'correct' fix. Please let me know what you think

Maiklins:
retest this please

dkapil:
test this please

PR #15529, opened 2023-12-31
baeldung-articles : BAEL-7041

Diegom203:
PR for the article "Check if List Contains At Least One Enum" which adds the "CheckIfListContainsEnumUnitTest" class to check if a list contains an enum in Java.


dkapil:
test this please

PR #15531, opened 2024-01-01
[rm-char-in-other-str] BAEL-7355 Remove characters from a string that…

sk1418:
… are present in a second string


dkapil:
test this please

PR #15577, opened 2024-01-09
JAVA-27512: Changes made for upgrading spring-websockets to Spring Bo…

Bipinkumar27:
…ot 3

### DIFF
--- a/spring-websockets/pom.xml
+++ b/spring-websockets/pom.xml
@@ -37,23 +37,6 @@
 
     <build>
         <finalName>spring-websockets</finalName>
-        <plugins>
-            <plugin>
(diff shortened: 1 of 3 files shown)


spring-websockets/pom.xml
```diff
@@ -37,6 +37,23 @@
 
     <build>
         <finalName>spring-websockets</finalName>
```
anastasiosioannidis, line 39: Could you remove the plugin spring-boot-maven-plugin and add:      
      
        <spring-boot.repackage.skip>true</spring-boot.repackage.skip>
        
down in <properties> tag?

Bipinkumar27: sure

PR #15618, opened 2024-01-13
JAVA-27759:- Fix Spring 3.1.x and Keycloak 22.x OAuth2 Tutorial 

amit2103:


spring-boot-modules/spring-boot-keycloak/src/main/java/com/baeldung/keycloak/SecurityConfig.java
```diff
@@ -19,6 +25,9 @@
 @EnableWebSecurity
 class SecurityConfig {
 
+    private static final String REALM_ACCESS_CLAIM = "groups";
```
Luiggi33, line 28: for me, this is "realm_access", may I know how you got that value?

amit2103: Hey, for keycloack distribution in the article this is where the claims are. You can inspect the token in debug and check.

Luiggi33: Thanks for the clarification!

PR #15631, opened 2024-01-14
JAVA-29307 Upgrade spring-security-web-boot-4

timis1:


spring-security-modules/spring-security-web-boot-4/src/main/java/com/baeldung/securityfilterchain/configuration/SecurityConfig.java
```diff
@@ -2,49 +2,42 @@
...
 
         return http.build();
     }
 
     @Bean
     public WebSecurityCustomizer webSecurityCustomizer() {
-        return (web) -> web.debug(securityDebug)
-          .ignoring()
-          .antMatchers("/css/**", "/js/**", "/img/**", "/lib/**", "/favicon.ico");
+        return web -> web.debug(securityDebug).ignoring().requestMatchers("/css/**", "/js/**", "/img/**", "/lib/**", "/favicon.ico");
```
alvin577austria, line 41: Can you check the formatting here pls, this line seems a bit too long.

spring-security-modules/spring-security-web-boot-4/src/main/java/com/baeldung/apikeyauthentication/configuration/ApiKeyAuthentication.java
```diff
@@ -22,4 +22,14 @@ public Object getCredentials() {
     public Object getPrincipal() {
         return apiKey;
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        return super.equals(obj);
```
alvin577austria, line 28: Is there a specific reason why we need to override these? And we are just returning what the superclass have so this seems unnecessary.

alvin577austria (changes requested):

spring-security-modules/spring-security-web-boot-4/src/main/java/com/baeldung/securityfilterchain/configuration/SecurityConfig.java
timis1: I have a sonar lint notification that warns me to override also the equals and hashCode since the methods are overridden in the super class. We can remove it.

timis1: Done

spring-security-modules/spring-security-web-boot-4/src/main/java/com/baeldung/apikeyauthentication/configuration/ApiKeyAuthentication.java
timis1: I have a sonar lint notification that warns me to override also the equals and hashCode since the methods are overridden in the super class. We can remove it.

alvin577austria (approved):

PR #15652, opened 2024-01-15
Correct artifactId in dependencies for maven-userdaomodule

knightspot:
In JAVA-16264: review module names was the artifactId in dependencies not changes. This should fix the problem


knightspot:
With this correction is it possible to build and run **multimodulemavenproject**

alvin577austria:
PR is valid. This will be taken cared of in https://team.baeldung.com/browse/JAVA-27739. Closing.

PR #15675, opened 2024-01-18
JAVA-27677: Changes made for adding sub-modules aren't being included…

Bipinkumar27:
… | core-java-os-2 and spring-6-rsocket


spring-6-rsocket/pom.xml
```diff
@@ -33,6 +33,26 @@
             <type>pom</type>
             <scope>import</scope>
         </dependency>
```
anastasiosioannidis, line 35: @Bipinkumar27 

Try to delete these 4 dependencies and check that build/integration works.
Of course leave the properties tag changes.

Bipinkumar27: It worked !! I have done the changes

PR #15694, opened 2024-01-20
JAVA-27509 :- Update Spring Boot validation to Spring Boot 3. 

amit2103:
Also added a application.properties as without the application was not starting due to the table name user being a reserved keyword.

### DIFF
--- a/spring-boot-modules/spring-boot-validation/src/main/resources/application.properties
+++ b/spring-boot-modules/spring-boot-validation/src/main/resources/application.properties
@@ -1,0 +1,14 @@
+spring.datasource.url=jdbc:h2:mem:mydb
+spring.datasource.driverClassName=org.h2.Driver
+spring.datasource.username=sa
+spring.datasource.password=
+spring.jpa.defer-datasource-initialization=true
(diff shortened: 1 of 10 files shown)


spring-boot-modules/spring-boot-validation/src/main/resources/application.properties
anastasiosioannidis, line 1: Hi @amit2103 
I removed this and build worked fine.

amit2103: @anastasiosioannidis  This is for the Caused by: org.h2.jdbc.JdbcSQLSyntaxErrorException: Syntax error in SQL statement "insert into [*]user (email, name, id) values (?, ?, ?)"; expected "identifier"; SQL statement:
insert into user (email, name, id) values (?, ?, ?) [42001-214].

This doesnt affect the build and only will happen when you run the build.

amit2103: User is a reserved keyword in H2 hence its failing without that

PR #15706, opened 2024-01-22
[Spring-Boot-Keycloak example] Allows preflight requests from browser

libor-vilimek-finbricks:
If anyone uses frontend to call the request with access token, the browsers will first try to send the OPTIONS preflight request. This update will allow them to get the 200 response and continue with actual request.


libor-vilimek-finbricks:
@amit2103 - might want to review this

PR #15708, opened 2024-01-22
BAEL-7199: Override bootstrap servers property

constantinurs:


spring-kafka-2/src/main/resources/application.properties
```diff
@@ -1,4 +1,4 @@
-spring.kafka.bootstrap-servers=localhost:9092,localhost:9093,localhost:9094,localhost:9095
```
constantinurs, line 1: I added localhost:9095 in a previous attempt to fix the tests - it's not required anymore

PR #15747, opened 2024-01-26
JAVA-29287 Upgrade spring-security-azuread

timis1:


spring-security-modules/spring-security-azuread/src/main/java/com/baeldung/security/azuread/config/JwtAuthorizationConfiguration.java
```diff
@@ -45,20 +39,18 @@ private OAuth2UserService<OidcUserRequest, OidcUser> customOidcUserService(JwtAu
...
+
             oidcUser = new NamedOidcUser(mappedAuthorities, oidcUser.getIdToken(), oidcUser.getUserInfo(),oidcUser.getName());
 
             return oidcUser;
         };
     }
-
-    
     
 //    @Bean
```
alvin577austria, line 55: Do we know why we have commented code here?

alvin577austria (commented):
Some comment.

spring-security-modules/spring-security-azuread/src/main/java/com/baeldung/security/azuread/config/JwtAuthorizationConfiguration.java
alvin577austria: Like, is it mentioned in the article that this bean can also be used or something.

alvin577austria: Yes I noticed that, maybe we can remove the commented code as part of the upgrade, If it is not needed of course.
It can be that it is there for spting boot 2 but no longer relevant for spring boot 3.
Can you verify please if we need to retain it.

timis1: Sure, I will check

timis1: As a conclusion: the MappingJwtGrantedAuthoritiesMapper doesn't exist, so in case some one will uncomment that code is unusable.
Do you suggest to remove that pice of commented code?

alvin577austria: Yes please, it seems it is some previous code that was forgotten to be removed. Thanks!

alvin577austria (changes requested):

spring-security-modules/spring-security-azuread/src/main/java/com/baeldung/security/azuread/config/JwtAuthorizationConfiguration.java
timis1: Done

PR #15759, opened 2024-01-28
BAEL-7351 Demo code

s9m33r:


s9m33r:
The tests I've added are passing now on my local. However, I still see other tests failing on my local when I run all tests in spring-kafka-2 module. Can you please advice?

s9m33r:
@asif-anwar we're good here right?

spring-kafka-2/src/test/java/com/baeldung/spring/kafka/groupId/MainLiveTest.java
```diff
@@ -0,0 +1,44 @@
...
+        consumer.resetLatch();
+    }
+
+    @Test
+    public void givenEmbeddedKafkaBroker_whenSendingWithSimpleProducer_thenMessageReceived() throws Exception {
+        String data = "Test 123...";
+
+        producer.send(data);
+
+        boolean messageConsumed = consumer.getLatch()
```
acerasoni, line 39: We don't need a newline here

s9m33r: Done.

pom.xml
```diff
@@ -689,8 +689,8 @@
                 <module>apache-httpclient-2</module>
                 <module>apache-httpclient4</module>
                 <module>apache-httpclient</module>
-                <module>apache-kafka-2</module>
                 <module>apache-kafka</module>
+                <module>apache-kafka-2</module>
```
KevinGilmore, line 693: Is this change necessary for the topic at hand? Generally, we should avoid changes of this nature and leave them for the editorial team to clean up at a later date.

spring-kafka-2/src/main/java/com/baeldung/spring/kafka/groupId/MyKafkaConsumer.java
```diff
@@ -0,0 +1,41 @@
+package com.baeldung.spring.kafka.groupId;
```
KevinGilmore, line 1: Note that this module is already associated with 7 articles, which is our max for Spring modules. Let's move the code to the spring-kafka-3 module.

pom.xml
s9m33r: It's not reverted.

spring-kafka-2/src/main/java/com/baeldung/spring/kafka/groupId/MyKafkaConsumer.java
s9m33r: Done. The code has been moved to spring-kafka-3.

PR #15780, opened 2024-01-31
Code for deep and shallow copy

mee-omi:
Added code for evaluation article for deep and shallow copy in Java. Kindly check.


mee-omi:
Replaced tabs with four spaces in formatting
removed Main class
reduced number of properties in example class

PR #15815, opened 2024-02-04
[mv-0-to-end-array] move zeros to the end

sk1418:


core-java-modules/core-java-arrays-operations-advanced-2/src/test/java/com/baeldung/movezerototheend/MoveZeroesToTheEndOfAnArrayUnitTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Test
+    void whenMovingZeroInTheOriginalArray_thenGetTheExpectedResult() {
+        int[] array = new int[] { 1, 2, 0, 3, 4, 0 };
+        int idx = 0;
+        for (int n : array) {
+            if (n != 0) {
+                array[idx++] = n;
+            }
+            System.out.println(Arrays.toString(array));
```
theangrydev, line 33: Can we remove this? If not, we have a guideline to use SLF4J for logging and not System.out

core-java-modules/core-java-arrays-operations-advanced-2/src/test/java/com/baeldung/movezerototheend/MoveZeroesToTheEndOfAnArrayUnitTest.java
```diff
@@ -0,0 +1,40 @@
+package com.baeldung.movezerototheend;
+
+import org.junit.jupiter.api.Test;
+
+import java.util.Arrays;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+
+public class MoveZeroesToTheEndOfAnArrayUnitTest {
+    static final int[] EXPECTED = new int[] { 1, 2, 3, 4, 0, 0 };
```
theangrydev, line 10: Can we pick a slightly different example that has some unordered non-zero elements instead of the ordered sequence 1234? Just to emphasise that this algorithm doesn't do any kind of sorting, and only changes the position of the zeros

theangrydev, line 10:
```suggestion
    private static final int[] EXPECTED = new int[] { 1, 2, 3, 4, 0, 0 };
```

theangrydev (changes requested):

core-java-modules/core-java-arrays-operations-advanced-2/src/test/java/com/baeldung/movezerototheend/MoveZeroesToTheEndOfAnArrayUnitTest.java
sk1418: I put it here to print step result, and used the output in the article. Of course, this shouldn't be in the code. I forgot to remove it. Thank you for spotting it. 👍 
fixed in the new commit.

core-java-modules/core-java-arrays-operations-advanced-2/src/test/java/com/baeldung/movezerototheend/MoveZeroesToTheEndOfAnArrayUnitTest.java
sk1418: fixed

sk1418: I changed it to `42, 2, 3, 4, 0, 0`. So, sorting is excluded.

theangrydev (approved):

PR #15823, opened 2024-02-06
JAVA-29170: Changes made for Upgrade spring-rest-http to Spring Boot 3 

Bipinkumar27:


spring-web-modules/spring-rest-http/src/main/java/com/baeldung/responseheaders/controllers/FilterResponseHeaderController.java
alvin577austria, line 1: While not part of your changes here, can you please rename the two methods here to start with a small "f" to make it camel case compliant please. Thanks!

alvin577austria (changes requested):
Small comment

spring-web-modules/spring-rest-http/src/main/java/com/baeldung/responseheaders/controllers/FilterResponseHeaderController.java
Bipinkumar27: Hi @alvin577austria I have done the changes

PR #15824, opened 2024-02-06
JAVA-31108 Upgrade parent-boot-3 to latest version

anuragkumawat:


persistence-modules/spring-mybatis/pom.xml
```diff
@@ -84,6 +84,7 @@
         <mybatis.version>3.5.2</mybatis.version>
         <mybatis-spring-boot-starter.version>3.0.3</mybatis-spring-boot-starter.version>
         <spring-boot.repackage.skip>true</spring-boot.repackage.skip>
```
anastasiosioannidis, line 86: @anuragkumawat 

Why are we adding 3.1.5, since it's pre-existing already?
If I remove this line, and run mvn dependency:tree, I get Spring boot 3.1.5

Same for the other instances below/up

anuragkumawat: Since the version is upgraded to 3.2.2 in parent-boot-3, the module picks 3.2.2 version and fails. In order to run the build I am forcing 3.1.5, Ideally the issue should be fixed and 3.2.2 version should be used.
Thanks

PR #15875, opened 2024-02-13
JAVA-29166: Changes made for Upgrade spring-caching to Spring Boot 3

Bipinkumar27:


spring-boot-modules/spring-caching/pom.xml
```diff
@@ -77,8 +79,19 @@
         </dependency>
     </dependencies>
```
anastasiosioannidis, line 81: Try to remove this, does build/integration work?

PR #15904, opened 2024-02-17
[JAVA-29500]

panos-kakos:


anastasiosioannidis:
retest this please

spring-kafka/src/test/java/com/baeldung/kafka/retryable/KafkaRetryableIntegrationTest.java
```diff
@@ -41,9 +35,9 @@ public class KafkaRetryableIntegrationTest {
 
     private static final String TOPIC = "topic";
```
anastasiosioannidis, line 37: Since this task is only about moving, why this change?

spring-kafka/src/main/java/com/baeldung/kafka/retryable/Farewell.java
```diff
@@ -1,4 +1,4 @@
-package com.baeldung.spring.kafka.retryable;
```
anastasiosioannidis, line 1: Why did we need to change the package here and in below cases?

PR #15932, opened 2024-02-20
JAVA-29330: Migrate spring-security-web-rest-custom to  parent-boot-3.

harry9656:


spring-security-modules/spring-security-web-rest-custom/src/main/java/com/baeldung/config/MainWebAppInitializer.java
```diff
@@ -2,33 +2,26 @@
...
-     */
     @Override
-    public void onStartup(final ServletContext sc) throws ServletException {
+    public void onStartup(final ServletContext sc) {
         System.out.println("MyWebAppInitializer.onStartup()");
 
-        // Create the 'root' Spring application context
         final AnnotationConfigWebApplicationContext root = new AnnotationConfigWebApplicationContext();
         root.scan("com.baeldung.config.parent");
         // root.getEnvironment().setDefaultProfiles("embedded");
```
alvin577austria, line 27: Do we know why we have this commented code here? Maybe we can remove this.

spring-security-modules/spring-security-web-rest-custom/src/main/java/com/baeldung/config/child/WebConfig.java
```diff
@@ -27,23 +26,19 @@
 //@ImportResource({ "classpath:prop.xml" })
```
alvin577austria, line 26: Same as comment above

alvin577austria (changes requested):

spring-security-modules/spring-security-web-rest-custom/src/main/java/com/baeldung/config/MainWebAppInitializer.java
alvin577austria: Like maybe it is mentioned in the article or so.

harry9656: In fact, there is no mention. Therefore, removing.

spring-security-modules/spring-security-web-rest-custom/src/main/java/com/baeldung/config/child/WebConfig.java
harry9656: It must be a consequence of a copy-paste from the ServiceConfig.class. I am removing them as well.

PR #15937, opened 2024-02-21
JAVA-31570:  Changes made for fixing test failures with a named param…

Bipinkumar27:
…eter

### DIFF
--- a/parent-boot-3/pom.xml
+++ b/parent-boot-3/pom.xml
@@ -73,6 +73,9 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-compiler-plugin</artifactId>
                     <version>${maven-compiler-plugin.version}</version>
+                    <configuration>
+                        <parameters>true</parameters>
+                    </configuration>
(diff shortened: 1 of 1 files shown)


harry9656 (commented):
@alvin577austria This change will have second order effects. Please consider them before merging. A would suggest for a more "local" fix then changing the parent-boot-3.

alvin577austria:
I agree, since parent-boot-3 is being used by other modules it would be best as of this moment to fix the failing test specific to the module.

alvin577austria:
> I agree, since parent-boot-3 is being used by other modules it would be best as of this moment to fix the failing test specific to the module.

alvin577austria:
I have closed this PR.
I checked spring-web-modules\spring-rest-http\pom.xml and it seems it was modified two days ago with this
https://github.com/eugenp/tutorials/pull/15897/files#diff-44fdff59d9f4b124fcc7f3a63c421b00a8e2c7ee20235214f302e53e49108403

This change have fixed the failing test in this module.

PR #15962, opened 2024-02-25
[modify-print-list] modify + print a list

sk1418:


core-java-modules/core-java-collections-list-6/src/test/java/com/baeldung/modifyandprint/ModifyAndPrintListElementsUnitTest.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.common.collect.Lists;
+
+public class ModifyAndPrintListElementsUnitTest {
+
+    private final Logger log = LoggerFactory.getLogger(ModifyAndPrintListElementsUnitTest.class);
+
+    @Test
+    void whenPrintingInForEach_thenListIsPrinted() {
+        List<String> theList = Lists.newArrayList("Kai", "Liam", "Eric", "Kevin");
+        theList.forEach(e -> log.info(e));
```
theangrydev, line 21: In these kinds of cases I usually prefer the log::info method reference syntax

I also prefer words for lambda variable names instead of single letters

Both of these are personal preferences, happy to keep things as-is if you prefer the current approach more

core-java-modules/core-java-collections-list-6/src/test/java/com/baeldung/modifyandprint/ModifyAndPrintListElementsUnitTest.java
```diff
@@ -0,0 +1,56 @@
...
+    @Test
+    void whenPrintingInMap_thenStreamIsModifiedAndPrinted() {
+        List<String> theList = List.of("Kai", "Liam", "Eric", "Kevin");
+        List<String> newList = theList.stream()
+            .map(e -> {
+                String newElement = e.toUpperCase();
+                log.info(newElement);
+                return newElement;
+            })
+            .filter(e -> e.length() == 4)
```
theangrydev, line 41: To keep the focus of the article, I feel like we should drop the filters and just stick to the case where we are modifying all the elements

theangrydev (changes requested):

core-java-modules/core-java-collections-list-6/src/test/java/com/baeldung/modifyandprint/ModifyAndPrintListElementsUnitTest.java
sk1418: In general, both are ok to me. With some FP and Python background, I personally find lambda (just slightly) easier to read. 
But I picked lambda expressions here as we can only use lambda in the map() with print() approach. I thought it might be good not to mix them in the article. Probably, it's an unimportant consideration. 😄 

I renamed the lambda variables from e -> element.

core-java-modules/core-java-collections-list-6/src/test/java/com/baeldung/modifyandprint/ModifyAndPrintListElementsUnitTest.java
sk1418: filter() removed

theangrydev (approved):

PR #15999, opened 2024-02-28
JAVA-31563-update-spring-kafka-to-latest-version

sam-gardner:


spring-boot-modules/spring-boot-libraries-3/pom.xml
```diff
@@ -84,6 +85,7 @@
         <spring-modulith-events-kafka.version>1.1.2</spring-modulith-events-kafka.version>
         <testcontainers.version>1.19.3</testcontainers.version>
         <awaitility.version>4.2.0</awaitility.version>
+        <spring-kafka.version>3.0.14</spring-kafka.version>
```
sam-gardner, line 88: Latest version compatible with the Spring Boot version declared on line 84

alvin577austria: Can you try to upgrade kafka to 3.1.2 as well here please. The property spring-boot.version is not even used in this pom so we can remove it. It should then get the spring-boot version from parent-boot-3 which is the same for other modules updated in this PR.

alvin577austria (changes requested):

spring-boot-modules/spring-boot-libraries-3/pom.xml
sam-gardner: Ive made the update and removed the unused spring property dependency. However I checked on master branch and the application doesn't run and one live test fails locally. The behaviour is exactly the same with this change.

I'd guess because there are no unit or integration tests this broke after upgrading to Spring Boot 3 or similar and it went unnoticed.

alvin577austria: Can you put here the failing test please, in my local using this branch and master branch all tests pass.

alvin577austria: I think I got waht you meant, I will create a JIRA for the failing application startup.

sam-gardner: For me whenArticleIsSavedToDB_thenItIsAlsoPublishedToKafka fails with this error:
```
java.util.concurrent.CompletionException: org.apache.kafka.common.KafkaException: Failed to construct kafka producer
	at org.springframework.util.concurrent.FutureUtils.lambda$toSupplier$0(FutureUtils.java:78) ~[spring-core-6.0.13.jar:6.0.13]
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[na:na]
	at java.base/java.lang.Thread.run(Thread.java:840) ~[na:na]
Caused by: org.apache.kafka.common.KafkaException: Failed to construct kafka producer
```
I'm using Java 17 and just running the tests in InteliJ. Is there anything else I should be doing?

PR #16006, opened 2024-02-29
[BAEL-7127] Implement two levels of caching with local and distributed cache

saikatcse03:
Implement two levels of caching using redis as distributed cache and caffeine as local cache for the application.


spring-boot-modules/spring-caching-2/src/main/java/com/baeldung/caching/twolevelcache/CacheConfig.java
```diff
@@ -0,0 +1,71 @@
...
+
+@Configuration
+@EnableCaching
+public class CacheConfig {
+
+    @Bean
+    @Primary
+    public CacheManager caffeineCacheManager() {
+        SimpleCacheManager manager = new SimpleCacheManager();
+        manager.setCaches(Arrays.asList(caffeineCacheConfig()));
```
acerasoni, line 31: Let's autowire the CaffineCache bean via method parameter rather than calling the method directly

saikatcse03: Added objects via method param in all such relevant config including this one.

PR #16028, opened 2024-03-02
This PR is related to BAEL-7427

BenHelmyBen:
This PR aims to add "ReadingGZIPUsingGZIPInputStreamUnitTest" and "Main" classes to the article BAEL-7427. Furthermore, the myFile.gz file to the resources folder.


core-java-modules/core-java-io-5/src/main/java/com/baeldung/usinggzipInputstream/Main.java
```diff
@@ -0,0 +1,57 @@
...
+    public static void processGZipFile(InputStream inputStream, LineProcessor lineProcessor) throws IOException {
+        try (Stream<String> lines = new BufferedReader(new InputStreamReader(new GZIPInputStream(inputStream)))
+                .lines()
+                .map(String::trim)) {
+            lineProcessor.process(lines);
+        }
+    }
+
+    @FunctionalInterface
+    public interface LineProcessor {
```
ashleyfrieze, line 54: This is just a `Consumer<Stream<String>>` by another name

core-java-modules/core-java-io-5/src/main/java/com/baeldung/usinggzipInputstream/Main.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class Main {
+    static String filePath = Objects.requireNonNull(Main.class.getClassLoader().getResource("myFile.gz")).getFile();
+
+    public static void main(String[] args) throws IOException {
+
+        try {
+            readGZipFileWithFilter(filePath, lines -> lines.forEach(System.out::println));
+        } catch (FileNotFoundException e) {
+            System.err.println(e.getMessage());
```
ashleyfrieze, line 18: Let's avoid this pattern. Just outputting an error to `System.err` isn't helpful. Let's allow main to `throws` any exception it encounters here.

core-java-modules/core-java-io-5/src/main/java/com/baeldung/usinggzipInputstream/Main.java
```diff
@@ -0,0 +1,41 @@
...
+             BufferedReader bufferedReader = new BufferedReader(inputStreamReader)) {
+
+            String line;
+            while ((line = bufferedReader.readLine()) != null) {
+                lines.add(line);
+            }
+        }
+
+        return lines;
+    }
```
ashleyfrieze, line 36: This algorithm is good. I also expect to see two more variations:

```java
    public static List<String> findInZipFile(String filePath, String toFind) throws IOException {
        try (InputStream inputStream = new FileInputStream(filePath);
             GZIPInputStream gzipInputStream = new GZIPInputStream(inputStream);
             InputStreamReader inputStreamReader = new InputStreamReader(gzipInputStream);
             BufferedReader bufferedReader = new BufferedReader(inputStreamReader)) {

             return bufferedReader.lines().filter(line -> line.contains(toFind)).collect(toList());
        }
    }

    public static List<String> useContentsOfZipFile(String filePath, Consumer<Stream<String>> consumer) throws IOException {
        try (InputStream inputStream = new FileInputStream(filePath);
             GZIPInputStream gzipInputStream = new GZIPInputStream(inputStream);
             InputStreamReader inputStreamReader = new InputStreamReader(gzipInputStream);
             BufferedReader bufferedReader = new BufferedReader(inputStreamReader)) {

            consumer.accept(bufferedReader.lines());
        }
    }


```

We can also unit test these. We explain how it's important to provide the `try-with-resources` in each case, hence needing the `lines` to be used within that construct.

ashleyfrieze (changes requested):

core-java-modules/core-java-io-5/src/main/java/com/baeldung/usinggzipInputstream/Main.java
```diff
@@ -0,0 +1,77 @@
...
+    public static List<String> useContentsOfZipFile(String filePath, Consumer<Stream<String>> consumer) throws IOException {
+        List<String> linesList;
+        try (InputStream inputStream = new FileInputStream(filePath);
+             GZIPInputStream gzipInputStream = new GZIPInputStream(inputStream);
+             InputStreamReader inputStreamReader = new InputStreamReader(gzipInputStream);
+             BufferedReader bufferedReader = new BufferedReader(inputStreamReader)) {
+
+            Stream<String> linesStream = bufferedReader.lines();
+            linesList = linesStream.collect(Collectors.toList());
+            consumer.accept(linesList.stream());
```
ashleyfrieze, line 72: This method misses the point of using a `Stream` so as not to keep the whole file in memory. Let's remove the `List` from it altogether and just go:

```java
consumer.accept(bufferedReader.lines());
```

PR #16039, opened 2024-03-04
BAEL-7598 Check if a number is a perfect number

alexyang0623:


algorithms-modules/algorithms-miscellaneous-7/src/test/java/com/baeldung/algorithms/perfectnumber/PerfectNumberUnitTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class PerfectNumberUnitTest {
+
+    @Test
+    void testIsPerfectBruteForce() {
```
JavierPeris, line 11: Test must follow the naming convention givenXX_whenXXX_thenXXX

algorithms-modules/algorithms-miscellaneous-7/src/main/java/com/baeldung/algorithms/perfectnumber/PerfectNumber.java
```diff
@@ -0,0 +1,27 @@
+package com.baeldung.algorithms.perfectnumber;
+
+public class PerfectNumber {
```
JavierPeris, line 3: We can remove this public modifier, JUnit 5 does not need it.

algorithms-modules/algorithms-miscellaneous-7/pom.xml
```diff
@@ -13,4 +13,40 @@
...
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <version>1.35</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>1.35</version>
```
JavierPeris, line 26: Please use a property for the version.

algorithms-modules/algorithms-miscellaneous-7/pom.xml
```diff
@@ -13,4 +13,40 @@
...
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>org.openjdk.jmh</groupId>
+                            <artifactId>jmh-generator-annprocess</artifactId>
+                            <version>1.35</version>
+                        </path>
+                    </annotationProcessorPaths>
+                </configuration>
+            </plugin>
+        </plugins>
```
JavierPeris, line 49: Do we really need to override the plugin configuration? I'm asking because I see the use jmh-generator-annprocess dependency in other maven modules but I don't see the plugin configuration.

travel2china: Removed. Some modules include this configuration. However, it is not required, as Maven will use service discovery to find the proper annotation processor.

alexyang0623: Removed. Some modules include this configuration. However, it is not required, as Maven will use service discovery to find the proper annotation processor.

JavierPeris: Thanks for the clarification!

PR #16101, opened 2024-03-12
Updated the JNats Publish and Receive article to latest version.

scottf:
Updated the JNats Publish and Receive article to latest version of JNats including the example NatsClient and the unit tests.


libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
```diff
@@ -1,123 +1,144 @@
...
+        if (serverURI == null || serverURI.isEmpty()) {
+            this.serverURI = "nats://localhost:4222";
+        }
+        else {
             this.serverURI = serverURI;
-        } else {
-            this.serverURI = "jnats://localhost:4222";
         }
+        initConnection();
+    }
```
theangrydev, line 36: I'm not a fan of having behaviour/logic inside constructors, can we refactor this to use a static factory method and keep the constructor responsibility just for assigning the fields

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
```diff
@@ -1,123 +1,144 @@
...
     private final String serverURI;
 
-    private final Connection natsConnection;
-
-    private final Map<String, Subscription> subscriptions = new HashMap<>();
+    private final Map<Subscription, Dispatcher> dispatcherBySubscription = new HashMap<>();
 
     private final static Logger log = LoggerFactory.getLogger(NatsClient.class);
 
+    private Connection natsConnection;
```
theangrydev, line 22:
```suggestion
    private final Connection natsConnection;
```

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
```diff
@@ -1,123 +1,144 @@
...
 
     public void closeConnection() {
         // Close connection
-        natsConnection.close();
+        try {
+            natsConnection.close();
+        }
+        catch (InterruptedException e) {
+            log.warn("Warning:", e);
+        }
```
theangrydev, line 69: When we catch InterruptedException it should always be followed by Thread.currentThread().interrupt() inside the catch block to restore the interrupt status

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
```diff
@@ -1,123 +1,144 @@
...
+        try {
+            Options options = new Options.Builder()
+                .server(serverURI)
+                .connectionListener((connection, event) -> log.info("Connection Event: {}", event.toString()))
+                .errorListener(new CustomErrorListener())
+                .build();
+            natsConnection = Nats.connect(options);
+        } catch (IOException | InterruptedException e) {
+            log.error("Error connecting to NATs! {}", e.toString());
+        }
```
theangrydev, line 59: When we catch InterruptedException it should always be followed by Thread.currentThread().interrupt() inside the catch block to restore the interrupt status

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
```diff
@@ -1,123 +1,144 @@
...
         }
     }
 
-    public void subscribeAsync(String topic) {
-
-        AsyncSubscription subscription = natsConnection.subscribe(
-          topic, msg -> log.info("Received message on {}", msg.getSubject()));
-
+    public Subscription subscribeAsync(String subject) {
+        Dispatcher dispatcher = natsConnection.createDispatcher();
```
theangrydev, line 89: What is the lifecycle of a Disptacher? When should it be closed?

theangrydev (changes requested):

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
scottf: Can you give me a better explanation? Maybe the question is would you have made the same comment on the original code, because it's pretty much the same.

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
scottf: will fix

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
scottf: It will be close when the connection is closed. Dispatcher is always created and maintained from the connection and the connection is AutoCloseable.

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,123 @@
 package com.baeldung.jnats;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
+import io.nats.client.Message;
+import io.nats.client.Subscription;
+import org.junit.Test;
```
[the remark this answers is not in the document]
scottf: This probably is somewhere above the example. This is inherited code.

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,123 @@
...
 
 public class NatsClientLiveTest {
 
+    private NatsClient connectClient() {
+        return new NatsClient();
+    }
+
     @Test
-    public void givenMessageExchange_MessagesReceived() throws Exception {
+    public void messageExchangeViaPublishSubscribe_MessagesReceived() throws Exception {
```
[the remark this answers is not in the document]
scottf: will fix

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
scottf: Fixed.

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
scottf: I made the client AutoCloseable

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
theangrydev: Sorry I missed your question here. I mean, the single responsibility of a constructor is usually simply to assign fields.

Putting if statements, or other init logic inside a constructor is generally considered bad practice and comes as a result of a lack of proper dependency injection. I'll leave a suggestion showing what I would usually expect to see here.

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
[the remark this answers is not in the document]
theangrydev: Sorry, we will still need to fix this to match current guidelines. Can you migrate this to JUnit5 please?

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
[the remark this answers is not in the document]
theangrydev: This hasn't been addressed, the syntax is either:
1. givenSomethingBeforeTheWhen_whenTheActionIsTaken_thenTheExpectedResult
2. whenTheActionIsTaken_thenTheExpectedResult

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
```diff
@@ -1,123 +1,142 @@
...
+        else {
             this.serverURI = serverURI;
-        } else {
-            this.serverURI = "jnats://localhost:4222";
         }
+        natsConnection = createConnection(this.serverURI);
+    }
 
-        natsConnection = initConnection(serverURI);
+    public static Connection createConnection(String serverURI) throws IOException, InterruptedException {
```
theangrydev, line 28:
```suggestion
    private static Connection createConnection(String serverURI) throws IOException, InterruptedException {
```

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
```diff
@@ -1,123 +1,142 @@
...
+        this(null);
     }
 
-    public NatsClient(String serverURI) {
-        if ((serverURI != null) && (!serverURI.isEmpty())) {
+    public NatsClient(String serverURI) throws IOException, InterruptedException {
+        if (serverURI == null || serverURI.isEmpty()) {
+            this.serverURI = "nats://localhost:4222";
+        }
+        else {
```
theangrydev: This is what I mean by a static factory method and a single responsibility constructor:
```suggestion
    private NatsClient(String serverURI, Connection natsConnection) {
        this.serverURI = serverURI;
        this. natsConnection = natsConnection;
    }
    
    public static NatsClient connectNatsClient(String serverURI) throws IOException, InterruptedException {
        Connection natsConnection = createConnection(serverURI);
        return new NatsClient(serverURI, natsConnection);
    }
```

I also think it's clearer if we remove the empty serverURI logic and make that explicit in the test

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
```diff
@@ -1,123 +1,142 @@
...
+    }
+
+    public void unsubscribe(Subscription subscription) {
+        // async subscriptions made with a dispatcher must be unsubscribed through the dispatcher
+        // sync subscriptions can be unsubscribed directly through the subscription
+        Dispatcher dispatcher = dispatcherBySubscription.remove(subscription);
+        if (dispatcher == null) {
+            subscription.unsubscribe();
+        }
+        else {
```
theangrydev, line 130: This looks like the Baeldung formatter is not being used, you can find it here:
https://github.com/Baeldung/dev-settings/tree/main/javaCodeStyle/intellij

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
```diff
@@ -1,123 +1,142 @@
...
-    private final Map<String, Subscription> subscriptions = new HashMap<>();
-
-    private final static Logger log = LoggerFactory.getLogger(NatsClient.class);
-
-    NatsClient() {
-        this.serverURI = "jnats://localhost:4222";
-        natsConnection = initConnection(serverURI);
+    public NatsClient() throws IOException, InterruptedException {
+        this(null);
     }
```
theangrydev:
```suggestion
```

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,124 @@
...
-
-        message = barSubscription.nextMessage(200);
-        assertNotNull("No message!", message);
-        assertEquals("hello back", new String(message.getData()));
-    }
-
-
-    private NatsClient connectClient() {
+    private NatsClient connectClient() throws IOException, InterruptedException {
         return new NatsClient();
```
theangrydev:
```suggestion
        return connectNatsClient("nats://localhost:4222");
```

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,124 @@
...
-
-        SyncSubscription fooSubscription = client.subscribeSync("foo.*");
-
-        client.publishMessage("foo.bar", "bar.foo", "hello there");
-
-        Message message = fooSubscription.nextMessage(200);
-        assertNotNull("No message!", message);
-        assertEquals("hello there", new String(message.getData()));
+    public void givenMessageExchangeViaPublishSubscribe_MessagesReceived() throws Exception {
+
```
theangrydev, line 23: Remove the blank lines at the start of methods

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,124 @@
...
+        try (NatsClient client = connectClient()) {
+            Subscription replySideSubscription = client.subscribeSync("requestSubject");
+            Subscription publishSideSubscription = client.subscribeSync("replyToSubject");
+            client.publishMessageWithReply("requestSubject", "replyToSubject", "hello there");
+
+            Message message = replySideSubscription.nextMessage(200);
```
theangrydev, line 29: Can we extract a constant to make it clear what this parameter is? (same throughout these tests)
```suggestion
            Message message = replySideSubscription.nextMessage(TIMEOUT_MILLIS);
```

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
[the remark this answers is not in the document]
scottf: Fixed

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,124 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 
-import org.junit.Test;
-
-import io.nats.client.Message;
-import io.nats.client.SyncSubscription;
+import static org.junit.Assert.*;
 
 public class NatsClientLiveTest {
```
theangrydev, line 15: Let's add a brief comment explaining that the test assumes a running server and link to https://docs.nats.io/nats-server/installation

theangrydev (changes requested):

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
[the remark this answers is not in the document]
theangrydev: No, this is not correct because:
1. There is no _whenXXX
2. There is no _thenXXX
3. _MessageReceived is not camelCase
4. givenXXX_thenXXX is not a valid pattern, it's always given_when_then or when_then, never just given_then

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
scottf: I changed this whole thing.

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
scottf: This is logic in the constructor, so I'm not sure what you mean.
```
Connection natsConnection = createConnection(serverURI);
```
Anyway, I changed it to only take a Connection in the constructor and provided a static method to make a connection. This is probably a more normal way this client would be done anyway, since making a connection can get very involved.

libraries-3/src/main/java/com/baeldung/jnats/NatsClient.java
scottf: Fixed.

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: yes, will do

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: fixing

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,128 @@
...
+            client.publishMessage("foo.bar.plop", "hello there");
 
+            message = greaterSubscription.nextMessage(TIMEOUT_MILLIS);
+            assertNotNull(message, "No message!");
+            assertEquals("hello there", new String(message.getData()));
+        }
     }
 
     @Test
     public void givenQueueMessage_OnlyOneReceived() throws Exception {
```
theangrydev, line 106: Test names still not matching BDD syntax

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,128 @@
...
+            message = fooGreaterSubscription.nextMessage(TIMEOUT_MILLIS);
+            assertNotNull(message, "No message!");
+            assertEquals("hello foo greater", new String(message.getData()));
+        }
     }
 
-
     @Test
-    public void givenRequest_ReplyReceived() {
+    public void whenWildCardSubscription_andNotMatchTopic_NoMessageReceived() throws Exception {
```
theangrydev, line 87: Test names still not matching BDD syntax

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,128 @@
...
-        SyncSubscription fooSubscription = client.subscribeSync("foo.*");
-
-        client.publishMessage("foo.bar.plop", "bar.foo", "hello there");
-
-        Message message = fooSubscription.nextMessage(200);
-        assertNull("Got message!", message);
-
-
-        SyncSubscription barSubscription = client.subscribeSync("foo.>");
+    public void whenWildCardSubscription_andMatchTopic_MessageReceived() throws Exception {
```
theangrydev, line 67: Test names still not matching BDD syntax

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,128 @@
...
+            assertNotNull(message, "No message!");
+            assertEquals("hello back", new String(message.getData()));
+        }
     }
 
     @Test
-    public void whenWildCardSubscription_andMatchTopic_MessageReceived() throws Exception {
-
-        NatsClient client = connectClient();
+    public void givenMessageExchangeViaRequestReply_MessagesReceived() throws Exception {
```
theangrydev, line 49: Test names still not matching BDD syntax

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,128 @@
...
-        assertEquals("hello back", new String(message.getData()));
+    private NatsClient connectClient() throws IOException, InterruptedException {
+        return new NatsClient(NatsClient.createConnection("nats://localhost:4222"));
     }
 
-
-    private NatsClient connectClient() {
-        return new NatsClient();
+    @Test
+    public void givenMessageExchangeViaPublishSubscribe_MessagesReceived() throws Exception {
```
theangrydev, line 31: Test names still not matching BDD syntax

theangrydev (changes requested):
Test names still not matching BDD syntax + we have a lot of unused methods in  NatsClient - can we either add tests for them, or remove the methods

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: This test and name was unchanged, so none of this is on me. I will fix though.

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: This was an original test/name. I will change.

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: This was an original test/name. I will change.

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: Will fix

theangrydev (changes requested):
Thanks, only concern remaining is the unused methods

scottf:
> Thanks, only concern remaining is the unused methods
I'll add some more tests to cover these.

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,202 @@
...
+            List<Message> messages = new ArrayList<>();
+            Subscription qSub1 = client.subscribeAsyncInQueueGroup("mySubject", "myQueue", messages::add);
+            Subscription qSub2 = client.subscribeAsyncInQueueGroup("mySubject", "myQueue", messages::add);
 
-        message = barSubscription.nextMessage(200);
-        assertNotNull("No message!", message);
-        assertEquals("hello there", new String(message.getData()));
+            client.publishMessage("mySubject", "data");
 
+            Thread.sleep(WAIT_MILLIS); // give the messages time to be published and received
```
theangrydev, line 110: Let's avoid Thread.sleep as much as possible so the test is not flaky

```suggestion
 @Test
    public void whenSubscribingAsynchronouslyWithQueueGroups_thenOnlyOneMessageHandlerInTheGroupShouldReceiveEachMessage() throws Exception {
        try (NatsClient client = connectClient()) {
            CountDownLatch awaitMessages = new CountDownLatch(1);
            List<Message> messages = new ArrayList<>();
            MessageHandler add = message -> {
                awaitMessages.countDown();
                messages.add(message);
            };
            Subscription qSub1 = client.subscribeAsyncInQueueGroup("mySubject", "myQueue", add);
            Subscription qSub2 = client.subscribeAsyncInQueueGroup("mySubject", "myQueue", add);

            client.publishMessage("mySubject", "data");

            awaitMessages.await();
            Thread.sleep(WAIT_MILLIS); // wait a bit longer to prove only one message is received
            assertEquals(1, messages.size());

            client.unsubscribe(qSub1);
            client.unsubscribe(qSub2);
        }
    }
```

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,202 @@
...
+            Subscription sub1 = client.subscribeAsync("mySubject", messages1::add);
+            Subscription sub2 = client.subscribeAsync("mySubject", messages2::add);
 
-        client.publishMessage("foo.bar.plop", "bar.foo", "hello there");
+            client.publishMessage("mySubject", "data1");
+            client.publishMessage("mySubject", "data2");
 
-        Message message = fooSubscription.nextMessage(200);
-        assertNull("Got message!", message);
+            Thread.sleep(WAIT_MILLIS); // give the messages time to be published and received
```
theangrydev, line 91: Let's avoid Thread.sleep
```suggestion
@Test
    public void whenSubscribingAsynchronouslyWithoutQueueGroups_thenEachMessageHandlerShouldReceiveEachMessage() throws Exception {
        try (NatsClient client = connectClient()) {
            CountDownLatch awaitMessages1 = new CountDownLatch(2);
            List<Message> messages1 = new ArrayList<>();
            MessageHandler add1 = message -> {
                awaitMessages1.countDown();
                messages1.add(message);
            };

            CountDownLatch awaitMessages2 = new CountDownLatch(1);
            List<Message> messages2 = new ArrayList<>();
            MessageHandler add2 = message -> {
                awaitMessages2.countDown();
                messages2.add(message);
            };

            Subscription sub1 = client.subscribeAsync("mySubject", add1);
            Subscription sub2 = client.subscribeAsync("mySubject", add2);

            client.publishMessage("mySubject", "data1");
            client.publishMessage("mySubject", "data2");

            awaitMessages1.await();
            awaitMessages2.await();
            assertEquals(2, messages1.size());
            assertEquals(2, messages2.size());

            client.unsubscribe(sub1);
            client.unsubscribe(sub2);
        }
    }
```

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,202 @@
...
+            Subscription sub2 = client.subscribeAsync("mySubject", messages2::add);
 
-        message = barSubscription.nextMessage(200);
-        assertNotNull("No message!", message);
-        assertEquals("hello back", new String(message.getData()));
-    }
+            client.publishMessage("mySubject", "data1");
+            client.publishMessage("mySubject", "data2");
 
+            Thread.sleep(WAIT_MILLIS); // give the messages time to be published and received
```
theangrydev, line 43: Let's avoid Thread.sleep
```suggestion
@Test
    public void whenSubscribingSynchronouslyWithoutQueueGroups_thenEachSubscriptionShouldReceiveEachMessage() throws Exception {
        try (NatsClient client = connectClient()) {
            CountDownLatch awaitMessages1 = new CountDownLatch(2);
            List<Message> messages1 = new ArrayList<>();
            MessageHandler add1 = message -> {
                awaitMessages1.countDown();
                messages1.add(message);
            };

            CountDownLatch awaitMessages2 = new CountDownLatch(1);
            List<Message> messages2 = new ArrayList<>();
            MessageHandler add2 = message -> {
                awaitMessages2.countDown();
                messages2.add(message);
            };

            Subscription sub1 = client.subscribeAsync("mySubject", add1);
            Subscription sub2 = client.subscribeAsync("mySubject", add2);

            client.publishMessage("mySubject", "data1");
            client.publishMessage("mySubject", "data2");

            awaitMessages1.await();
            awaitMessages2.await();
            assertEquals(2, messages1.size());
            assertEquals(2, messages2.size());

            client.unsubscribe(sub1);
            client.unsubscribe(sub2);
        }
    }
```

theangrydev (changes requested):
Looks good, but let's not encourage readers to write tests involving Thread.sleep, I refactored to use CountDownLatch instead, see the suggested comments

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: So I understand the point of not using the sleep whenever possible, but it actually reduces the point of the test, because you are asserting a fact that you have forced by an external behavior in the test.

I think a better answer would be to clearly document why the sleep is being used.

scottf:
I added this note to tests:

IMPORTANT: Using Thread.sleep() is not a generally recommended development concept.
In the case of tests, it is the easiest way to simulate the real behavior
that messages take some amount of time to go from being published
to being received. This amount of time will vary depending on many factors including:
* network latency
* server cluster configuration
* server computing power relative to the load of work

theangrydev:
@scottf the latch approach is not "forced by an external behavior in the test", if you introduced a bug in the production code then the test would fail (time out)

The problem with the Thread.sleep approach is that the test becomes flaky and non deterministic via false negative failures - we can't guarantee how long we need to sleep - I don't want to encourage readers to write flaky tests that could cause them issues in CI. And I also don't want to encourage putting very large sleep values in tests.

If you have an alternative suggestion to avoid writing flaky tests happy to hear it

theangrydev (changes requested):
Need a solution for the Thread.sleep tests that is deterministic

theangrydev:
We've also used awaitility https://www.baeldung.com/java-thread-sleep-vs-awaitility-await in the past for these kinds of cases - a polling based solution that abstracts away the waiting logic

scottf:
Replaced Thread.sleep with Awaitility.await

theangrydev (approved):

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,336 @@
...
+        @Override
+        public void exceptionOccurred(Connection conn, Exception exp) {
+            System.err.println("Exception Occurred: " + exp);
+        }
     }
 
+//    public static void main(String[] args) throws Exception {
+//        Connection conn = createConnection("nats://localhost:4222");
+//        Thread.sleep(60000);
+//    }
```
theangrydev, line 91: Remove
```suggestion
```

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,336 @@
...
+    public static final int POLL_DELAY_MILLIS = 50;
+
+    private static Connection createConnection() throws IOException, InterruptedException {
+        return createConnection("nats://localhost:4222");
+    }
 
-        NatsClient client = connectClient();
+    private static Connection createConnection(String uri) throws IOException, InterruptedException {
+        Options options = new Options.Builder().server(uri)
+            .connectionListener((connection, event) -> System.out.println("Connection Event: " + event))
```
theangrydev, line 56: We use SLF4J for logging in tests, either change to a logger or remove

(similar in a few other places throughout the test)

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,336 @@
...
+ * </ul>
+ * </p>
+ */
 public class NatsClientLiveTest {
 
-    @Test
-    public void givenMessageExchange_MessagesReceived() throws Exception {
+    public static final int TIMEOUT_MILLIS = 200;
+    public static final int WAIT_AT_MOST_MILLIS = 300;
+    public static final int POLL_DELAY_MILLIS = 50;
```
theangrydev, line 48: private static final

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,336 @@
...
+        try (Connection natsConnection = createConnection()) {
+            List<Message> messages = new ArrayList<>();
+
+            Dispatcher dispatcher = natsConnection.createDispatcher();
+            Subscription qSub1 = dispatcher.subscribe("mySubject", "myQueue", messages::add);
+            Subscription qSub2 = dispatcher.subscribe("mySubject", "myQueue", messages::add);
+
+            natsConnection.publish("mySubject", "data".getBytes());
+
+            // Simulate real world time for messages to propagate.
```
theangrydev, line 201: We aren't really simulating anything here, we are literally waiting for the message to arrive
```suggestion
            // wait for messages to propagate.
```

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,336 @@
...
+            fail("There should not be an exception");
+        }
     }
 
     @Test
-    public void whenWildCardSubscription_andMatchTopic_MessageReceived() throws Exception {
+    public void givenNoSubscribersAndReportNoResponders_whenRequestReply_thenReceiveNoResponders() {
+        try (Connection natsConnection = createConnectionWithReportNoResponders()) {
+            natsConnection.publish("noOneIsSubscribed", null);
+            CompletableFuture<Message> f = natsConnection.request("noOneIsSubscribed", null);
```
theangrydev, line 106:
```suggestion
            CompletableFuture<Message> future = natsConnection.request("noOneIsSubscribed", null);
```

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,336 @@
...
+            f.get(WAIT_AT_MOST_MILLIS, TimeUnit.MILLISECONDS);
+            fail("There should be an exception");
+        }
+        catch (ExecutionException e) {
+            Throwable cause = e.getCause();
+            assertNotNull(cause);
+            assertInstanceOf(JetStreamStatusException.class, cause);
+            assertTrue(cause.getMessage().contains("503 No Responders Available For Request"));
```
theangrydev, line 114: For complex assertions we use AssertJ, which has helpers like:
https://www.baeldung.com/assertj-exception-assertion

A test like this would usually be written like:
> assertThatThrownBy(() -> {
> }).hasCauseInstanceOf(JetStreamStatusException.class)
>  .hasMessageContaining("...");

Avoiding the need for the manual fail("There should be an exception");

(similar in a few other places throughout the test)

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,336 @@
...
+ * that messages take some amount of time to go from being published
+ * to being received. This amount of time will vary depending on many factors including:
+ * <ul>
+ *     <li>network latency</li>
+ *     <li>server cluster configuration</li>
+ *     <li>server computing power relative to the load of work</li>
+ * </ul>
+ * </p>
+ */
 public class NatsClientLiveTest {
```
theangrydev, line 44: Apply the Baeldung formatter to this file:
https://github.com/Baeldung/dev-settings/tree/main/javaCodeStyle/intellij

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,336 @@
...
-    private NatsClient connectClient() {
-        return new NatsClient();
+    @Test
+    public void givenNoSubscribers_whenPublishingMessages_thenItCannotBeDeterminedIfMessageIsReceived() {
+        try (Connection natsConnection = createConnection()) {
+            natsConnection.publish("mySubject", "myData".getBytes());
+        }
+        catch (Exception e) {
+            fail("There should not be an exception");
+        }
```
theangrydev, line 99: We can just let the test fail with the exception itself if you remove the catch clause
```suggestion
```

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,336 @@
...
 
-        NatsClient client = connectClient();
+    @Test
+    public void whenSubscribingSynchronouslyWithoutQueueGroups_thenEachSubscriptionShouldReceiveEachMessage() throws Exception {
+        try (Connection natsConnection = createConnection()) {
+            Subscription subscription1 = natsConnection.subscribe("mySubject");
+            Subscription subscription2 = natsConnection.subscribe("mySubject");
 
-        SyncSubscription fooSubscription = client.subscribeSync("foo.*");
+            natsConnection.publish("mySubject", "data".getBytes());
```
theangrydev, line 126: We should really be passing a character set like StandardCharSets.UTF_8 whenever we do String.getBytes and new String, to ensure the code is portable

(similar in a few other places throughout the test)

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,336 @@
...
+            List<Message> messages2 = new ArrayList<>();
+
+            Dispatcher dispatcher = natsConnection.createDispatcher();
+            Subscription subscription1 = dispatcher.subscribe("mySubject", messages1::add);
+            Subscription subscription2 = dispatcher.subscribe("mySubject", messages2::add);
+
+            natsConnection.publish("mySubject", "data1".getBytes());
+            natsConnection.publish("mySubject", "data2".getBytes());
+
+            // Simulate real world time for messages to propagate.
```
theangrydev, line 176: We aren't really simulating anything here, we are literally waiting for the message to arrive
```suggestion
            // wait for messages to propagate.
```

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,336 @@
...
-
+import io.nats.client.Nats;
+import io.nats.client.Options;
+import io.nats.client.Subscription;
+
+/**
+ * All the tests in this class require that a NATS server be running on localhost at the default port.
+ * See {@link <a href="https://docs.nats.io/nats-server/installation">Installing a NATS Server</a>}.
+ * <p>
+ * IMPORTANT: Awaitility.await is used to simulate the real behavior
```
theangrydev, line 34: We aren't really simulating anything here, we are literally waiting for the message to arrive
```suggestion
 * IMPORTANT: Awaitility.await is used to account for the real behavior
```

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,336 @@
...
+            assertNotNull(message, "No message!");
+            assertEquals("Please respond!", new String(message.getData()));
+            natsConnection.publish(message.getReplyTo(), "Message Received By Subscription 2".getBytes());
+
+            int responsesFrom1 = 0;
+            int responsesFrom2 = 0;
+            message = publishSideSubscription.nextMessage(TIMEOUT_MILLIS);
+            assertNotNull(message, "No message!");
+            String response = new String(message.getData());
+            assertTrue(response.contains("Message Received"));
```
theangrydev, line 237: For complex assertions we use AssertJ, which has helpers like:
> assertThat(response).contains("..");

And usually if we use AssertJ at all we would use it everywhere for consistency

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,336 @@
...
+            message = publishSideSubscription.nextMessage(TIMEOUT_MILLIS);
+            assertNotNull(message, "No message!");
+            response = new String(message.getData());
+            assertTrue(response.contains("Message Received"));
+            if (response.contains("1")) {
+                responsesFrom1++;
+            }
+            else if (response.contains("2")) {
+                responsesFrom2++;
+            }
```
theangrydev, line 254: This can be written as a 2 element for loop like:

for (int i = 0; i < 2; i++ ) {
    // duplicated receive logic
}

theangrydev (changes requested):

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: removed

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: This is not logging. This is just to demonstrate that an event can be handled. I don't want to do logging because I don't want the user to get the idea that this is the purpose of this construct.

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: fixed

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: fixed

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: fixed all instances with
```
getBytes(StandardCharsets.UTF_8)
```

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: fixed

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: fixed

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: I give up. Is this really that big of a deal, it would require going back through every single test? These assertions are not complex.

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: Why is that better? Do you want me to use an array instead of `responsesFrom1` and `responsesFrom2` I think this is clearer.

theangrydev: It's unclear at a glance if this is the same pattern repeated twice, or there is some subtle difference, the loop idea (or a helper method) would clarify that it's the same

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
theangrydev: I can link you to our guidelines on this and show you what my reference point is. We have guidelines in order to achieve consistency between all of the articles on the site.

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
theangrydev: And the same for new String

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
theangrydev: The guideline we have is that, we cannot use System.out.println in an article unless the article is specifically about testing System.out

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
theangrydev: Alternatively we can wait until final review (when another person checks over the PR and article) and let them decide any points you would like not to do

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: fixed

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: fixed

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
scottf: made conversions helper functions for all string to bytes and bytes to string

libraries-3/src/test/java/com/baeldung/jnats/NatsClientLiveTest.java
```diff
@@ -1,115 +1,337 @@
...
+            Subscription segmentStarSubscription = natsConnection.subscribe("segment.*");
 
-        NatsClient client = connectClient();
+            natsConnection.publish("segment.star", convertStringToBytes("hello segment star"));
 
-        SyncSubscription queue1 = client.joinQueueGroup("foo.bar.requests", "queue1");
-        SyncSubscription queue2 = client.joinQueueGroup("foo.bar.requests", "queue1");
+            Message message = segmentStarSubscription.nextMessage(TIMEOUT_MILLIS);
+            assertNotNull(message, "No message!");
+            assertEquals("hello segment star", new String(message.getData()));
```
theangrydev, line 302: We still have a few new String() usages without a charset specified

theangrydev (changes requested):

theangrydev (approved):

PR #16102, opened 2024-03-13
JAVA-32170: Changes made for Fix formatting of POMs

Bipinkumar27:


aws-modules/pom.xml
```diff
@@ -5,23 +5,14 @@
     <modelVersion>4.0.0</modelVersion>
     <artifactId>aws-modules</artifactId>
     <name>aws-modules</name>
```
anastasiosioannidis, line 7: should packaging be before name?

PR #16137, opened 2024-03-15
Calc percentile

sk1418:


core-java-modules/core-java-lang-math-4/src/test/java/com/baeldung/percentile/CalculatePercentileUnitTest.java
```diff
@@ -0,0 +1,75 @@
...
+    void whenCallingGetPercentileWithAnArray_thenGetExpectedResult() {
+
+        //prepare the input array
+        long[] theArray = LongStream.of(-1, 200, 30, 42, -5, 7, 8, 92)
+            .toArray();
+
+        //convert the long[] array to a List<Long>
+        List<Long> list8 = Arrays.stream(theArray)
+            .boxed()
+            .toList();
```
theangrydev, line 68: This is an O(n) operation, but if we use Arrays.asList instead then it's O(1)

core-java-modules/core-java-lang-math-4/src/test/java/com/baeldung/percentile/CalculatePercentileUnitTest.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.LongStream;
+
+import org.junit.jupiter.api.Test;
+
+public class CalculatePercentileUnitTest {
```
theangrydev, line 16: Let's show the 0th and 100th percentiles in the test too just as a boundary sanity check

theangrydev (changes requested):

core-java-modules/core-java-lang-math-4/src/test/java/com/baeldung/percentile/CalculatePercentileUnitTest.java
sk1418: Are you suggesting using `Long[]` instead of `long[]` as the example, and then `Arrays.asList(theLongArray)`?

theangrydev: Sorry my bad you're right, ignore this one

core-java-modules/core-java-lang-math-4/src/test/java/com/baeldung/percentile/CalculatePercentileUnitTest.java
sk1418: tests added

theangrydev (approved):

PR #16139, opened 2024-03-15
JAVA-31001: Upgrade spring-boot-springdoc to boot 3

amit2103:


alvin577austria:
Please also uncomment the module in the spring-boot-modules pom.xml

spring-boot-modules/spring-boot-springdoc/pom.xml
```diff
@@ -10,9 +10,10 @@
     <description>Project for Springdoc integration</description>
 
     <parent>
-        <groupId>com.baeldung.spring-boot-modules</groupId>
-        <artifactId>spring-boot-modules</artifactId>
-        <version>1.0.0-SNAPSHOT</version>
+        <groupId>com.baeldung</groupId>
```
alvin577austria, line 13: Please update similar with the other modules under spring-boot-modules.

alvin577austria (changes requested):

spring-boot-modules/spring-boot-springdoc/pom.xml
amit2103: Thanks Alvin. Updated

alvin577austria (approved):

PR #16146, opened 2024-03-16
BAEL-7600: Get last n characters from a String

danielmcnally285:
Get last n characters from a String unit tests


core-java-modules/core-java-string-operations-8/src/test/java/com/baeldung/lastncharacters/LastNCharactersUnitTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Test
+    void givenString_whenUsingTwoArgSubstringMethod_thenObtainLastNCharacters() {
+        int beginIndex = s.length() - n;
+        String result = s.substring(beginIndex, s.length());
+
+        assertThat(result).isEqualTo("2024");
+    }
+
+    @BeforeEach
```
vinipx, line 59: let's move the @BeforeEach to the beginning of the class

vinipx (changes requested):

core-java-modules/core-java-string-operations-8/src/test/java/com/baeldung/lastncharacters/LastNCharactersUnitTest.java
danielmcnally285: Moved as suggested

PR #16147, opened 2024-03-16
[improvement-merging-maps] merge maps into one

sk1418:


core-java-modules/core-java-streams-simple/src/test/java/com/baeldung/streams/flatmap/map/Java8MapAndFlatMapUnitTest.java
```diff
@@ -48,4 +49,35 @@ void givenOptional_whenCalledFlatMap_thenProduceFlattenedOptional() {
             .flatMap(s -> Optional.of("STRING")));
     }
 
-}
+    @Test
+    void givenMaps_whenCalledFlatMap_thenMultipleMapsMergedIntoOneMap() {
```
theangrydev, line 53: Can we add one more test to show the limitation of this approach when there are duplicate keys, if I remember correctly Collectors.toMap will throw an exception on duplicate keys

theangrydev: Similar for null keys I think

PR #16149, opened 2024-03-16
JAVA-29183 :- Upgrade spring-5-webflux & spring-5-webflux-2 to Spring Boot 3.

amit2103:
…

### DIFF
--- a/spring-5-webflux-2/pom.xml
+++ b/spring-5-webflux-2/pom.xml
@@ -10,9 +10,9 @@
 
     <parent>
         <groupId>com.baeldung</groupId>
-        <artifactId>parent-boot-2</artifactId>
+        <artifactId>parent-boot-3</artifactId>
         <version>0.0.1-SNAPSHOT</version>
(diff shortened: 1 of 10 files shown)


spring-5-webflux-2/pom.xml
```diff
@@ -114,8 +114,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
```
anastasiosioannidis, line 114: Can you remove the whole maven-compiler-plugin dependency and try building?

spring-5-webflux/pom.xml
```diff
@@ -90,12 +90,22 @@
                 <groupId>org.springframework.boot</groupId>
                 <artifactId>spring-boot-maven-plugin</artifactId>
             </plugin>
```
anastasiosioannidis, line 92: Can you try removing this also?

amit2103: Thanks done, also did for the other one.

PR #16157, opened 2024-03-18
BAEL-7677: How to fix Hibernate UnknownEntityException: Could not resolve root entity

davidmbochi:


persistence-modules/hibernate-exceptions/src/main/java/com/baeldung/hibernate/unknownentityexception/Customer.java
```diff
@@ -0,0 +1,41 @@
+package com.baeldung.hibernate.unknownentityexception;
+
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+
+@Entity
+@Table(name = "customer")
+public class Customer {
```
theangrydev, line 10: Please apply the Baeldung formatter to all files, you can find it here:
https://github.com/Baeldung/dev-settings/tree/main/javaCodeStyle/intellij

persistence-modules/hibernate-exceptions/src/test/java/com/baeldung/hibernate/unknownentityexception/UnknownEntityExceptionUnitTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    session.getTransaction().commit();
+    session.close();
+
+    session = sessionFactory.openSession();
+    session.beginTransaction();
+
+    assertThrows(IllegalArgumentException.class, () -> {
+      session.createQuery("FROM customer", Customer.class);
+    });
```
theangrydev, line 58: Let's assert on the exception message, and the cause type too to make it clear we are actually addressing the org.hibernate.query.sqm.UnknownEntityException problem that the article is about

You can use AssertJ for those assertions, it has nice helpers for exception assertions:
https://www.baeldung.com/assertj-exception-assertion

persistence-modules/hibernate-exceptions/src/test/java/com/baeldung/hibernate/unknownentityexception/UnknownEntityExceptionUnitTest.java
```diff
@@ -0,0 +1,86 @@
...
+  public void whenDatabaseTableNameUsed_thenExceptionIsThrown() {
+    Customer customer = new Customer();
+    customer.setEmail("john@baeldung.com");
+    session.persist(customer);
+
+    session.getTransaction().commit();
+    session.close();
+
+    session = sessionFactory.openSession();
+    session.beginTransaction();
```
theangrydev, line 54: To simplify the examples, is there an autocommit feature we can enable? Just to keep things focused on the key steps to reproduce

theangrydev (changes requested):

persistence-modules/hibernate-exceptions/src/test/java/com/baeldung/hibernate/unknownentityexception/UnknownEntityExceptionUnitTest.java
davidmbochi: Yes. I have made the following changes to enable the autocommit.
```java
 @BeforeEach
    public void setUp() {
        session = sessionFactory.openSession();
        session.beginTransaction();
        session.doWork(connection -> connection.setAutoCommit(true));
    }
```

theangrydev (approved):

PR #16163, opened 2024-03-18
[SR] Map enum with mapstruct

sagaofsilence:
Source code and tests for mapping enums with MapStruct.


sagaofsilence:
@eric-martin Please review the PR for the article.

PR #16167, opened 2024-03-18
JAVA-30337 Fix failing module spring-security-web-boot-2 schema for s…

timis1:
…pring-security-login-redirect

### DIFF
--- a/spring-security-modules/spring-security-web-boot-2/src/main/java/com/baeldung/loginredirect/LoginRedirectSecurityConfig.java
+++ b/spring-security-modules/spring-security-web-boot-2/src/main/java/com/baeldung/loginredirect/LoginRedirectSecurityConfig.java
@@ -13,8 +13,9 @@ import org.springframework.security.provisioning.InMemoryUserDetailsManager;
 import org.springframework.security.web.SecurityFilterChain;
 import org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter;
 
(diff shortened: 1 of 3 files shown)


spring-security-modules/spring-security-web-boot-2/src/main/java/com/baeldung/loginredirect/LoginRedirectSecurityConfig.java
alvin577austria, line 1: Can you please do a cleanup of the imports please, remove unused ones. Thanks!

alvin577austria (changes requested):

spring-security-modules/spring-security-web-boot-2/src/main/java/com/baeldung/loginredirect/LoginRedirectSecurityConfig.java
timis1: That imports that looks as unused are for that annotations that I commented in this class. I would suggest to leave as it is, and in case you need to run Java class configuration just to un-comment that lines without needed to import some annotations.

alvin577austria: ok, let's leave it as is then.

alvin577austria (approved):

PR #16215, opened 2024-03-24
JAVA-32064: Changes made for Upgrade spring-session-redis to Spring B…

Bipinkumar27:
…oot 3

### DIFF
--- a/spring-web-modules/spring-session/spring-session-redis/src/main/java/com/baeldung/spring/session/SecurityConfig.java
+++ b/spring-web-modules/spring-session/spring-session-redis/src/main/java/com/baeldung/spring/session/SecurityConfig.java
@@ -1,9 +1,12 @@
 package com.baeldung.spring.session;
 
+import static org.springframework.security.config.Customizer.withDefaults;
+
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+import org.springframework.security.config.http.SessionCreationPolicy;
 import org.springframework.security.core.userdetails.User;
 import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.security.crypto.password.PasswordEncoder;
@@ -25,13 +28,12 @@ public class SecurityConfig {
 
     @Bean
     public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
-        http.httpBasic()
-            .and()
-            .authorizeRequests()
-            .antMatchers("/")
-            .hasRole("ADMIN")
-            .anyRequest()
-            .authenticated();
+        http.httpBasic(withDefaults())
+            .sessionManagement(httpSecuritySessionManagementConfigurer -> httpSecuritySessionManagementConfigurer.sessionCreationPolicy(SessionCreationPolicy.ALWAYS))
+            .authorizeRequests((authorizeRequests) -> authorizeRequests.requestMatchers("/")
+                .hasRole("ADMIN")
+                .anyRequest()
+                .authenticated());
         return http.build();
     }
 }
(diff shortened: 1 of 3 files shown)


spring-web-modules/spring-session/spring-session-redis/src/main/java/com/baeldung/spring/session/SecurityConfig.java
```diff
@@ -25,13 +28,12 @@ public InMemoryUserDetailsManager userDetailsService(PasswordEncoder passwordEnc
 
     @Bean
     public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
-        http.httpBasic()
-            .and()
-            .authorizeRequests()
-            .antMatchers("/")
-            .hasRole("ADMIN")
-            .anyRequest()
-            .authenticated();
```
anastasiosioannidis, line 34: Hi @Bipinkumar27 

Could you analyze:
 .sessionManagement(httpSecuritySessionManagementConfigurer -> httpSecuritySessionManagementConfigurer.sessionCreationPolicy(SessionCreationPolicy.ALWAYS))
 
 Have you tested to check this works as before?

Bipinkumar27: Yes @anastasiosioannidis,

In Spring Security 6, the SecurityContextPersistenceFilter and SessionManagementFilter are not set by default.
It required in our use case of storing the session data in Redis in SessionControlIntegrationTest Inside testRedisControlsSession() method 

After doing the above changes it got fixed and test executed successfully

Bipinkumar27: However, currently I have set the session creation policy to AWAYS which we can change to IF_REQUIRED.

httpSecuritySessionManagementConfigurer.sessionCreationPolicy(SessionCreationPolicy.ALWAYS) 

Let me know if that needs to change.

anastasiosioannidis: Hi @Bipinkumar27 

Since before it was if-required, I guess we could revert to this.
Please check if tests work.

anastasiosioannidis: Also check if we can remove the lambda and simplify it

Bipinkumar27: Hi @anastasiosioannidis 
As highlighted earlier, it's imperative to ensure explicit configuration of session management post-transition to Spring 6, as adopted in parent-boot-3. This became necessary after a test case broke due to improper session management configuration.

Regarding simplification through lambda expressions, although it may lead to larger method sizes, it's noteworthy that lambda expressions are widely utilized in Spring documentation. Hence, it would be beneficial for us to adopt the same practice, aligning with Spring standards.

Please share your views

anastasiosioannidis: Hi @Bipinkumar27 

Thanks for the analysis.
I get build/test errors for the integration profile (Redis couldn't start). Could you share how you ran it?

Bipinkumar27: Hey @anastasiosioannidis,

I have run it by using the below cmd:

`mvn clean install -U -Pintegration-jdk9-and-above --pl spring-web-modules/spring-session/spring-session-redis`

Screenshot 2024-04-05 at 7 24 49 PM



Also, I have run this on Inteij IDE :
Screenshot 2024-04-05 at 7 27 01 PM


I am not sure why Redis is not running on your local machine may be due to Redis default port not might occupied by other task.

PR #16235, opened 2024-03-26
JAVA-30184: Changes made for Upgrade junit-jupiter library in the mai…

Bipinkumar27:
…n pom.xml


ksqldb/pom.xml
```diff
@@ -45,6 +45,7 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <!--check this -->
```
alvin577austria, line 48: Can you check please if we can update this to the latest version.

libraries/pom.xml
```diff
@@ -385,7 +385,7 @@
         <rcaller.version>3.0</rcaller.version>
         <rserve.version>1.8.1</rserve.version>
         <agrona.version>1.17.1</agrona.version>
-        <exec-maven-plugin.version>1.6.0</exec-maven-plugin.version>
+        <!--<exec-maven-plugin.version>1.6.0</exec-maven-plugin.version>-->
```
alvin577austria, line 388: Why do we need to comment this out?

microservices-modules/micronaut/pom.xml
```diff
@@ -80,6 +80,7 @@
             <scope>compile</scope>
         </dependency>
         <dependency>
+            <!--check once -->
```
alvin577austria, line 83: Can you check please if we can update this to the latest version.

alvin577austria (changes requested):

ksqldb/pom.xml
Bipinkumar27: Updated

libraries/pom.xml
Bipinkumar27: @alvin577austria the parent of this module already has the latest version of 3.1.0. I was just checking if that will work and if we are getting build errors if we use parent updated version.
I am removing this as this is working fine.

alvin577austria (approved):

alvin577austria (approved):

messaging-modules/spring-apache-camel/pom.xml
```diff
@@ -77,7 +77,7 @@
     <properties>
         <camel.version>4.3.0</camel.version>
         <camel-swagger.version>3.22.0</camel-swagger.version>
-        <junit-jupiter.version>5.10.1</junit-jupiter.version> <!--Version not compatible with the one inherited from parent-->
+        <junit-jupiter.version>5.10.2</junit-jupiter.version> <!--Version not compatible with the one inherited from parent-->
```
alvin577austria, line 80: The parent module does not seem to have any junit-jupiter property set, can you remove the comment here please. Thanks!

alvin577austria (changes requested):

alvin577austria (approved):

maven-modules/maven-exec-plugin/pom.xml
```diff
@@ -84,6 +84,7 @@
     <properties>
         <maven-compiler-plugin.version>3.12.1</maven-compiler-plugin.version>
         <exec-maven-plugin.version>3.1.0</exec-maven-plugin.version>
+        <junit-jupiter.version>5.10.0</junit-jupiter.version>
```
alvin577austria, line 87: Is it necessary to use 5.10.0 here? is 5.10.2 not working?

alvin577austria (changes requested):

maven-modules/maven-exec-plugin/pom.xml
alvin577austria: I can't remember if I have asked about the same and you have already answered before though.

alvin577austria (approved):

PR #16239, opened 2024-03-26
Deep vs Shallow Copy - sandysimonton1337@gmail.com

sandysimonton1337:
This is my trial submission. I would have done it differently, i.e., had a single commit instead of multiple because of downstream pulls, but this PR won't be merged anyway. This is the revised version, Thanks!


sandysimonton1337:
Closed per submission requirements

PR #16256, opened 2024-03-29
bael-7718

imsurajmishra:
Jira : https://jira.baeldung.com/browse/BAEL-7718


persistence-modules/spring-data-jpa-query-4/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+                <configuration>
+                    <source>17</source>
+                    <target>17</target>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+    <properties>
+        <javafaker.version>0.15</javafaker.version>
```
eric-martin, line 54: I don't think we are using this ... let's remove

PR #16313, opened 2024-04-06
Parallel Collectors with Virtual Threads

pivovarit:


libraries-stream/src/test/java/com/baeldung/parallel_collectors/ParallelCollectorsVirtualThreadsManualTest.java
```diff
@@ -0,0 +1,69 @@
...
+    public void processInParallelOnOSThreads() {
+        int parallelProcesses = 5_000;
+
+        var e = Executors.newFixedThreadPool(parallelProcesses);
+
+        var result = timed(() -> Stream.iterate(0, i -> i + 1).limit(parallelProcesses)
+          .collect(ParallelCollectors.parallel(i -> fetchById(i), toList(), e, parallelProcesses))
+          .join());
+
+        System.out.println(result);
```
acerasoni, line 28: Could we use slf4j with logback instead of Sys.out?

libraries-stream/src/test/java/com/baeldung/parallel_collectors/ParallelCollectorsVirtualThreadsManualTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Test
+    public void processInParallelOnVirtualThreads() {
```
acerasoni, line 32: Could we update the test names to comply with the BDD convention: givenX_whenY_thenZ?

libraries-stream/src/test/java/com/baeldung/parallel_collectors/ParallelCollectorsVirtualThreadsManualTest.java
```diff
@@ -0,0 +1,69 @@
...
+import java.time.Instant;
+import java.util.List;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+
+import static java.util.stream.Collectors.toList;
+
+public class ParallelCollectorsVirtualThreadsManualTest {
```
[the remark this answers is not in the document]
pivovarit: The `Manual` effectively excludes it - since those tests are creating unreasonable number of threads (both os and virtual), it's better to not stress tests the Jenkins worker

libraries-stream/src/test/java/com/baeldung/parallel_collectors/ParallelCollectorsVirtualThreadsManualTest.java
pivovarit: We could, but there's not a single use of slf4j in this module - do you still want to do the change?

libraries-stream/src/test/java/com/baeldung/parallel_collectors/ParallelCollectorsVirtualThreadsManualTest.java
pivovarit: Done

alvin577austria:
Hi, since this is for JDK21 is this module also expected to be excluded from integration-jdk9-and-above profile? I see it has been commented out from default-jdk9-and-above profile.

PR #16360, opened 2024-04-10
[fix-merging-overlapping] fix

sk1418:


algorithms-modules/algorithms-miscellaneous-8/src/main/java/com/baeldung/algorithms/mergeintervals/MergeOverlappingIntervals.java
```diff
@@ -1,29 +1,27 @@
...
-        intervals.sort((one, two) -> one.start - two.start);
-
-        // Create somewhere to put the merged list, start it off with the earliest starting interval
+        intervals.sort(Comparator.comparingInt(interval -> interval.start));
         ArrayList<Interval> merged = new ArrayList<>();
         merged.add(intervals.get(0));
 
-        // Loop over each interval and merge if start time is before the end time of the
-        // previous interval
         intervals.forEach(interval -> {
```
theangrydev, line 16: I'd personally prefer a for-each loop for (Interval interval : intervals) since we aren't doing anything functional here

theangrydev (approved):
Looks good, one personal preference up to you

algorithms-modules/algorithms-miscellaneous-8/src/main/java/com/baeldung/algorithms/mergeintervals/MergeOverlappingIntervals.java
sk1418: When we use `forEach(),` it's functional. The consumer lambda is a function. 😃 Well, I'm joking. I know what you meant.
I think using`Iterable.forEach()` is fine here. Java's lambdas are sometimes inconvenient (compared to Kotlin) in terms of error handling, effective final/final variable limitation, etc. If these inconveniences are not relevant to our implementation, I will take forEach() since it's more readable and natural to write. Besides, we've used comparingInt() with a lambda in the code, perhaps we could stick to that style? What do you think?

theangrydev: Sure your call here

PR #16437, opened 2024-04-17
Update DinoController.java

JavaSoftwareEntwickler:
With this fix, the controller returns the correct template value, otherwise the controller does not work.


kasramp:
Thanks @JavaSoftwareEntwickler for sending the patch.

PR #16439, opened 2024-04-18
Sachin bael 7827 mapping enum to string using mapstruct

sagaofsilence:
Code examples for article https://drafts.baeldung.com/?p=197211&preview=true


sagaofsilence:
Bradon, please review.
Code for https://drafts.baeldung.com/?p=197211&preview=true

mapstruct/src/main/java/com/baeldung/mapstruct/enumtostring/model/ExternalOrder.java
Bwvolleyball, line 1: I think we can remove this now.

PR #16442, opened 2024-04-18
BAEL-7783: Assert properties of custom exceptions with catchThrowableOfType in AssertJ

azhwani:
Explain how to assert properties of custom exceptions

### DIFF
--- a/testing-modules/assertion-libraries/src/main/java/com/baeldung/assertj/exceptions/CityUtils.java
+++ b/testing-modules/assertion-libraries/src/main/java/com/baeldung/assertj/exceptions/CityUtils.java
@@ -1,0 +1,17 @@
+package com.baeldung.assertj.exceptions;
+
+import java.util.Arrays;
+import java.util.List;
+
+public final class CityUtils {
+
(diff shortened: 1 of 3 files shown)


testing-modules/assertion-libraries/src/main/java/com/baeldung/assertj/exceptions/CityUtils.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+public final class CityUtils {
+
+    private static final List<String> CITIES = Arrays.asList("Tamassint", "Londn", "Madrid", "New york");
+
+    public static String search(String searchedCity) {
+        return CITIES.stream()
+          .filter(city -> searchedCity.equals(city))
```
JavierPeris, line 12: We can simplify with a method reference:
```suggestion
          .filter(searchedCity::equals)
```

azhwani: Done

PR #16477, opened 2024-04-23
JAVA-33989: Migrate k8s-operator to heavy profile

harry9656:


kubernetes-modules/pom.xml
```diff
@@ -17,7 +17,7 @@
         <module>k8s-admission-controller</module>
         <module>kubernetes-spring</module>
         <module>k8s-java-heap-dump</module>
-        <module>k8s-operator</module>
+<!--        <module>k8s-operator</module>--> <!-- Migrated to heavy profile -->
```
alvin577austria, line 20: I guess we can remove this here altogether.

alvin577austria (changes requested):

kubernetes-modules/pom.xml
harry9656: It is better to keep this comment. In the future, we will know where the definition of this module is.

alvin577austria (approved):

PR #16513, opened 2024-04-27
BAEL-6333: Avoiding the IndexOutOfBoundsException in subList in Java

azhwani:
+ Create a safe alternative of subList method to avoid IndexOutOfBoundsException

### DIFF
--- a/core-java-modules/core-java-collections-list-6/src/main/java/com/baeldung/sublistindexoutofboundsexception/SubListUtils.java
+++ b/core-java-modules/core-java-collections-list-6/src/main/java/com/baeldung/sublistindexoutofboundsexception/SubListUtils.java
@@ -1,0 +1,16 @@
+package com.baeldung.sublistindexoutofboundsexception;
+
+import java.util.Collections;
+import java.util.List;
+
+public class SubListUtils {
+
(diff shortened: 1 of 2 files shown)


core-java-modules/core-java-collections-list-6/src/main/java/com/baeldung/sublistindexoutofboundsexception/SubListUtils.java
```diff
@@ -0,0 +1,24 @@
...
+        int size = myList.size();
+        if (fromIndex >= size || toIndex <= 0 || fromIndex >= toIndex) {
+            return Collections.emptyList();
+        }
+
+        fromIndex = Math.max(0, fromIndex);
+        toIndex = Math.min(size, toIndex);
+
+        return myList.subList(fromIndex, toIndex);
+    }
```
JavierPeris, line 22: A simpler version of this method:
```suggestion
public static List<String> safeSubList(List<String> myList, int fromIndex, int toIndex) {
  if (myList == null || fromIndex >= myList.size() || toIndex <= 0 || fromIndex >= toIndex) {
    return Collections.emptyList();
  }
  return myList.subList(Math.max(0, fromIndex), Math.min(myList.size(), toIndex));
}
```

JavierPeris (approved):

PR #16527, opened 2024-04-30
optional to arraylist

wynnteo:


core-java-modules/core-java-collections-conversions-3/src/main/java/optionaltoarraylist/OptionalToArrayListConverter.java
```diff
@@ -0,0 +1,49 @@
+package optionaltoarraylist;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public class OptionalToArrayListConverter {
+
+    public static List usingIsPresent(Optional<String> obj) {
```
KevinGilmore, line 10: We should stick to List<String> throughout.

PR #16542, opened 2024-05-01
BAEL-7884: String replacement back reference vs lookaround

harry9656:


core-java-modules/core-java-string-operations-9/pom.xml
```diff
@@ -14,6 +14,17 @@
...
     <dependencies>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <version>1.37</version>
+        </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>1.37</version>
```
JavierPeris, line 25: Please, let's use properties instead of hardcode values.

PR #16566, opened 2024-05-03
JAVA-34026:Changes made for formatting xml

Bipinkumar27:


persistence-modules/pom.xml
```diff
@@ -113,7 +113,7 @@
         <module>spring-jpa-2</module>
         <module>spring-jpa-3</module>
         <module>spring-jdbc</module>
```
anastasiosioannidis, line 115: Have you added some space before <module>spring-jdbc-2</module> ?
Do we need it?

Bipinkumar27: yes it was wrongly indented earlier

PR #16574, opened 2024-05-04
BAEL-7932: CompletableFuture vs Mono

harry9656:


reactor-core/src/main/java/com/baeldung/reactor/completablefuturevsmono/CompletableFutureDemo.java
```diff
@@ -0,0 +1,26 @@
...
+                Thread.currentThread()
+                    .interrupt();
+            }
+            return "Finished completableFuture";
+        });
+
+        try {
+            completableFuture.get(); // This blocks the main thread
+        } catch (InterruptedException | ExecutionException e) {
+            e.printStackTrace();
```
JavierPeris, line 23: Same thing here.

PR #16579, opened 2024-05-04
BAEL-7887: Reading CSV file using Java 8 Stream API

azhwani:
Add new methods to read CSV files (I tried to keep the same structure as the existing examples: Junit 4, same assertion logic ...)

### DIFF
--- a/core-java-modules/core-java-io-conversions-2/src/test/java/com/baeldung/csv/ReadCSVInArrayUnitTest.java
+++ b/core-java-modules/core-java-io-conversions-2/src/test/java/com/baeldung/csv/ReadCSVInArrayUnitTest.java
@@ -1,11 +1,24 @@
 package com.baeldung.csv;
 
-import com.opencsv.CSVReader;
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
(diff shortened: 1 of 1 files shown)


core-java-modules/core-java-io-conversions-2/src/test/java/com/baeldung/csv/ReadCSVInArrayUnitTest.java
```diff
@@ -94,4 +98,51 @@ public void givenCSVFile_whenOpencsv_thenContentsAsExpected() throws IOException
...
+        List<List<String>> records = Files.lines(path)
+          .map(line -> Arrays.asList(line.split(COMMA_DELIMITER)))
+          .collect(Collectors.toList());
+
+        for (int i = 0; i < EXPECTED_ARRAY.size(); i++) {
+            Assert.assertArrayEquals(EXPECTED_ARRAY.get(i)
+                .toArray(),
+                records.get(i)
+                    .toArray());
+        }
```
JavierPeris, line 146: Here we can use try-with-resources:
```suggestion
        try (Stream<String> lines = Files.lines(Paths.get(CSV_FILE))) {
            List<List<String>> records = lines.map(line -> Arrays.asList(line.split(COMMA_DELIMITER)))
                .collect(Collectors.toList());
            for (int i = 0; i < EXPECTED_ARRAY.size(); i++) {
                Assert.assertArrayEquals(EXPECTED_ARRAY.get(i)
                    .toArray(), records.get(i)
                    .toArray());
            }
        }
```

azhwani: Done

PR #16582, opened 2024-05-04
[BAEL-7825] changes of sorting a map contains list of elements as value using java streams

sgrverma2493:
Story: https://jira.baeldung.com/browse/BAEL-7825: sort map value list by element field using Java streams


lor6:
retest this please

PR #16583, opened 2024-05-05
JAVA-33817: Fix tests for spring-data-dynamodb

harry9656:


core-java-modules/core-java-os/pom.xml
```diff
@@ -67,6 +67,7 @@
         <collections-generic.version>4.01</collections-generic.version>
         <unix4j.version>0.4</unix4j.version>
         <grep4j.version>1.8.7</grep4j.version>
+        <mockito.version>4.4.0</mockito.version>
```
alvin577austria, line 70: I guess the module was failing if the mockito version is beyond this? 5.11.0 is the latest.

alvin577austria (changes requested):

core-java-modules/core-java-os/pom.xml
harry9656: yes the module fails if the version is the latest one.

alvin577austria (approved):

PR #16609, opened 2024-05-08
Update README.md

lor6:


rishiraj88:
@lor6 , thanks for the update. However, it is not clear to me why attempting to instantiate the default constructor results in a compilation error for:
`@Test 
void whenUsingRequiredArgsConstructorWithStaticName_thenHideTheConstructor() { 
    SpecialPerson person = SpecialPerson.construct("value1", "value2"); 
    assertNotNull(person); 
}`

Could we have more detailing over this point (article 6) please?

PR #16617, opened 2024-05-09
https://jira.baeldung.com/browse/BAEL-7696

Michaelin007:


spring-reactive-modules/spring-reactive-client-2/src/main/java/com/baeldung/fluxmapandondonext/DoOnNext.java
```diff
@@ -0,0 +1,10 @@
+package com.baeldung.fluxmapandondonext;
+
+import reactor.core.publisher.Flux;
+
+public class DoOnNext {
+
+    public Flux<Integer> number_service_range() {
+        return Flux.just(10, 11, 12, 13, 14);
+    }
+}
```
ashleyfrieze, line 10: Do we need this and `FluxMap.java`?

spring-reactive-modules/spring-reactive-client-2/src/test/java/com/baeldung/fluxmapandondonext/DoOnNextUnitTest.java
```diff
@@ -0,0 +1,51 @@
...
+                LOGGER.info(String.valueOf(number));
+                counter.incrementAndGet();
+            })
+            .onErrorResume(Flux::error);
+
+        StepVerifier.create(numberFlux)
+            .expectNextCount(5)
+            .verifyComplete();
+
+        Assertions.assertEquals(counter.get(), 5);
```
ashleyfrieze, line 31: Let's `static import` the `assertEquals` function here. Also, note that the expected is the first parameter in JUnit assertions, not the second as you have it here.

spring-reactive-modules/spring-reactive-client-2/src/main/java/com/baeldung/fluxmapandondonext/DoOnNext.java
Michaelin007: Hi @ashleyfrieze, both DoOnNext.java and FluxMap.java removed.

spring-reactive-modules/spring-reactive-client-2/src/test/java/com/baeldung/fluxmapandondonext/DoOnNextUnitTest.java
Michaelin007: Done.
Thanks a lot.

PR #16698, opened 2024-05-21
BAEL-7998: RxJava Observable concat() vs merge()

tapankavasthi:


rxjava-modules/rxjava-observables/src/test/java/com/baeldung/rxjava/combine/ObservableMergeVsConcatUnitTest.java
```diff
@@ -0,0 +1,129 @@
...
+import org.junit.Test;
+
+import rx.Observable;
+import rx.observers.TestSubscriber;
+
+public class ObservableMergeVsConcatUnitTest {
+
+    @Test
+    public void givenTwoSynchronousObservables_whenConcatenated_thenAllValuesEmittedInOrder() {
+
```
theangrydev, line 18: Omit the newline at the start of all of the test methods
```suggestion
```

rxjava-modules/rxjava-observables/src/test/java/com/baeldung/rxjava/combine/ObservableMergeVsConcatUnitTest.java
```diff
@@ -0,0 +1,129 @@
...
+        TestSubscriber<Integer> testSubscriberForMerge = new TestSubscriber<>();
+
+        Observable.merge(observable1, observable2)
+            .subscribe(testSubscriberForMerge);
+
+        testSubscriberForMerge.awaitTerminalEvent();
+
+        testSubscriberForMerge.assertValues(4, 5, 6, 1, 2, 3);
+    }
```
theangrydev, line 84: Is this a deterministic ordering?

Interleaved would be more like 1, 4, 2, 5, 3, 6

This just seems to have put the observable2 values before the observable1 ones

Do you understand why the order is 4, 5, 6, 1, 2, 3?

Ah I think I got it, observable1 is 100ms and observable2 is 10ms interval. Hmm okay, I think this might require a bit more explanation in the article to avoid confusion, because it doesn't actually demonstrate an interleaving.

Thoughts?

rxjava-modules/rxjava-observables/src/test/java/com/baeldung/rxjava/combine/ObservableMergeVsConcatUnitTest.java
```diff
@@ -0,0 +1,129 @@
...
+        Observable.concat(observable1, observable2)
+            .subscribe(testSubscriberForConcat);
+
+        testSubscriberForConcat.awaitTerminalEvent();
+
+        testSubscriberForConcat.assertValues(1, 2, 3, 4, 5, 6);
+    }
+
+    @Test
+    public void givenTwoAsynchronousObservablesWithSimilarDelays_whenMerged_thenAllValuesEmittedRegardlessOfOrder() {
```
theangrydev, line 108: I see, you demonstrate an actual mixed interleaving here

theangrydev (changes requested):

rxjava-modules/rxjava-observables/src/test/java/com/baeldung/rxjava/combine/ObservableMergeVsConcatUnitTest.java
tapankavasthi: Fixed

rxjava-modules/rxjava-observables/src/test/java/com/baeldung/rxjava/combine/ObservableMergeVsConcatUnitTest.java
tapankavasthi: As per your suggestion, we have now changed the number of emissions to depict a clear interleaving. I think it explains the concept more clearly now. Thanks!

rxjava-modules/rxjava-observables/src/test/java/com/baeldung/rxjava/combine/ObservableMergeVsConcatUnitTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import rx.Observable;
+import rx.observers.TestSubscriber;
+
+public class ObservableMergeVsConcatUnitTest {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ObservableMergeVsConcatUnitTest.class);
```
theangrydev, line 19: According to https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names this should be lowercase since is is not actually deeply immutable
```suggestion
    private static final Logger logger = LoggerFactory.getLogger(ObservableMergeVsConcatUnitTest.class);
```

theangrydev (changes requested):

PR #16712, opened 2024-05-24
[JAVA-34980] Enable-JDK-12-13-and-above-based-projects

vunamtien:


core-java-modules/core-java-14/src/test/java/com/baeldung/java14/foreign/api/ForeignMemoryUnitTest.java
```diff
@@ -1,84 +1,78 @@
...
+            buffer.putInt(4, 20);
+            buffer.putInt(8, 30);
+
+            assertEquals(10, buffer.getInt(0));
+            assertEquals(20, buffer.getInt(4));
+            assertEquals(30, buffer.getInt(8));
         }
     }
 
     @Test
```
anastasiosioannidis, line 42: Hi @vunamtien 

I think we haven't replicated the original purpose and parameters of the old test method.
For instance, with ByteBuffer we allocate memory on the heap, while with MemorySegment we allocate native memory.
Moreover, in original test, we allocate memory for a sequence of 25 values, while in new test, I don't see that.

Could you give your feedback on this?

core-java-modules/core-java-14/pom.xml
```diff
@@ -20,6 +20,12 @@
             <version>1.18.24</version>
             <scope>provided</scope>
         </dependency>
```
anastasiosioannidis, line 22: Hi @vunamtien 
Please remove this junit dependency

core-java-modules/core-java-14/pom.xml
```diff
@@ -28,25 +34,26 @@
...
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <release>${maven.compiler.release}</release>
-                    <compilerArgs>--enable-preview</compilerArgs>
-                    <source>14</source>
-                    <target>14</target>
+                    <compilerArgs>
+                        <arg>--add-modules</arg>
+                        <arg>jdk.incubator.foreign</arg>
+                    </compilerArgs>
```
anastasiosioannidis, line 40: @vunamtien 
source/target aren't needed, since we build this module with default-jdk9 profiles which contain source/target 17 already. You can verify with the mvn command plus "-X"

PR #16714, opened 2024-05-24
[BAEL-7993] Count the number of occurrences in an array in Java

lucaCambi77:
count occurrences in array and tests

### DIFF
--- a/algorithms-modules/algorithms-miscellaneous-9/src/test/java/com/baeldung/algorithms/permutation/StringPermutationUnitTest.java
+++ b/algorithms-modules/algorithms-miscellaneous-9/src/test/java/com/baeldung/algorithms/permutation/StringPermutationUnitTest.java
@@ -1,7 +1,5 @@
 package com.baeldung.algorithms.permutation;
 
-import org.junit.jupiter.api.Test;
-
 import static com.baeldung.algorithms.permutation.StringPermutation.isPermutationInclusion;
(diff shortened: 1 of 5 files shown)


algorithms-modules/algorithms-miscellaneous-9/src/test/java/com/baeldung/algorithms/permutation/StringPermutationUnitTest.java
```diff
@@ -10,6 +8,8 @@
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import org.junit.jupiter.api.Test;
+
```
ashleyfrieze, line 12: Do we need this?

ashleyfrieze (changes requested):

lor6:
retest this please

algorithms-modules/algorithms-miscellaneous-9/src/test/java/com/baeldung/algorithms/permutation/StringPermutationUnitTest.java
ashleyfrieze: > this is a minor thing we can do at any time

Including never. It just adds noise to this PR. However, leave it be.

I appreciate the "boy scout rule" here, cleaning up the code. Perhaps this isn't the project to do that in. But it's not worth taking time to revert it.

PR #16715, opened 2024-05-24
 BAEL-7984 - Embedded PostgreSQL for Spring Boot Tests

sIvanovKonstantyn:


spring-boot-modules/spring-boot-testing-3/src/test/resources/db/migrations/V1_0_0__create_schema.sql
```diff
@@ -0,0 +1,8 @@
+CREATE SEQUENCE IF NOT EXISTS person_seq INCREMENT 50;
+;
```
Maiklins, line 2: Do we need this additional semicolon?

sIvanovKonstantyn: No. Fixed.

PR #16758, opened 2024-05-30
[JAVA-28864] Renamed profiles + clean up parent pom.xml and README.md

panos-kakos:


README.md
```diff
@@ -36,29 +36,25 @@ Profile based segregation
...
-| default-second             | Second set of projects      | *UnitTest            |
-| integration-lite-second    | Second set of projects      | *IntegrationTest     |
-| default-heavy              | Heavy/long running projects | *UnitTest            |
-| integration-heavy          | Heavy/long running projects | *IntegrationTest     |
-| default-jdk9-and-above     | JDK9 and above projects     | *UnitTest            |
-| integration-jdk9-and-above | JDK9 and above projects     | *IntegrationTest     |
-| parents                    | Set of parent modules       | None                 |
+| Profile           | Includes                    | Type of test enabled |
+|-------------------|-----------------------------| -------------------- |
+| default           | JDK9 and above projects     | *UnitTest            |
```
dkapil, line 51: Let's reword this to JDK17 and above projects

pom.xml
```diff
@@ -1127,12 +999,6 @@
                 <module>libraries-data-io-2</module>
             </modules>
 
-            <properties>
```
dkapil, line 1130: Let's move these tags inside profiles, that is how we will restrict the Java version profile-wise.
This java.version should not be used anywhere inside in internal pom.xml except for the root pom.xml

dkapil (changes requested):

README.md
panos-kakos: ok

pom.xml
panos-kakos: ok, reverted the tags changes.

PR #16802, opened 2024-06-07
BAEL-7934 : improve https://www.baeldung.com/spring-cloud-gateway-bff-oauth2

ch4mpy:
Add Logout failure handlers to UI projects.

In the case of a misconfigured BFF (for instance a bad OAuth2 client secret), all 3 frontend projects were completely silent.

This PR:
- specifies a specific route for the frontend when the authorization code flow fails on the backend
- adds components to display the `error` provided by the backend
- changes the default OAuth2 client secret to the value set by the `build.sh`

It also adds:
- some sample unit & integration testing for the Spring services (BFF & resource server)
- configuration for Back-Channel Logout (Keycloak and BFF)


lor6:
retest this please

ch4mpy:
@lor6 tests have finally passed.

PR #16816, opened 2024-06-10
Creating a deep vs shallow copy of an object in Java

thisara:
This PR contains the supporting repositories for the following eval article.

https://drafts.baeldung.com/?p=202858&preview=true

And there are two projects available;

1. object_copy - This has a spring-boot application which is specifically used for examples in the article.
2. deep_vs_shallow_copy - This is a standard Java project will Unit tests, which is easy to understand.


asif-anwar:
retest this please

PR #16860, opened 2024-06-16
BAEL-7166 Code for the Pass JVM arguments through Maven article

thibaultfaure:


maven-modules/maven-jvm-args/src/test/java/com/baeldung/jvmargs/MyControllerIntegrationTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+@SpringBootTest(classes = MyApplication.class)
+@AutoConfigureMockMvc
+class MyControllerIntegrationTest {
+
+    @Autowired
+    private MockMvc mockMvc;
+
+    @Test
+    void whenGetLength_ThenZero() throws Exception {
```
vinipx, line 19: then - should be lower case

PR #16871, opened 2024-06-18
Use properly reactive endpoint

franz1981:
https://github.com/quarkusio/quarkus-quickstarts/commit/dadc5bc3ed6ec8cf58453501547affcb835f083d 

by default Quarkus run "non async types"  e.g. String in the blocking thread pool, behaving differently from its reactive ones .e.g. Uni<String> .

Adding this annotation is needed to making it behave correctly.


lor6:
Thanks @franz1981 We have a larger task to improve the related article. I'll add this note to our internal task and close here.

franz1981:
I am aware; I am one (the performance Lead of Quarkus) of the Red Hat folks which shared a doc to help on it ;)

PR #16951, opened 2024-06-30
BAEL-8104: junit5 @AutoClosable

etrandafir93:


testing-modules/junit5-annotations/src/test/java/com/baeldung/junit5/autoclosable/AutoClosableExtensionUnitTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.junit.jupiter.api.AutoClose;
+import org.junit.jupiter.api.Test;
+
+class AutoClosableExtensionUnitTest {
+
+    @AutoClose
+    DummyAutoClosableResource resource = new DummyAutoClosableResource();
+
+    @AutoClose("clear")
```
vinipx, line 13: shouldn't we add a test for the DummyClearableResource too?

etrandafir93: @vinipx The test itself doesn't really matter.. in fact, in the article, I haven't even shown it. 

On the other hand, we can add a test so that it looks a bit more realistic.. because right now it the field is not used anywhere. I'll add it now 👍

vinipx: Correct, it is more for a consistency perspective. Thanks!

PR #17076, opened 2024-07-17
JAVA-36106: Upgrade jackson.version

harry9656:


apache-kafka/pom.xml
```diff
@@ -176,7 +177,7 @@
     </build>
 
     <properties>
-        <jackson.version>2.13.4</jackson.version>
+        <jackson.version>2.15.4</jackson.version> <!-- Upgrading to version >= 2.16 breaks tests -->
```
alvin577austria, line 180: I guess the version 2.17 also fails here?

harry9656: Yes, all versions after the 2.16 broke a test and I couldn't figure out what the reason was. 2.15.4 is the latest patched version (Released Feb 15, 2024)

alvin577austria (approved):

PR #17111, opened 2024-07-20
JAVA-37247:- Check inconsistent test apache-kafka/.../BackupCreatorUnitTest.java

amit2103:
tested all tests working. Changed to CopyOnWriteArrayList to avoid contention and make tests faster.


alvin577austria (changes requested):
The fix looks OK but I guess the sink.backups can be accessed via static way.

alvin577austria (approved):

alvin577austria (approved):

PR #17115, opened 2024-07-21
bael-8311: implementation

imsurajmishra:


core-java-modules/core-java-networking-2/src/main/java/com/baeldung/mail/EmailService.java
```diff
@@ -42,29 +42,41 @@ public EmailService(String host, int port) {
```
imsurajmishra, line 42: this PR contains major implementation along with some reformatting introduced due to running intellij formating on the whole file.

for implementation, I have extracted some methods so that its easy to use for across multiple methods.

PR #17203, opened 2024-07-31
BAEL-5865 - Getting Request Payload from POST Request in Java servlet

rajnikantpandey50:


web-modules/javax-servlets-2/src/main/java/com/baeldung/servlets/FileUploadServlet.java
```diff
@@ -0,0 +1,53 @@
...
+        String uploadPath = getServletContext().getRealPath("") + File.separator + UPLOAD_DIRECTORY;
+        File uploadDir = new File(uploadPath);
+        if (!uploadDir.exists()) {
+            uploadDir.mkdir();
+        }
+
+        Part filePart = request.getPart("file");
+        if (filePart != null) {
+            String fileName = filePart.getSubmittedFileName();
+            String filePath = uploadPath + File.separator + fileName;
```
jzheaux, line 34: I'm concerned about directory traversal here. We should either link to OWASP for more guidance or we should use more hardened code here.

web-modules/javax-servlets-2/src/main/java/com/baeldung/servlets/FormDataServlet.java
```diff
@@ -0,0 +1,24 @@
...
+public class FormDataServlet extends HttpServlet {
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+        String firstName = req.getParameter("first_name");
+        String lastName = req.getParameter("last_name");
+
+        resp.getWriter()
+            .append("Full Name: ")
+            .append(firstName)
```
jzheaux, line 20: This is, of course, vulnerable to XSS. So please again either point to the OWASP reference or use something more robust.

web-modules/javax-servlets-2/src/main/java/com/baeldung/servlets/OrderServlet.java
```diff
@@ -0,0 +1,45 @@
...
+                payload.append(line);
+            }
+        } catch (IOException ex) {
+            req.setAttribute("message", "There was an error: " + ex.getMessage());
+        }
+
+        XStream xStream = new XStream();
+        xStream.allowTypesByWildcard(new String[] { "com.baeldung.**" });
+        xStream.alias("Order", Order.class);
+        Order order = (Order) xStream.fromXML(payload.toString());
```
jzheaux, line 37: I'm wondering why read the whole thing into memory when you can stream it into `XStream#fromXML(Reader)`. I agree the above works, I think we just want the code to be as simple as possible so you can stay focused on the concept.

web-modules/javax-servlets-2/src/main/java/com/baeldung/servlets/ProductServlet.java
```diff
@@ -0,0 +1,41 @@
...
+            String line;
+            while ((line = reader.readLine()) != null){
+                payload.append(line);
+            }
+        } catch (IOException ex) {
+            req.setAttribute("message", "There was an error: " + ex.getMessage());
+        }
+
+        Gson gson = new Gson();
+        Product newProduct = gson.fromJson(payload.toString(), Product.class);
```
jzheaux, line 35: Same question here about `fromJson` and using the reader directly.

jzheaux (commented):
Thanks, @rajnikantpandey50! I've left some feedback inline.

web-modules/javax-servlets-2/src/main/java/com/baeldung/servlets/FormDataServlet.java
jzheaux: I think it would be helpful to mark it here in the code as well. Something like:

```
// Don't use this code as-is! It is vulnerable to XSS. 
// See 
// https://www.baeldung.com/cs/cross-site-scripting-xss-explained and
// https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html
// to learn how to harden it.
```

web-modules/javax-servlets-2/src/main/java/com/baeldung/servlets/FileUploadServlet.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        Part filePart = request.getPart("file");
+        if (filePart != null) {
+            String fileName = Paths.get(filePart.getSubmittedFileName()).getFileName().toString();
+            if(fileName.isEmpty()){
+                response.getWriter().println("Invalid File Name!");
+                return;
+            }
+            if(!fileName.endsWith(".txt")){
```
jzheaux, line 41: This is vulnerable to forging a file extension. It's quite simple to name an executable as "malicious.exe.txt" or even "malicious.txt" and then find a vector to execute it on the server.

I recommend linking to https://cheatsheetseries.owasp.org/cheatsheets/File_Upload_Cheat_Sheet.html

web-modules/javax-servlets-2/src/main/java/com/baeldung/servlets/FileUploadServlet.java
```diff
@@ -0,0 +1,64 @@
...
+                response.getWriter().println("Only .txt files are allowed!");
+                return;
+            }
+
+            File file = new File(uploadPath, fileName);
+
+            try (InputStream fileContent = filePart.getInputStream()) {
+                Files.copy(fileContent, file.toPath(), StandardCopyOption.REPLACE_EXISTING);
```
jzheaux, line 49: This is vulnerable to directory traversal. I recommend linking to https://cheatsheetseries.owasp.org/cheatsheets/File_Upload_Cheat_Sheet.html

PR #17204, opened 2024-07-31
[all-eq-in-array] check all elements are equal in array

sk1418:


core-java-modules/core-java-arrays-operations-basic-2/src/test/java/com/baeldung/array/allelementsareequal/AllElementsAreEqualInArrayUnitTest.java
```diff
@@ -0,0 +1,144 @@
...
+        assertTrue(isAllEqual(INT_ARRAY_SINGLE_EL));
+
+        assertFalse(isAllEqual(INT_ARRAY_NOT_EQ));
+    }
+
+    public boolean isAllEqualByDistinct(int[] array) {
+        if (array == null || array.length == 0) {
+            return false;
+        }
+        return Arrays.stream(array)
```
theangrydev, line 116: We don't need to do the array.length == 0 check, it's already handled by Arrays.stream (same elsewhere)
```suggestion
        if (array == null) {
            return false;
        }
        return Arrays.stream(array)
```

core-java-modules/core-java-arrays-operations-basic-2/src/test/java/com/baeldung/array/allelementsareequal/AllElementsAreEqualInArrayUnitTest.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    // primitive array approaches below (int[] as example)
+
+    public boolean isAllEqual(int[] array) {
+        if (array == null || array.length == 0) {
+            return false;
+        }
+        for (int i = 1; i < array.length; i++) {
+            if (!Objects.equals(array[0], array[i])) {
```
theangrydev, line 97: Avoid boxing, make use of the fact it is a non-null primitive:
```suggestion
            if (array[0] != array[i]) {
```

theangrydev (changes requested):

core-java-modules/core-java-arrays-operations-basic-2/src/test/java/com/baeldung/array/allelementsareequal/AllElementsAreEqualInArrayUnitTest.java
sk1418: That's true. But I would leave the length == 0 check as a pre-check. If it's empty, we can immediately return the result without creating unnecessary objects, such as Stream, ArraySpliterator, etc. what do you think?

theangrydev (approved):

PR #17206, opened 2024-08-01
BAEL-8556 add mapper and UT

alexyang0623:


mapstruct-2/src/main/java/com/baeldung/mapstruct/date/mapper/DateMapper.java
```diff
@@ -0,0 +1,18 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public class DateMapper {
+
+    public Date mapStringToDate(String date) {
+        try {
+            SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd");
+            return dateFormat.parse(date);
+        } catch (ParseException e) {
```
theangrydev, line 13: Is mapstruct unable to cope with checked exceptions? If it can cope with them, I'd prefer to see the checked exception declared than wrapped in a RuntimeException

As in: `public Date mapStringToDate(String date) throws ParseException {`

mapstruct-2/src/main/java/com/baeldung/mapstruct/date/mapper/UserMapper.java
```diff
@@ -0,0 +1,29 @@
...
+import com.baeldung.mapstruct.date.model.UserDTO;
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+@Mapper
+public interface UserMapper {
```
theangrydev, line 13: This is not tested

mapstruct-2/src/test/java/com/baeldung/mapstruct/date/mapper/DateMapperTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.junit.jupiter.api.Test;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class DateMapperTest {
```
theangrydev, line 11: This test is not using mapstruct at all

mapstruct-2/src/main/java/com/baeldung/mapstruct/date/mapper/UserMapper.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @Mapping(source = "birthDate", target = "birthDate", dateFormat = "yyyy-MM-dd")
+    User toUser(UserDTO userDTO);
+
+    @Mapping(source = "birthDate", target = "birthDate")
+    User toUserCustom(UserDTO userDTO);
+
+    default Date mapStringToDate(String date) {
```
theangrydev, line 21: Similar comment about rethrowing ParseException

theangrydev (changes requested):

mapstruct-2/src/main/java/com/baeldung/mapstruct/date/model/UserDTO.java
```diff
@@ -0,0 +1,24 @@
+package com.baeldung.mapstruct.date.model;
+
+public class UserDTO {
```
theangrydev, line 3: We follow https://google.github.io/styleguide/javaguide.html
```suggestion
public class UserDto {
```

(same elsewhere, DTO -> Dto)

theangrydev (changes requested):

mapstruct-2/src/main/java/com/baeldung/mapstruct/date/mapper/ReuseUserMapper.java
```diff
@@ -0,0 +1,11 @@
+package com.baeldung.mapstruct.date.mapper;
+
+import com.baeldung.mapstruct.date.model.User;
+import com.baeldung.mapstruct.date.model.UserDTO;
+import org.mapstruct.Mapper;
+
+@Mapper(uses = DateMapper.class)
+public interface ReuseUserMapper {
+
+    User toUser(UserDTO userDTO);
```
theangrydev, line 10: Why is there no `@Mapping` here? How does mapstruct know which field to map, is it just by field name matching exactly? I think we should make this explicit to avoid confusion for the readers

alexyang0623: This is for section 5 of the article to demonstrate the idea of resusing general conversion method. Mapstruct convert field by field name matching, if no extra attribute need to be set, mapping annotation is not required.

mapstruct-2/src/test/java/com/baeldung/mapstruct/date/mapper/DateMapperTest.java
alexyang0623: This is for testing the general conversion method which is designed to be utilized by ReuseUserMapper.

mapstruct-2/src/main/java/com/baeldung/mapstruct/date/model/UserDTO.java
alexyang0623: Will fix later

mapstruct-2/src/main/java/com/baeldung/mapstruct/date/mapper/UserMapper.java
alexyang0623: Wrote the UT but failed to add it in the commit. Will fix it later

mapstruct-2/src/main/java/com/baeldung/mapstruct/date/mapper/ReuseUserMapper.java
theangrydev: I think we should stay on topic and only vary the parts that are relevant to this specific article. I don't want the reader to have to understand this additional concept that isn't part of the intended topic scope. So let's use `@Mapping` consistently across the entire article please.

mapstruct-2/src/test/java/com/baeldung/mapstruct/date/mapper/DateMapperTest.java
theangrydev: I'm not sure this is a valuable test for this article scope, I'd prefer we keep the tests from the perspective of a developer using mapstruct

mapstruct-2/src/main/java/com/baeldung/mapstruct/date/mapper/ReuseUserMapper.java
alexyang0623: @Mapping is one of the ways to convert string to date. Using @Mapper is another way to do so. I shared the outline and you confirmed its okay.

theangrydev: Either way, we should make sure the field mapping is explicit. Is this possible with `@Mapper`? I can't let this pass until we have a way to explicitly specify the field name

theangrydev: My understanding is that `@Mapper(uses = DateMapper.class)` is like an import for another mapper. Perhaps this means `DateMapper` should have `@Mapping`?

There must be a way to specify field names with this approach and it can be a very critical thing - we don't always have class field names that match the names of the fields we want to match 1:1

theangrydev: Perhaps it's better to use an example where the field name doesn't match so we can clarify how that is mapped in both of the use cases you are covering

theangrydev: `public Date mapStringToDate(String date) ` gives us information on how to map from a String to a Date, like a library method. It doesn't tell us how to map the fields.

We should be using both `@Mapping` and `@Mapper` here, they cover totally separate concerns

theangrydev (approved):

theangrydev (approved):

PR #17236, opened 2024-08-05
ON CONFLICT Clause for Insert Queries in Hibernate

mee-omi:


persistence-modules/hibernate-queries-2/src/test/java/com/baeldung/hibernate/HibernateOnConflictUnitTest.java
```diff
@@ -0,0 +1,113 @@
...
+        long rowCountBefore = getRowCount();
+        int updated = session.createMutationQuery("""
+            insert into Student (studentId, name) values (2, 'Sean') 
+            on conflict(studentId) do update 
+            set name = excluded.name
+            """).executeUpdate();
+        long rowCountAfter = getRowCount();
+
+        assertEquals(updated, 1);
+        assertEquals(rowCountBefore, rowCountAfter);
```
KevinGilmore, line 59: The argument order is reversed on all of the assertEquals calls. The expected value should be the first argument, and the actual value should be the second argument. Note that by using AssertJ rather than plain vanilla JUnit assertions, we reduce our chance of making this error.

PR #17283, opened 2024-08-11
[BAEL-8270] How to Print an Array content in Java

Anees1214:


core-java-modules/core-java-arrays-guides/src/test/java/com/baeldung/printarray/PrintArrayTest.java
```diff
@@ -0,0 +1,63 @@
+package com.baeldung.printarray;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+public class PrintArrayTest {
```
JavierPeris, line 6: Using JUnit 5 we don't need public modifier.

core-java-modules/core-java-arrays-guides/src/test/java/com/baeldung/printarray/PrintArrayTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Test
+    public void testPrintArrayUsingForLoop() {
```
JavierPeris, line 9: Same for methods, using JUnit 5 we don't need `public` modifier.

JavierPeris (changes requested):
Please format the files

core-java-modules/core-java-arrays-guides/src/main/java/com/baeldung/printarrays/PrintArrayJava.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Arrays;
+
+public class PrintArrayJava {
+
+    // Print array content using a for loop
+    public String printArrayUsingForLoop(String[] empArray) {
+        StringBuilder result = new StringBuilder();
+        for (int i = 0; i < empArray.length; i++) {
+            result.append(empArray[i]).append(" ");
+        }
```
JavierPeris, line 11: Please use enhanced for loops.

```suggestion
        for (String s : empArray) {
            result.append(s)
                .append(" ");
        }
```

core-java-modules/core-java-arrays-guides/src/main/java/com/baeldung/printarrays/PrintArrayJava.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    // Print 2D array content using nested loops
+    public String print2DArrayUsingNestedLoops(int[][] exampleArr) {
+        StringBuilder result = new StringBuilder();
+        for (int i = 0; i < exampleArr.length; i++) {
+            for (int j = 0; j < exampleArr[i].length; j++) {
+                result.append(exampleArr[i][j]).append(" ");
+            }
+            result.append("\n");
```
JavierPeris, line 48: Same here, enhanced for loops.
```suggestion
        for (int[] ints : exampleArr) {
            for (int anInt : ints) {
                result.append(anInt)
                    .append(" ");
            }
            result.append("\n");
        }
```

core-java-modules/core-java-arrays-guides/src/main/java/com/baeldung/printarrays/PrintArrayJava.java
Anees1214: I've already explained the enhanced for loop in the second method, which is why I used the traditional for loop here to print the array. If I switch to the enhanced for loop, both methods would be the same.

core-java-modules/core-java-arrays-guides/src/main/java/com/baeldung/printarrays/PrintArrayJava.java
Anees1214: As you mentioned, there's no need to discuss 2D arrays since a dedicated guide has already been published on this topic, so I removed that code from my article.

PR #17467, opened 2024-08-27
JAVA-37707 :- Upgrade maven-modules to JDK21.

amit2103:


maven-modules/pom.xml
```diff
@@ -90,5 +90,6 @@
 
     <properties>
         <maven-resources-plugin.version>3.1.0</maven-resources-plugin.version>
```
anastasiosioannidis, line 92: lets remove this line

amit2103: Done.

PR #17468, opened 2024-08-27
codebase/a-basic-aws-lambda-example-with-java [BAEL-8396] [Improvement]

hardikSinghBehl:
[Improvement] Codebase for the article: **A Basic AWS Lambda Example With Java**


aws-modules/aws-lambda-modules/lambda-function/src/main/java/com/baeldung/lambda/LambdaStreamHandler.java
```diff
@@ -0,0 +1,27 @@
...
+public class LambdaStreamHandler implements RequestStreamHandler {
+
+    @Override
+    public void handleRequest(InputStream input, OutputStream output, Context context) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
+        Request request = mapper.readValue(input, Request.class);
+
+        BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(output));
+        writer.write("Hello " + request.name() + ", Baeldung has great Java content for you!");
+        writer.flush();
```
theangrydev, line 22: Can we use a try-with-resources block here, or is there some specific reason not to?

theangrydev (changes requested):
Let's add a LiveTest if possible

aws-modules/aws-lambda-modules/lambda-function/src/main/java/com/baeldung/lambda/LambdaStreamHandler.java
hardikSinghBehl: Yes, we should use try-with-resources here, review incorporated.

theangrydev (approved):

PR #17471, opened 2024-08-28
BAEL-7810: Change lena image with a free portrait from unsplash

harry9656:


image-processing/src/main/java/com/baeldung/imageprocessing/opencv/FaceDetection.java
```diff
@@ -12,36 +13,43 @@
 import org.opencv.objdetect.Objdetect;
 import javafx.scene.image.Image;
 import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.nio.file.Files;
```
JavierPeris, line 17: I'm having some issues with this project so just a question, are we using these new imports?

PR #17495, opened 2024-08-31
fixes for jdk21 profile

panos-kakos:


lor6:
@panos-kakos is this PR still valid?

PR #17512, opened 2024-09-03
[BAEL-8422] Exception handling through spring AOP and AspectJ

rjraunak846:


theangrydev (changes requested):

spring-aop-2/src/test/java/com/baeldung/exceptionhandling/ExceptionHandlingAspectTest.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Test
+    public void givenInvalidPaymentDetails_whenValidationFails_thenAspectLogsException() {
+        // When
+        try {
+            paymentService.validatePaymentDetails("");  // Invalid input
+        } catch (IllegalArgumentException e) {
+            // Exception is expected, so no action needed
+        }
```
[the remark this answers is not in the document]
rjraunak846: Using assertThatThrownBy() now to validate the thrown exception. Thanks.

spring-aop-2/src/main/java/com/baeldung/exceptionhandling/ExceptionHandlingAspect.java
```diff
@@ -0,0 +1,26 @@
...
+
+@Aspect
+@Component
+public class ExceptionHandlingAspect {
+
+    private static final Logger logger = LoggerFactory.getLogger(ExceptionHandlingAspect.class);
+
+    @Pointcut("execution(* com.baeldung.exceptionhandling..*(..))")
+    public void serviceLayerPointcut() {
+        // Pointcut expression
```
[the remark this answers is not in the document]
rjraunak846: I have replaced it with a more clear explanation. Thanks

spring-aop-2/src/test/java/com/baeldung/exceptionhandling/ExceptionHandlingAspectTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Test
+    public void givenInvalidPaymentDetails_whenValidationFails_thenAspectLogsException() {
+        // When
+        assertThatThrownBy(() -> paymentService.validatePaymentDetails("")).isInstanceOf(IllegalArgumentException.class)
+            .hasMessage("Payment details cannot be null or empty.");
+
+        // Then
+        List<ILoggingEvent> logsList = listAppender.list;
+        assertThat(logsList).isNotEmpty();
```
theangrydev, line 44: This is not necessary, the following assertion already implies the list is not empty (same in the other test)

spring-aop-2/src/test/java/com/baeldung/exceptionhandling/ExceptionHandlingAspectTest.java
```diff
@@ -0,0 +1,61 @@
...
+        listAppender = new ListAppender<>();
+        listAppender.start();
+
+        logger.addAppender(listAppender);
+    }
+
+    @Test
+    public void givenInvalidPaymentDetails_whenValidationFails_thenAspectLogsException() {
+        // When
+        assertThatThrownBy(() -> paymentService.validatePaymentDetails("")).isInstanceOf(IllegalArgumentException.class)
```
theangrydev, line 39: Please use the Baeldung formatter, we should be putting chained method calls line by line https://github.com/Baeldung/dev-settings/tree/main/javaCodeStyle/intellij

spring-aop-2/src/test/java/com/baeldung/exceptionhandling/ExceptionHandlingAspectTest.java
```diff
@@ -0,0 +1,61 @@
...
+            .hasMessage("Payment details cannot be null or empty.");
+
+        // Then
+        List<ILoggingEvent> logsList = listAppender.list;
+        assertThat(logsList).isNotEmpty();
+        assertThat(logsList).extracting(Object::toString)
```
theangrydev, line 45: We are still not showing that the Exception is logged:
```
logger.error("Exception occurred: {}", ex.getMessage(), ex);
```

The last parameter `ex` is untested at the moment

theangrydev (changes requested):

spring-aop-2/src/test/java/com/baeldung/exceptionhandling/ExceptionHandlingAspectTest.java
rjraunak846: Removed that. Thanks.

spring-aop-2/src/test/java/com/baeldung/exceptionhandling/ExceptionHandlingAspectTest.java
rjraunak846: I have already configured my IntelliJ with that code style,  but still on formatting the code, it doesn't split all the chained calls to new lines.

spring-aop-2/src/test/java/com/baeldung/exceptionhandling/ExceptionHandlingAspectTest.java
rjraunak846: The last parameter "ex" is not actually needed. I have removed it in the latest commit. 

`assertThat(logsList).extracting(Object::toString)
            .containsExactly("[ERROR] Exception occurred: Payment details cannot be null or empty.");`

This assertion captures the logs and checks whether the exception that actually occurred matches the expected log.

theangrydev: Sorry but I have to disagree here. It's essential when logging exceptions to preserve the stack trace - can we find a way to test this please?

rjraunak846: I agree that preserving the stacktrace is important. But I can't think of a way to create an assertion for that. One way can be to actually assert each line of the stacktrace printed but that will be too clumsy. 

On a side note, isn't it implicit that if the code has reached the error logger and thrown an error message that we have already asserted, then the stack trace is already tested? Let me know if you think otherwise and has a solution to test.

theangrydev: Ok I'll show you how to do this later when I get some time

theangrydev: And no it's not implicit. The test would still be green if I replaced the ex parameter with a dummy new Exception()

rjraunak846: Yup! Got it.

theangrydev: You can do something like:
```
        assertThat(logsList)
            .extracting(event -> event.getThrowableProxy().getMessage())
            .containsExactly("Payment processing failed due to invalid details.");
```

or alternatively, wire a mock/spy Logger into the ExceptionHandlingAspect and use Mockito to verify

PR #17582, opened 2024-09-11
BAEL-7303

BahaaHelmy:
Convert String XML Fragment to Document Node in Java

### DIFF
--- a/xml-3/src/test/java/com/baeldung/xml/XmlDocumentUnitTest.java
+++ b/xml-3/src/test/java/com/baeldung/xml/XmlDocumentUnitTest.java
@@ -23,6 +23,11 @@ public class XmlDocumentUnitTest {
 
         assertNotNull(document);
         assertEquals("root", document.getDocumentElement().getNodeName());
+
+        Element rootElement = document.getDocumentElement();
+        assertNotNull(rootElement.getElementsByTagName("child"));
+        assertEquals(1, rootElement.getElementsByTagName("child").getLength());
+        assertEquals("Example", rootElement.getElementsByTagName("child").item(0).getTextContent());
     }
 
     @Test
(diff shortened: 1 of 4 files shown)


xml-3/src/test/java/com/baeldung/xml/XmlDocumentUnitTest.java
```diff
@@ -0,0 +1,48 @@
...
+    public void givenXmlString_whenConvertToDocument_thenSuccess() throws Exception {
+        String xmlString = "<root><child>Example</child></root>";
+        DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+        DocumentBuilder builder = factory.newDocumentBuilder();
+
+        Document document = builder.parse(new InputSource(new StringReader(xmlString)));
+
+        assertNotNull(document);
+        assertEquals("root", document.getDocumentElement().getNodeName());
+    }
```
dominiqueplante, line 26: Let's add asserts that we can also read the child node

xml-3/src/test/java/com/baeldung/xml/XmlDocumentUnitTest.java
```diff
@@ -0,0 +1,53 @@
...
+        DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+        DocumentBuilder builder = factory.newDocumentBuilder();
+
+        Document document = builder.parse(new InputSource(new StringReader(xmlString)));
+
+        assertNotNull(document);
+        assertEquals("root", document.getDocumentElement().getNodeName());
+
+        Element rootElement = document.getDocumentElement();
+        assertNotNull(rootElement.getElementsByTagName("child"));
```
dominiqueplante, line 28: Since we are using `rootElement.getElementsByTagName("child")` 3 times here, let's extract that into a variable and update each test accordingly to use the variable.

PR #17586, opened 2024-09-12
This PR is related to BAEL-8116

BenHelmyBen:
This PR aims to add a Java class to the module.

### DIFF
--- a/core-java-modules/core-java-8-datetime-3/src/test/java/com/baeldung/datetimecombination/DateTimeCombinationUnitTest.java
+++ b/core-java-modules/core-java-8-datetime-3/src/test/java/com/baeldung/datetimecombination/DateTimeCombinationUnitTest.java
@@ -1,0 +1,38 @@
+package com.baeldung.datetimecombination;
+
+import org.junit.jupiter.api.Test;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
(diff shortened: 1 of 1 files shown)


core-java-modules/core-java-8-datetime-3/src/test/java/com/baeldung/datetimecombination/DateTimeCombinationUnitTest.java
```diff
@@ -0,0 +1,38 @@
...
+    public void givenDateAndTime_whenUsingAtTime_thenCombined() {
+        LocalDateTime combinedDateTime = date.atTime(time);
+
+        assertEquals(expectedDateTime, combinedDateTime);
+    }
+
+    @Test
+    public void givenDateAndTime_whenAdjustingTime_thenCombined() {
+        LocalTime adjustedTime = LocalTime.now().with(ChronoField.MINUTE_OF_HOUR, 0);
+        LocalDateTime expectedAdjustedDateTime = LocalDateTime.of(date, adjustedTime);
```
KevinGilmore, line 34: By using this, you're essentially only showing that the two previous approaches yield the same result. It would be better to simply define expectedAdjustedDateTime explicitly using LocalDateTime.parse.

PR #17599, opened 2024-09-13
[impr-record] first commit

sk1418:


core-java-modules/core-java-14/src/main/java/com/baeldung/java14/recordsCustomConstructors/StudentRecord.java
```diff
@@ -1,12 +1,11 @@
-package org.example;
-
-import java.util.Comparator;
-import java.util.List;
+package com.baeldung.java14.recordsCustomConstructors;
```
theangrydev, line 1: Package names should be lowercase

core-java-modules/core-java-14/src/test/java/com/baeldung/java14/recordsCustomConstructors/StudentRecordJUnitTest.java
```diff
@@ -20,18 +26,46 @@ public void givenStudentRecordData_whenCreated_thenStudentPropertiesMatch() {
...
+        StudentRecordV2 studentV2 = new StudentRecordV2("Jane", 2, 85);
+        assertEquals('B', studentV2.grade());
+    }
+
+    @Test
+    void whenCreateStudentRecordV3_thenCorrect() {
+        StudentRecordV3 studentV3 = new StudentRecordV3("Jane");
+
+        assertFalse(studentV3.id().isEmpty());
+        assertTrue(studentV3.hobbies().isEmpty());
```
theangrydev, line 58: I think we have the guideline to use AssertJ for complex assertions (like checking a collection is not empty) rather than using assertTrue/False with boolean conditions

theangrydev (changes requested):

core-java-modules/core-java-14/src/main/java/com/baeldung/java14/recordsCustomConstructors/StudentRecord.java
sk1418: Oh, thanks for spotting it. I didn't check the package name.. fixed.

core-java-modules/core-java-14/src/test/java/com/baeldung/java14/recordsCustomConstructors/StudentRecordJUnitTest.java
sk1418: I converted them to assertj assertions.

PR #17601, opened 2024-09-14
JAVA-39452 Upgrade apache-httpclient4 from spring5 to spring6

anuragkumawat:


apache-httpclient4/src/test/java/com/baeldung/client/ClientLiveTest.java
```diff
@@ -36,7 +36,7 @@
 class ClientLiveTest extends GetRequestMockServer {
 
 
-    @Test
+    /*@Test
```
alvin577austria, line 39: Did we comment out these tests because they were not needed? If yes please remove.

apache-httpclient4/src/test/java/com/baeldung/client/RestClientV4LiveManualTest.java
```diff
@@ -39,7 +39,7 @@ class RestClientV4LiveManualTest {
 
     final String urlOverHttps = "http://localhost:8082/httpclient-simple/api/bars/1";
 
-    @Test
+    /*@Test
```
alvin577austria, line 42: Same here.

alvin577austria (changes requested):

PR #17630, opened 2024-09-18
[JAVA-39222]

panos-kakos:


spring-batch-2/pom.xml
```diff
@@ -70,9 +70,46 @@
             <artifactId>spring-core</artifactId>
             <version>${spring-core.version}</version>
         </dependency>
```
anastasiosioannidis, line 72: jakarta.xml.bind-api, httpclient and jettison
aren't required in spring-batch pom, let's check.

PR #17636, opened 2024-09-19
Bael 8407 spring ai openai chat client

etrandafir93:


spring-ai/src/main/java/com/baeldung/springai/chatclient/rest/BlogsController.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @GetMapping("v3")
+    List<Article> askForRecommendationFromContext(@RequestParam(name = "question") String question) {
+        return chatClient.prompt()
+            .advisors(new QuestionAnswerAdvisor(vectorStore, SearchRequest.defaults()))
+            .user(question)
+            .call()
+            .entity(new ParameterizedTypeReference<List<Article>>() {
+
```
vinipx, line 73: I don't think we need this empty line here : )

spring-ai/src/main/java/com/baeldung/springai/chatclient/rest/BlogsController.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @GetMapping("v2")
+    List<Article> askForArticleRecommendation(@RequestParam(name = "question") String question) {
+        return chatClient.prompt()
+            .user(question)
+            .call()
+            .entity(new ParameterizedTypeReference<List<Article>>() {
+
```
vinipx, line 62: I don't think we need this empty line here : )

vinipx (commented):
maybe some tests?

spring-ai/src/main/java/com/baeldung/springai/chatclient/rest/BlogsController.java
etrandafir93: I have removed the newlines now, they were added by the formatted 👍

etrandafir93:
> maybe some tests?

@vinipx - I have replied in Jira about the test: it won't be very simple and I don't think they will bring much value here, let's discuss! :)

PR #17657, opened 2024-09-22
JAVA-39729 Fix whenBothFuturesFail_thenGetErrorMessage

timis1:


alvin577austria:
retest this please.

spring-security-modules/spring-security-web-digest-auth/src/main/java/com/baeldung/StartApp.java
```diff
@@ -0,0 +1,13 @@
+package com.baeldung;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
```
alvin577austria, line 4: Maybe you missed some change in pom xml that was not committed?

alvin577austria (changes requested):

spring-security-modules/spring-security-web-digest-auth/src/main/java/com/baeldung/StartApp.java
timis1: I removed that.

PR #17695, opened 2024-09-25
BAEL-8318 Code and Unit Tests for Years Start on Sunday

vBarbaros:
Code and Unit Tests for Years Start on Sunday

- Implements the solution using Date and Calendar, and the new approach using java.time API;
- Implements an optimized example using Spliterator<LocalDate>;
- Provides unit tests for each of the added classes;

Articles: BAEL-8318


vBarbaros:
need to do some cleanup

PR #17728, opened 2024-09-30
codebase/user-agent-parsing-using-yauaa [BAEL-8700]

hardikSinghBehl:
Codebase for the article: **User Agent Parsing Using Yauaa**

### DIFF
--- a/libraries-5/pom.xml
+++ b/libraries-5/pom.xml
@@ -14,6 +14,13 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-maven-plugin</artifactId>
+                <configuration>
+                    <mainClass>com.baeldung.yauaa.Application</mainClass>
+                </configuration>
+            </plugin>
+            <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
@@ -65,13 +72,13 @@
 
         <dependency>
             <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter</artifactId>
-            <version>${springBoot.version}</version>
+            <artifactId>spring-boot-starter-web</artifactId>
+            <version>${spring-boot.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.springframework</groupId>
-            <artifactId>spring-context</artifactId>
-            <version>${spring.version}</version>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-thymeleaf</artifactId>
+            <version>${spring-boot.version}</version>
         </dependency>
 
         <dependency>
@@ -110,15 +117,14 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-test</artifactId>
-            <version>${springBoot.version}</version>
+            <version>${spring-boot.version}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
 
     <properties>
         <jetcache.version>2.7.6</jetcache.version>
-        <springBoot.version>3.3.0</springBoot.version>
-        <spring.version>6.1.8</spring.version>
+        <spring-boot.version>3.1.0</spring-boot.version>
         <manifold.version>2024.1.20</manifold.version>
         <armeria.version>1.29.2</armeria.version>
         <yauaa.version>7.28.1</yauaa.version>
--- a/libraries-5/pom.xml
+++ b/libraries-5/pom.xml
@@ -102,7 +102,11 @@
             <groupId>com.linecorp.armeria</groupId>
             <artifactId>armeria-graphql</artifactId>
         </dependency>
-
+        <dependency>
+            <groupId>nl.basjes.parse.useragent</groupId>
+            <artifactId>yauaa</artifactId>
+            <version>${yauaa.version}</version>
+        </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-test</artifactId>
@@ -117,6 +121,7 @@
         <spring.version>6.1.8</spring.version>
         <manifold.version>2024.1.20</manifold.version>
         <armeria.version>1.29.2</armeria.version>
+        <yauaa.version>7.28.1</yauaa.version>
     </properties>
 
 </project>
--- a/libraries-5/pom.xml
+++ b/libraries-5/pom.xml
@@ -14,13 +14,6 @@
     <build>
         <plugins>
             <plugin>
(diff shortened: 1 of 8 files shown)


libraries-5/pom.xml
```diff
@@ -102,21 +109,25 @@
...
+            <version>${spring-boot.version}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
 
     <properties>
         <jetcache.version>2.7.6</jetcache.version>
-        <springBoot.version>3.3.0</springBoot.version>
-        <spring.version>6.1.8</spring.version>
+        <spring-boot.version>3.1.0</spring-boot.version>
```
theangrydev, line 127: Any reason that we went from 3.3.0 to 3.1.0 here? Prefer latest usually - same for other versions

libraries-5/src/main/java/com/baeldung/yauaa/HomePageController.java
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.web.bind.annotation.RequestHeader;
+import org.springframework.web.servlet.ModelAndView;
+
+import nl.basjes.parse.useragent.UserAgent;
+import nl.basjes.parse.useragent.UserAgentAnalyzer;
+
+@Controller
+public class HomePageController {
+
+    private static final List<String> SUPPORTED_DEVICES = List.of("Mobile", "Tablet", "Phone");
```
theangrydev, line 18:
```suggestion
    private static final List<String> SUPPORTED_MOBILE_DEVICE_CLASSES = List.of("Mobile", "Tablet", "Phone");
```

libraries-5/src/main/java/com/baeldung/yauaa/HomePageController.java
```diff
@@ -0,0 +1,37 @@
...
+    private final UserAgentAnalyzer userAgentAnalyzer;
+
+    public HomePageController(UserAgentAnalyzer userAgentAnalyzer) {
+        this.userAgentAnalyzer = userAgentAnalyzer;
+    }
+
+    @GetMapping("/home")
```
theangrydev, line 25: Since we reject non mobile devices, maybe putting mobile in the URL would make this clearer?

Or happy to be convinced otherwise that this is truly a mobile-only website example, just not something I've seen IRL as far as I remember.

```suggestion
    @GetMapping("/mobile/home")
```

libraries-5/src/main/java/com/baeldung/yauaa/UserAgentAnalyzerConfiguration.java
```diff
@@ -0,0 +1,21 @@
...
+import nl.basjes.parse.useragent.UserAgentAnalyzer;
+
+@Configuration
+public class UserAgentAnalyzerConfiguration {
+
+    @Bean
+    public UserAgentAnalyzer userAgentAnalyzer() {
+        return UserAgentAnalyzer
+            .newBuilder()
+            .withCache(1000)
```
theangrydev, line 16: What does the cache parameter do? Why 1000, any reason to not keep the default?

libraries-5/src/main/java/com/baeldung/yauaa/UserAgentAttributeLoggingFilter.java
```diff
@@ -0,0 +1,45 @@
...
+    public UserAgentAttributeLoggingFilter(UserAgentAnalyzer userAgentAnalyzer) {
+        this.userAgentAnalyzer = userAgentAnalyzer;
+    }
+
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+        throws ServletException, IOException {
+        String userAgentString = request.getHeader("User-Agent");
+
+        if (StringUtils.hasText(userAgentString)) {
```
theangrydev, line 33: Does this mean that .parse can't cope with empty/null strings?

libraries-5/src/main/java/com/baeldung/yauaa/UserAgentAttributeLoggingFilter.java
```diff
@@ -0,0 +1,45 @@
...
+            UserAgent userAgent = userAgentAnalyzer.parse(userAgentString);
+            userAgent
+                .getAvailableFieldNamesSorted()
```
theangrydev, line 36: How does this differ from getCleanedAvailableFieldNamesSorted?

theangrydev (changes requested):

libraries-5/src/main/java/com/baeldung/yauaa/UserAgentAttributeLoggingFilter.java
hardikSinghBehl: The getCleanedAvailableFieldNamesSorted() method does not include the fields where the value is unavailable. In the article, since we show the unavailable value fields, it'll be better to use getAvailableFieldNamesSorted(). But I've added in a new sentence to mention the other method as well.

libraries-5/src/main/java/com/baeldung/yauaa/UserAgentAttributeLoggingFilter.java
hardikSinghBehl: It can, I've removed this check.

libraries-5/src/main/java/com/baeldung/yauaa/UserAgentAnalyzerConfiguration.java
hardikSinghBehl: I used a different value than the default to show the readers that the cache size can be updated as required. However to convey it better and also mention the default size, i've added the following in the article:

`The UserAgentAnalyzer, by default uses an in-memory cache with size of 10000, but we can update it as per requirement using the withCache() method as we've done in our example. Caching helps us improve performance by avoiding repeated parsing of the same user agent strings.`

libraries-5/src/main/java/com/baeldung/yauaa/HomePageController.java
hardikSinghBehl: review incorporated.

libraries-5/src/main/java/com/baeldung/yauaa/HomePageController.java
hardikSinghBehl: review incorporated.

libraries-5/pom.xml
hardikSinghBehl: Fixed.

hardikSinghBehl: I was getting a lot of incompatible version errors due to the jetcache dependencies, spent some time trying to fix it but eventually decided to downgrade the version. 

I suspect this was the same reason the spring context was downgraded by the original committer.

hardikSinghBehl: Fixed.

theangrydev (approved):

PR #17756, opened 2024-10-04
BAEL-8597 read multiple integers from one line

Tirth007:
BAEL-8597 read multiple integers from one line


davidmartinezbarua:
retest this please

PR #17793, opened 2024-10-10
BAEL-7003 - HTTP DELETE With Request Body

ulisseslima:
Draft: https://drafts.baeldung.com/wp-admin/post.php?post=216383&action=edit


JavierPeris (commented):
Nice job @ulisseslima, the final touch is to format the files: there are some trailing whitespaces and in the Body class there's no space between class definition and the first property.

PR #17815, opened 2024-10-12
[filter-list-by-any-field] full-text search

sk1418:


theangrydev:
retest this please

core-java-modules/core-java-collections-list-7/src/test/java/com/baeldung/filterbyanyfield/FilterListByAnyMatchingFieldUnitTest.java
```diff
@@ -0,0 +1,188 @@
...
+                    Object value = field.get(obj);
+                    if (value != null && value.toString()
+                        .contains(keyword)) {
+                        return true;
+                    }
+                    if (!field.getType()
+                        .isPrimitive() && !(value instanceof String)) {
+                        return fullTextSearchOnObject(value, keyword);
+                    }
+                } catch (InaccessibleObjectException | IllegalAccessException e) {
```
theangrydev, line 108: There's a convention to name the variable ignored if it is
```suggestion
                } catch (InaccessibleObjectException | IllegalAccessException ignored) {
```

core-java-modules/core-java-collections-list-7/src/test/java/com/baeldung/filterbyanyfield/FilterListByAnyMatchingFieldUnitTest.java
```diff
@@ -0,0 +1,188 @@
...
+                field.setAccessible(true);
+                try {
+                    Object value = field.get(obj);
+                    if (value != null && value.toString()
+                        .contains(keyword)) {
+                        return true;
+                    }
+                    if (!field.getType()
+                        .isPrimitive() && !(value instanceof String)) {
+                        return fullTextSearchOnObject(value, keyword);
```
theangrydev, line 106: Intentional to not pass on the excludedFields too?

core-java-modules/core-java-collections-list-7/src/test/java/com/baeldung/filterbyanyfield/FilterListByAnyMatchingFieldUnitTest.java
```diff
@@ -0,0 +1,188 @@
...
+                    }
```
theangrydev, line 107: This will exit early with false, I think we need something like this instead?
```suggestion
                    if (!field.getType()
                        .isPrimitive() && !(value instanceof String) && value != null && fullTextSearchOnObject(value, keyword)) {
                        return true;
                    }
```

theangrydev (changes requested):

core-java-modules/core-java-collections-list-7/src/test/java/com/baeldung/filterbyanyfield/FilterListByAnyMatchingFieldUnitTest.java
theangrydev: (not sure what kind of test case would catch this)

core-java-modules/core-java-collections-list-7/src/test/java/com/baeldung/filterbyanyfield/FilterListByAnyMatchingFieldUnitTest.java
sk1418: I am not really sure about this. If user wants to merely exclude the super.someProp or All "someProp"s at any level. 
I think it depends on the requirement. 
I'll pass the excludedFields in the recursion. Readers should get the idea behind this solution.

core-java-modules/core-java-collections-list-7/src/test/java/com/baeldung/filterbyanyfield/FilterListByAnyMatchingFieldUnitTest.java
sk1418: Good catch 👍 . We should only return in case the recursion returns true.
I'd make it two if(...) return true blocks instead of one single if() with "or". I think it's easier to read.

core-java-modules/core-java-collections-list-7/src/test/java/com/baeldung/filterbyanyfield/FilterListByAnyMatchingFieldUnitTest.java
sk1418: renamed

PR #17852, opened 2024-10-19
BAEL-8583: Validations using Yavi

sazzer:


libraries-5/src/test/java/com/baeldung/yavi/ArgumentsAnnotationUnitTest.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        @ConstraintArguments
+        boolean isOlderThan(int check) {
+            return this.age() > check;
+        }
+    }
+
+    @Test
+    void givenAConstructorValidator_whenCalledWithInvalidArguments_nothingIsConstructed() {
```
theangrydev, line 26: We use a given_when_then syntax (same for other tests)

theangrydev (changes requested):

theangrydev (approved):

PR #17853, opened 2024-10-19
BAEL-8487: Fix the JsonMappingException: Can not deserialize instance of java.util.HashMap out of START_ARRAY token

azhwani:
+ Create Person class
+ Create test cases to showcase how to reproduce and fix JsonMappingException


jackson-modules/jackson-exceptions/src/test/java/com/baeldung/mappingexception/JacksonMappingExceptionUnitTest.java
```diff
@@ -89,4 +94,38 @@ public void givenEmptyBean_whenSerializing_thenCorrect() throws JsonProcessingEx
         assertEquals("{}", json);
     }
 
+    @Test
+    public void givenJsonArray_whenDeserializingToMap_thenThrowException() throws JsonProcessingException {
```
JavierPeris, line 98: The IDE says that this exception cannot be thrown. Could we remove it?

PR #17912, opened 2024-10-31
Finding IP Address of a URL in Java (biivincent4@gmail.com)

vincentbii:
Update GetURLIPAddressUnitTest.java - removed `validate()` function from test class. pointed test cases' validations to use the newly created `validate` function in `urlipaddress` class


lor6:
retest this please

PR #17918, opened 2024-11-02
Java 41287 Moving some article links on Github - core-java-collections

anuragkumawat:


core-java-modules/core-java-collections-4/src/main/java/com/baeldung/collections/stackreversal/Employee.java
anastasiosioannidis, line 1: @anuragkumawat 
I see Employee class is in "sorting".
Why is this moved in "stackreversal", is this by mistake?

PR #18008, opened 2024-11-22
Typo correction

ZungTa:
Hello.
I found a typo while looking through the document.
I think it should be "millis".
If I'm wrong, sorry.

And I fixed the spacing a little.


lor6:
Thanks @ZungTa I've merged the PR.

PR #18026, opened 2024-11-27
JAVA-42138 upgrade or disable modules that fail due to restlet error

lor6:


lor6:
New PR: https://github.com/eugenp/tutorials/pull/18035 to add JIRA id's for modules disabled in this PR.

PR #18076, opened 2024-12-09
BAEL-7288 Reactive kafka stream

thiagohora:
BAEL-7288


thiagohora:
Test failing seems to have no relation with my PR

PR #18104, opened 2024-12-19
https://jira.baeldung.com/browse/BAEL-8843

Michaelin007:


spring-boot-modules/spring-boot-3-4/pom.xml
```diff
@@ -0,0 +1,27 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
Maiklins, line 1: As this is a new module, let's add it to the parent pom as well.

Michaelin007: Alright @Maiklins 
Thanks!

PR #18106, opened 2024-12-20
BAEL-7881: Introduction to SootUp

sazzer:


libraries-5/src/test/java/com/baeldung/sootup/AnalyzeUnitTest.java
```diff
@@ -0,0 +1,57 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class AnalyzeUnitTest {
+    @Test
+    void whenAnalyzingTheJVM_thenWeCanListClasses() {
```
theangrydev, line 19: We follow https://google.github.io/styleguide/javaguide.html#s5.3-camel-case
```suggestion
    void whenAnalyzingTheJvm_thenWeCanListClasses() {
```

libraries-5/src/test/java/com/baeldung/sootup/AnalyzeUnitTest.java
```diff
@@ -0,0 +1,57 @@
...
+        AnalysisInputLocation inputLocation = new JrtFileSystemAnalysisInputLocation();
+
+        JavaView view = new JavaView(inputLocation);
+
+        assertTrue(view.getClasses().size() > 0);
```
theangrydev, line 24: For complex assertions we use AssertJ

libraries-5/src/test/java/com/baeldung/sootup/MethodBodyUnitTest.java
```diff
@@ -0,0 +1,84 @@
...
+                ));
+        assertTrue(method.isPresent());
+
+        SootMethod sootMethod = method.get();
+
+        Body methodBody = sootMethod.getBody();
+        StmtGraph<?> stmtGraph = methodBody.getStmtGraph();
+        List<Stmt> stmts = stmtGraph.getStmts();
+        for (Stmt stmt : stmts) {
+            System.out.println(stmt);
```
theangrydev, line 75: Can we assert here instead of printing?

libraries-5/src/test/java/com/baeldung/sootup/MethodBodyUnitTest.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+
+
+    private void someMethod(String name) {
+        var capitald = name.toUpperCase();
+        System.out.println("Hello, " + capitald);
```
theangrydev, line 82: We have a guideline to use SLF4J not System.out for print statements

libraries-5/src/test/java/com/baeldung/sootup/MethodUnitTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+        IdentifierFactory identifierFactory = view.getIdentifierFactory();
+        ClassType javaClass = identifierFactory.getClassType("com.baeldung.sootup.MethodUnitTest");
+
+        SootClass sootClass = view.getClassOrThrow(javaClass);
+        Set<? extends SootMethod> method = sootClass.getMethodsByName("someMethod");
+        assertEquals(1, method.size());
+    }
+
+    private void someMethod(String name) {}
```
theangrydev, line 74: Can we show multiple overloaded methods with the same name listed here perhaps?

theangrydev (changes requested):

libraries-5/src/test/java/com/baeldung/sootup/MethodBodyUnitTest.java
sazzer: This method isn't being used. It's purely here so that it can be analyzed. I just needed to do *something* with the `capitals` variable was all.

libraries-5/src/test/java/com/baeldung/sootup/MethodBodyUnitTest.java
sazzer: We can, but I was concerned that this would be a bit flaky. It's possible that the exact format of the statements - or even the number of them - would change between versions of SootUp, and *maybe* between versions of Java.

However, I've updated it as you said so if you want to keep it then it's there :)

theangrydev (approved):

PR #18115, opened 2024-12-26
BAEL-6483 example code

alexyang0623:


logging-modules/jul-to-slf4j/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jul-to-slf4j</artifactId>
+            <version>2.0.16</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+            <version>2.0.16</version>
```
KevinGilmore, line 31: Let's use a property for the version number - especially since it's repeated.

PR #18182, opened 2025-01-15
JAVA-40698 Moved code for article spring-rest-docs-vs-openapi

Bipinkumar27:


spring-boot-modules/spring-boot-springdoc-2/README.md
anastasiosioannidis, line 1: lets revert swagger article, and add it to swagger springfox module

PR #18230, opened 2025-01-27
BAEL 7697 - How to Distinguish Between Field Absent vs Null in Jackson

ulisseslima:
Draft: https://drafts.baeldung.com/wp-admin/post.php?post=224918&action=edit


jackson-modules/jackson-conversions-3/src/test/java/com/baeldung/jackson/absentvsnull/JacksonAbsentVsNullUnitTest.java
```diff
@@ -0,0 +1,130 @@
...
+import com.fasterxml.jackson.databind.exc.MismatchedInputException;
+
+class JacksonAbsentVsNullUnitTest {
+
+    private static final String ID_FIELD = "id";
+    private static final String PRIMITIVE_FIELD = "amount";
+    private static final ObjectMapper DEFAULT_MAPPER = new ObjectMapper();
+
+    private static final Sample ZERO_ARG_SAMPLE = new Sample();
+    private static final Sample FULL_SAMPLE = new Sample();
```
jzheaux, line 31: Could you instead add a constructor that takes a set of default values? Or a static factory like `Sample.defaults()`? If you do this, then you wouldn't need a constant here and could simply do `Sample.defaults()` inside each test. This would also remove the need for `@AfterEach`.

jackson-modules/jackson-conversions-3/src/test/java/com/baeldung/jackson/absentvsnull/JacksonAbsentVsNullUnitTest.java
```diff
@@ -0,0 +1,130 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.exc.MismatchedInputException;
+
+class JacksonAbsentVsNullUnitTest {
+
+    private static final String ID_FIELD = "id";
+    private static final String PRIMITIVE_FIELD = "amount";
+    private static final ObjectMapper DEFAULT_MAPPER = new ObjectMapper();
+
+    private static final Sample ZERO_ARG_SAMPLE = new Sample();
```
jzheaux, line 30: I think it's a bit misleading to call these constants since their values can change.

I'd recommend you give your learners less to read by doing:

```java
Sample empty = new Sample();
```

Even better, doing `new Sample()` is so simple, I'd strongly recommend simply doing that in each test instead of sharing an instance. This makes it easier to run tests concurrently and it reduces the amount of scrolling a reader needs to do to comprehend a given test.

jackson-modules/jackson-conversions-3/src/test/java/com/baeldung/jackson/absentvsnull/JacksonAbsentVsNullUnitTest.java
```diff
@@ -0,0 +1,130 @@
...
+        FULL_SAMPLE.setName(ID_FIELD);
+        FULL_SAMPLE.setAmount(3);
+        FULL_SAMPLE.setKeys(keys);
+        FULL_SAMPLE.setValues(values);
+    }
+
+    @Test
+    final void whenSerializingWithDefaults_thenNullValuesIncluded() {
+        Sample sample = ZERO_ARG_SAMPLE;
+        Map<String, Serializable> map = DEFAULT_MAPPER.convertValue(sample, SampleUtils.MAP_TYPE);
```
jzheaux, line 53: It's unclear why `SampleUtils` is holding onto `MAP_TYPE` while this class is holding onto other static fields.

jackson-modules/jackson-conversions-3/src/test/java/com/baeldung/jackson/absentvsnull/JacksonAbsentVsNullUnitTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+        assertEquals(ZERO_ARG_SAMPLE.getAmount(), map.get("amount"));
+        assertTrue(map.containsKey(ID_FIELD));
+        assertNull(map.get(ID_FIELD));
+    }
+
+    @Test
+    final void whenDeserializingToMapWithDefaults_thenAbsentFieldsAreIgnored() throws JsonProcessingException {
+        String json = "{\"values\": [2], \"keys\": []}";
+        Map<String, Serializable> map = DEFAULT_MAPPER.readValue(json, SampleUtils.MAP_TYPE);
```
jzheaux, line 63: It's quite confusing to have a `DEFAULT_MAPPER` and a `MAPPER`. The semantics are lost.

jackson-modules/jackson-conversions-3/src/test/java/com/baeldung/jackson/absentvsnull/JacksonAbsentVsNullUnitTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+        assertFalse(map.containsKey(PRIMITIVE_FIELD));
```
jzheaux, line 65: It seems like there is some context missing with these assertions. For example, the `json` has `values` and `keys`, but the test here is about `amount`. This is a bit jarring for the reader because these don't feel related. An assertion like:

```java
Set<String> keySet = map.keySet();
assertTrue(keySet.size() == 2);
assertTrue(keySet.containsAll(List.of("keys", "values")));
```

might make more sense since it remains in the context of what is declared in the test givens.

Another might be:

```java
assertTrue(map.size() == 2);
assertTrue(map.containsKey("keys"));
assertTrue(map.containsKey("values"));
```

jackson-modules/jackson-conversions-3/src/test/java/com/baeldung/jackson/absentvsnull/JacksonAbsentVsNullUnitTest.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    @Test
+    final void whenDeserializingToMapWithDefaults_thenNullPrimitiveIsDefaulted() throws JsonProcessingException {
+        String json = "{\"amount\": null}";
+        Sample sample = DEFAULT_MAPPER.readValue(json, Sample.class);
+
+        assertNotEquals(ZERO_ARG_SAMPLE.getAmount(), sample.getAmount());
```
jzheaux, line 73: I think this could be clearer. Instead of demonstrating what it _isn't_, you can demonstrate what it is (this is what your test name says anyway):

```java
assertEquals(sample.getAmount(), 0);
```

jackson-modules/jackson-conversions-3/src/test/java/com/baeldung/jackson/absentvsnull/JacksonAbsentVsNullUnitTest.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    @Test
+    final void whenValidatingNullPrimitives_thenFailOnNullAmount() {
+        String json = "{\"amount\": null}";
+
+        assertThrows(MismatchedInputException.class, () -> SampleUtils.MAPPER
```
jzheaux, line 80: I think this is an example of a place where it's a bit confusing because the `MAPPER` configurations are being used for multiple things. I think what you are trying to demonstrate is that when Jackson uses `FAIL_ON_NULL_FOR_PRIMITIVES`, then this kind of error happens. However, that's a bit lost that this is the setting that achieves that.

A test that demonstrates this clearer would be:

```java
String json = "{ \"amount\": null }";
ObjectMapper mapper = new ObjectMapper();
mapper.enable(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES);
assertThrows(MismatchedInputException.class, () -> mapper.readValue(json, Sample.class));
```

jackson-modules/jackson-conversions-3/src/main/java/com/baeldung/jackson/absentvsnull/model/Sample.java
```diff
@@ -0,0 +1,57 @@
...
+    public List<Integer> getValues() {
+        return values;
+    }
+
+    public void setValues(List<Integer> values) {
+        this.values = values;
+    }
+
+    @Override
+    public String toString() {
```
jzheaux, line 54: You might consider `Objects.toString(this)`.

jackson-modules/jackson-conversions-3/src/main/java/com/baeldung/jackson/absentvsnull/SampleUtils.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    private SampleUtils() {
+    }
+
+    public static Sample deserialize(String json) throws JsonProcessingException {
+        return MAPPER.readValue(json, Sample.class);
+    }
+
+    public static Map<String, Serializable> deserializeAsMap(String json) throws JsonProcessingException {
```
jzheaux, line 38: I think this method could be removed. There are many places in the test that simply deserialize directly. Otherwise, this is only used in `updateNonAbsent`.

jackson-modules/jackson-conversions-3/src/main/java/com/baeldung/jackson/absentvsnull/SampleUtils.java
```diff
@@ -0,0 +1,100 @@
...
+
+        MAPPER.setSerializationInclusion(Include.NON_NULL);
+        MAPPER.setSerializationInclusion(Include.NON_ABSENT);
+        MAPPER.setSerializationInclusion(Include.NON_DEFAULT);
+    }
+
+    private SampleUtils() {
+    }
+
+    public static Sample deserialize(String json) throws JsonProcessingException {
```
jzheaux, line 34: I think this could be removed. Since you are trying to teach Jackson usage, it's nice to see the actual Jackson in context of the usage instead of having to scroll up to this method (from `updateConsideringNulls`, say) to see what `deserialize` does.

jzheaux (changes requested):
I've left some feedback here as well. Overall, I'd recommend to please keep the `public static` usage to a minimum, especially for values that can change. And see if you can simply these three classes and get them down to one. This will reduce scrolling and allows readers to see you point all as one unit. Please see the rest of my feedback inline.

jzheaux (approved):

PR #18254, opened 2025-02-02
BAEL-8864: Select a Random Key from a HashMap in Java

azhwani:
+ Add test cases to showcase how to select a random key from a Map


asif-anwar:
retest this please

PR #18357, opened 2025-03-03
JAVAP-40683

panos-kakos:


spring-boot-modules/spring-boot-3-2/pom.xml
```diff
@@ -18,22 +18,27 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-web</artifactId>
```
anastasiosioannidis, line 20: all below explicit version setups, aren't needed, correct?

PR #18418, opened 2025-03-18
Have authenticationConverter use the interface AuthoritiesConverter

Psynbiotik:
The article says this:
```
Because of generics type erasure in the JVM and because there could be many Converter beans with different inputs and outputs in an application context, this AuthoritiesConverter interface can be a useful tip for the bean factory when it searches for a Converter<Map<String, Object>, Collection<GrantedAuthority>> bean.
```
Then it never uses the interface for the authenticationConverter.


lor6:
Thanks @Psynbiotik I'll update the article as well.

PR #18438, opened 2025-03-30
BAEL-9212 How to control tag order in SpringDoc OpenAPI

dvohra09:


spring-boot-modules/spring-boot-springdoc/src/main/java/com/baeldung/springdoc/tagorderexample/BooksController.java
```diff
@@ -0,0 +1,73 @@
...
+     @Tag(name = "findBook", description = "Find book from inventory"),
+     @Tag(name = "updateBook", description = "Update book in inventory")
+})**/
+@Tag(name = "books", description = "Books related tag")
+public class BooksController {
+
+    @Tag(name = "create")
+    @Tag(name = "books")
+    @Tag(name = "createBook")
+    @RequestMapping(path = "/book", method = RequestMethod.POST)
```
KevinGilmore, line 40: On the methods, let's use the more recent Spring RestController annotations, @PostMapping, @GetMapping, and @DeleteMapping, instead of the older @RequestMapping.

PR #18440, opened 2025-03-30
Fix report generation for custom Maven plugin

Marcono1234:
Since https://issues.apache.org/jira/browse/MPLUGIN-467 a dedicated plugin has to be used for report generation.
The previous setup here was already failing, but maybe it was not noticed because CI does not run `mvn site` for it?

> [!IMPORTANT]\
> The code snippets on https://www.baeldung.com/maven-plugin#documentation have to be updated as well.

### DIFF
--- a/maven-modules/maven-custom-plugin/counter-maven-plugin/pom.xml
+++ b/maven-modules/maven-custom-plugin/counter-maven-plugin/pom.xml
@@ -55,15 +55,8 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-plugin-plugin</artifactId>
-                <version>${maven-compiler-plugin.version}</version>
-                <reportSets>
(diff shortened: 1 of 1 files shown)


maven-modules/maven-custom-plugin/counter-maven-plugin/pom.xml
```diff
@@ -55,15 +55,8 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-plugin-plugin</artifactId>
-                <version>${maven-compiler-plugin.version}</version>
```
Marcono1234, line 59: Note that there was also previously a copy & paste error here: It erroneously used <code>maven-<b>compiler</b>-plugin.version</code>.

lor6:
Thanks @Marcono1234 I will update the article as well.

PR #18467, opened 2025-04-11
BAEL-8307 Map empty string to null in MapStruct

sam-gardner:
Adds code for the mapping empty string to null in mapstruct article.


eric-martin:
retest this please

PR #18519, opened 2025-04-29
Bael 7230 modify properties

bpstelios10:


core-java-modules/core-java-properties/pom.xml
```diff
@@ -14,4 +14,16 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-configuration2</artifactId>
+            <version>2.11.0</version>
+        </dependency>
+        <dependency>
+            <groupId>commons-beanutils</groupId>
+            <artifactId>commons-beanutils</artifactId>
+            <version>1.10.1</version>
```
KevinGilmore, line 26: Let's use properties for all version numbers as per guidelines

core-java-modules/core-java-properties/src/test/java/com/baeldung/core/java/properties/ApacheCommonsPropertyMutatorUnitTest.java
```diff
@@ -0,0 +1,56 @@
...
+        assertEquals("TestApp", propertyMutator.getProperty("name"));
+
+        propertyMutator.addProperty("new.property", "new-value");
+
+        assertEquals("new-value", propertyMutator.getProperty("new.property"));
+        assertEquals("TestApp", propertyMutator.getProperty("name"));
+    }
+
+    @Test
+    public void updateProperty_whenApacheCommonsIsUsed_thenReturnsUpdatedPropertyWithoutAffectingOtherProperties() throws ConfigurationException {
```
KevinGilmore, line 47: In this file and the other unit test files, let's stick to the [givenX_]whenY_thenZ pattern rather than prefixing the test names. Here are a couple of example naming conventions you might use, given the first test as an example:

- whenAddPropertyUsingApacheCommons_then...
- givenApacheCommons_thenAddProperty_then...

bpstelios10: thanks @KevinGilmore 
 i'm always carried away and use uncle bob's rule of test method name first 😬

PR #18527, opened 2025-05-04
How to Implement Retry for JUnit Tests

3ga01:


testing-modules/junit/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+    <groupId>com.baeldung</groupId>
+    <artifactId>junit</artifactId>
+    <version>1.0-SNAPSHOT</version>
+
+    <parent>
+        <groupId>com.baeldung</groupId>
+        <artifactId>testing-modules</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+<dependencies>
```
KevinGilmore, line 16: This section should be indented.

testing-modules/junit/src/main/java/com/baeldung/Main.java
```diff
@@ -0,0 +1,7 @@
+package com.baeldung;
+
+public class Main {
+	public static void main(String[] args) {
```
KevinGilmore, line 4: Indentation in Java  code should always use spaces (four per level) and not tabs.

PR #18537, opened 2025-05-09
Update pom.xml

jsfan3:
"skip true" tells the plugin to bypass both compilation and execution of its J2CL tests, so the missing test_summary.json error should disappear and the rest of the multi-module build should keep running its normal unit / integration tests.

A project with packaging pom is not required to produce a primary JAR/WAR artefact, so Maven’s package phase won’t expect the maven-war-plugin at all.


jsfan3:
please test again

PR #18573, opened 2025-05-25
BAEL-9317 Parsing JSON boolean Value in Java - improve to talk about 0/1 representation

Deepak-Vohra:


json-modules/json/src/test/java/com/baeldung/jsonjava/ParseJsonBooleanUnitTest.java
```diff
@@ -0,0 +1,62 @@
...
+        assertTrue(isActive);
+    }
+
+    @Test
+    void givenJSONWithMixedRepresentationForBoolean_whenParsed_correctBooleanValueReturned() {
+        String jsonString = "{\"name\":\"lorem ipsum\",\"active\": 0,\"id\":1}";
+        JSONObject jsonObject = new JSONObject(jsonString);
+        Object activeObject = jsonObject.get("active");
+        boolean active;
+        if (activeObject instanceof Integer) {
```
jzheaux, line 36: Please use the `instanceof` for casting instead:

```java
if (activeObject instanceof Integer value) {
    assertFalse(value == 1);
} else if (activeObject instanceof Boolean value) {
    assertFalse(value);
}
```

Also, there appears to be no need for the final assert if done this way.

json-modules/json/src/test/java/com/baeldung/jsonjava/JSONObjectIntegrationTest.java
```diff
@@ -22,7 +23,29 @@ public void givenJSONJava_thenCreateNewJSONObject() {
...
+                                              }
+                            }
+                            """;
+        JSONObject jsonObject = new JSONObject(jsonString);
+        String type = jsonObject.getString("type");
+        String geometry = jsonObject.getString("geometry");
+        JSONObject properties = jsonObject.getJSONObject("properties");
+        boolean isValid = properties.getBoolean("isValid");
+        assertEquals(type,"Feature");
+        assertEquals(geometry,"Point");
```
KevinGilmore, line 43: Please reverse the order of the arguments. It's intended to be assertEquals(expected, actual)

Deepak-Vohra: Reversed

PR #18582, opened 2025-05-28
BAEL-7988: Event-Driven LISTEN/NOTIFY Support in Java using PostgreSQL

sazzer:


persistence-modules/core-java-persistence-4/src/test/java/com/baeldung/listennotify/ListenNotifyLiveTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+import com.impossibl.postgres.api.jdbc.PGNotificationListener;
+import org.junit.jupiter.api.Test;
+import org.postgresql.PGNotification;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.sql.*;
+
+public class ListenNotifyLiveTest {
```
theangrydev, line 11: Can we write these in a way that they actually assert something? Or, explain how they were intended to be run (some tests only notify, some tests only listen, unclear how to see the listeners in action, run multiple tests simultaneously?)

theangrydev (changes requested):

persistence-modules/core-java-persistence-4/src/test/java/com/baeldung/listennotify/ListenNotifyLiveTest.java
sazzer: I've updated the tests to assert that the notifications were really received.

I've done it using a Set and not a List because I'm not sure if there's any guarantee of the order of delivery.

Cheers

theangrydev (approved):

PR #18597, opened 2025-05-31
[JAVA-47035] Created new module libraries-data-db-2 and moved code fr…

panos-kakos:
…om libraries-data-db


panos-kakos:
test this please

PR #18620, opened 2025-06-09
Update README.md

mingMy-00:
### _**Summary**_
Updated the README documentation to reflect the current number of Maven profiles and project lists.

### _**Changes**_
Updated profile count from 9 to 13 to include JDK22 and JDK23 profiles
Updated project list count from 4 to 6 to include default-jdk22 and default-jdk23
This brings the documentation in sync with the actual profile table

### _**Motivation**_
The documentation was outdated after JDK22 and JDK23 support was added. The profile table showed 13 profiles but the text still mentioned 9 profiles, causing confusion for contributors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,13 +36,13 @@ Profile-based segregation
 
 We use Maven build profiles to segregate the huge list of individual projects in our repository.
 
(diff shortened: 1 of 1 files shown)


lor6:
Thanks @mingMy-00 I've merged the PR here.

PR #18668, opened 2025-07-08
BAEL-9305

LeoHelfferich:


persistence-modules/spring-boot-persistence-5/src/main/java/com/baeldung/transactional/OrderService.java
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+@Service
+public class OrderService {
+
+    @Autowired
+    private TestOrderRepository orderRepository;
+
+    @Transactional //<-- marks this method
```
theangrydev, line 13:
```suggestion
    @Transactional
```
I think we should omit this, usually the guideline is to describe in the article rather than in comments

persistence-modules/spring-boot-persistence-5/src/main/java/com/baeldung/transactional/TestOrder.java
```diff
@@ -0,0 +1,21 @@
+package com.baeldung.transactional;
+
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+
+import lombok.Getter;
```
theangrydev, line 9: We have a guideline to not use Lombok in articles that aren't specifically demonstrating Lombok features

persistence-modules/spring-boot-persistence-5/src/test/java/com/baeldung/transactional/TestOrderServiceTest.java
theangrydev, line 1: Great! Very clean

persistence-modules/spring-boot-persistence-5/src/test/java/com/baeldung/transactional/TestOrderServiceTest.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @AfterEach
+    void afterEach() {
+        testOrderRepository.deleteAll();
+    }
+
+    @Test
+    void givenPublicTransactionalMethod_whenCallingIt_thenShouldRollbackOnException() {
+        assertThat(testOrderRepository.findAll()).isEmpty();
```
theangrydev, line 65:
```suggestion
        assertThat(testOrderRepository.findAll())
            .isEmpty();
```
Let's format the chained method calls on separate lines to make them more visible

theangrydev (changes requested):

persistence-modules/spring-boot-persistence-5/src/test/java/com/baeldung/transactional/TestOrderServiceTest.java
LeoHelfferich: Is this really the right way? I apply the Baeldung code style which, after I added line break like you suggested, corrected it back to the way it is now.

liam-rwaj: @LeoHelfferich from the author guidelines:

> formatting - fluent APIs
> - **sometimes, the default formatter does need a bit of manual help**
> - for example, fluent APIs are more readable if we split each method call on its own line
> - if the method call is small enough, and very logically connected to the next one - you can also group 2 on a single line

PR #18675, opened 2025-07-11
BAEL-7662: Introduction to Smithy

sazzer:


gradle-modules/gradle-customization/gradle-smithy/gradle.properties
```diff
@@ -0,0 +1,2 @@
+smithyGradleVersion=1.3.0
+smithyJavaVersion=0.0.1
```
theangrydev, line 2: Any issue with using 0.0.2? It seems available:
https://mvnrepository.com/artifact/software.amazon.smithy.java

gradle-modules/gradle-customization/gradle-smithy/gradle/wrapper/gradle-wrapper.jar
theangrydev, line 1: I think the JAR itself is usually left in the .gitignore

gradle-modules/gradle-customization/gradle-smithy/src/main/java/com/baeldung/smithy/books/server/Main.java
theangrydev, line 1: Is this generated? If not, we should tidy it up to match the Baeldung guidelines

gradle-modules/gradle-customization/gradle-smithy/gradle.properties
sazzer: It's not. Only 0.0.1 is: https://mvnrepository.com/artifact/software.amazon.smithy.java.codegen/plugins

sazzer: Ok - I see.

0.0.2 exists for everything in `software.amazon.smithy.java` but not for anything in `software.amazon.smithy.java.codegen`. I'll fix.

sazzer: But, of course, the two don't seem to be compatible! :( So I'm leaving it on 0.0.1 for now since that works.

gradle-modules/gradle-customization/gradle-smithy/gradle/wrapper/gradle-wrapper.jar
sazzer: I copied the .gitignore from the adjacent gradle-avro module.
And removing the JAR breaks the build:
```
-> % ./gradlew clean
Error: Could not find or load main class org.gradle.wrapper.GradleWrapperMain
Caused by: java.lang.ClassNotFoundException: org.gradle.wrapper.GradleWrapperMain
```

gradle-modules/gradle-customization/gradle-smithy/src/main/java/com/baeldung/smithy/books/server/Main.java
sazzer: It's not generated but it is basically copied as-is from the Smithy examples. However, I've now run all of the code through the formatter so hopefully that should help.

I have left it using System,out instead of SLF4J because that seems to be the norm in the Gradle modules, and because this doesn't require a bunch more dependencies just for two log lines...

gradle-modules/gradle-customization/gradle-smithy/gradle.properties
theangrydev: Huh strange, ok good to know!

gradle-modules/gradle-customization/gradle-smithy/gradle/wrapper/gradle-wrapper.jar
theangrydev: Ok, looks like they do say to keep it, I'm surprised honestly
https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper

"Adding the JAR file to version control is expected. Some organizations do not allow projects to submit binary files to version control, and there is no workaround available."

gradle-modules/gradle-customization/gradle-smithy/src/main/java/com/baeldung/smithy/books/server/Main.java
theangrydev: Ok, let's confirm at final review if an exception to the guidelines makes sense here

PR #18677, opened 2025-07-12
BAEL-7699 Avoid busy-waiting examples

apelan:
Avoid busy-waiting examples

### DIFF
--- a/core-java-modules/core-java-concurrency-basic-4/src/test/java/com/baeldung/concurrent/busywaiting/BusyWaitingUnitTest.java
+++ b/core-java-modules/core-java-concurrency-basic-4/src/test/java/com/baeldung/concurrent/busywaiting/BusyWaitingUnitTest.java
@@ -1,0 +1,78 @@
+package com.baeldung.concurrent.busywaiting;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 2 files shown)


core-java-modules/core-java-concurrency-basic-4/src/test/java/com/baeldung/concurrent/busywaiting/BusyWaitingUnitTest.java
```diff
@@ -0,0 +1,78 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BusyWaitingUnitTest {
```
theangrydev, line 13: Since we don't want to delay the entire CI suite with these intentional sleep demonstrations, let's rename to ManualTest suffix so it doesn't run as part of the regular CI build automatically

apelan: Makes sense, changed

PR #18722, opened 2025-08-01
BAEL-8530: Introduction to Netty-SocketIO

sazzer:


libraries-server-2/src/test/java/com/baeldung/socketio/ServerLiveTest.java
```diff
@@ -0,0 +1,41 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.corundumstudio.socketio.Configuration;
+import com.corundumstudio.socketio.SocketIOServer;
+
+public class ServerLiveTest {
+    private static final Logger LOG = LoggerFactory.getLogger(ServerLiveTest.class);
+
+    @Test
+    void startServer() throws Exception {
```
sazzer, line 14: I've written this as a live test, but it's not really a test. It's just a convenient way to run the server code.

libraries-server-2/src/test/resources/socketio.html
sazzer, line 1: This is a sample HTML file that uses the Socket.IO JavaScript library to connect to the server.

libraries-server-2/src/test/java/com/baeldung/socketio/ServerLiveTest.java
```diff
@@ -0,0 +1,41 @@
...
+            LOG.info("New connection from client {}", client.getRemoteAddress());
+            client.joinRoom("testroom");
+
+            client.getNamespace().getRoomOperations("testroom")
+                .sendEvent("message", "Hello from " + client.getRemoteAddress());
+        });
+
+        server.addDisconnectListener(client -> {
+            LOG.info("Disconnection from client {}", client.getRemoteAddress());
+        });
```
theangrydev, line 31: Should we demo the client leaving the room here, or that is automatic?

theangrydev (changes requested):

PR #18723, opened 2025-08-01
BAEL-9323

Neetika23:


spring-batch-2/pom.xml
```diff
@@ -104,6 +104,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>com.graphql-java</groupId>
+            <artifactId>java-dataloader</artifactId>
+            <version>3.2.0</version>
```
Bwvolleyball, line 110: This can be done asynchronously, but let's pull this out to be a property, like the other dependencies.


```suggestion
            <version>${java-dataloader.version}</version>
```

PR #18825, opened 2025-09-23
BAEL-9183: Configuration Management using Apollo

sazzer:


apollo/src/main/resources/application.properties
```diff
@@ -0,0 +1,7 @@
+spring.application.name=demo
+
+server.port=9090
+
+app.id: baeldung-test
+apollo.cluster: emea
+apollo.meta: http://localhost:8080
```
theangrydev, line 7: Curious I didn't realise properties files worked with colon as well as = 

Can we use = everywhere for consistency? Or use Yaml instead if you prefer

theangrydev (changes requested):

PR #18864, opened 2025-10-15
BAEL-9473 Support @CacheConfig("myCacheName") Declarations in Spring

dvohra09:


KevinGilmore (changes requested):
Remove trailing spaces in pom.xml

PR #18906, opened 2025-11-02
BAEL-9154

umara-123:
[BAEL-9154]
Mapping Abstract Classes in Mapstruct


umara-123:
[BAEL-9154]
Mapping Abstract Classes in Mapstruct

PR #18943, opened 2025-11-14
Bael 9487

Deepak-Vohra:


spring-scheduling/pom.xml
```diff
@@ -1,53 +1,88 @@
...
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-    <artifactId>spring-scheduling</artifactId>
+    <artifactId>spring-retry-module</artifactId>
     <version>0.1-SNAPSHOT</version>
     <packaging>jar</packaging>
-    <name>spring-scheduling</name>
+    <name>spring-retry-module</name>
```
KevinGilmore: As per guidelines, the artifactId and name must match the directory name, so we shouldn't be renaming them here.

PR #18950, opened 2025-11-17
BAEL-7214 | Adding Apache POI HSSFWorkbook code

programenth:
Apache POI HSSFWorkbook: Workbook to Byte Streams and Back


eugenp:
test again

PR #18977, opened 2025-11-22
Bael 9469 bael 9487 bael 9552

Deepak-Vohra:


spring-scheduling/src/main/java/com/baeldung/async/AsyncComponent.java
```diff
@@ -1,28 +1,35 @@
...
+        try {
+                Thread.sleep(5000);
+                return CompletableFuture.completedFuture("hello world !!!!");
+        } catch (InterruptedException e) {
+              return CompletableFuture.failedFuture(e);
+          }
+    }
 
-        return null;
+    @Async("threadPoolTaskExecutor")
```
KevinGilmore: There are two methods here with the same signature.

PR #19039, opened 2025-12-15
Use IDENTITY strategy for JPA ID generation in Foo entity

kriti20041:
### What was changed
- Updated JPA ID generation strategy in Foo entity from AUTO to IDENTITY

### Why
- Makes the example explicit and provider-independent
- Aligns with common JPA usage patterns

### How tested
- Example-only change; no functional behavior impacted


lor6:
Thanks @kriti20041 I've merged the PR here.

PR #19096, opened 2026-01-15
Upgrade Lucene module to version 10.3.2

ashu17188:
Migrate from Lucene 7.4.0 to 10.3.2 with all necessary API changes including StoredFields API, ByteBuffersDirectory, and updated analyzers.


lor6:
Thanks @ashu17188

PR #19145, opened 2026-02-11
BAEL-6448: Time-Sorted Unique Identifiers with Hypersistence TSID

sazzer:


libraries-7/src/test/java/com/baeldung/tsid/TsidUnitTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+import io.hypersistence.tsid.TSID;
+
+public class TsidUnitTest {
+    private static final Instant NOW = Instant.parse("2026-02-11T12:34:56Z");
+
+    private static final Clock CLOCK = Clock.fixed(NOW, ZoneId.of("UTC"));
+
+    @Test
+    void whenGeneratingTsids_thenUniqueValuesAreReturned() {
```
theangrydev: This test relies on TSID implementing equals and hashcode - can you check if they intentionally support this and if so mention it briefly in the article

PR #19202, opened 2026-04-13
test(spring-spel): remove redundant new() on @Autowired fields

awais-haider:
integrationTest

@Autowired field injection replaces field values after construction; the initial new() calls were unused and could mislead readers about the source of the beans.


lor6:
Thanks @awais-haider I've merged the PR.

PR #19226, opened 2026-05-20
Fix ineffective assertion in OuterUnitTest

aki-raa:
Fixes #16237

The assertion `is(nestMembers.size()).equals(2)` was creating a Hamcrest 
matcher but never actually asserting anything. Replaced with 
`assertEquals(2, nestMembers.size())` and removed the unused import.

PR #19228, opened 2026-05-21
fix: replace ineffective assertion in OuterUnitTest (#16237)

### LINKED ISSUE #16237: Wrong testcase,
https://github.com/eugenp/tutorials/blame/e1e94a088ab18095b3745acb96d492db82c8387f/core-java-modules/core-java-11/src/test/java/com/baeldung/OuterUnitTest.java#L32

image

Both test cases are expecting different values for the same code

aki-raa:
## Problem
The assertion `is(nestMembers.size()).equals(2)` was ineffective — 
it creates a Hamcrest matcher but never passes it to assertThat(), 
so the test always passed regardless of the actual value.

## Fix
Replaced with `assertEquals(2, nestMembers.size())` which actually 
validates the size.

Fixes #16237

### DIFF
--- a/core-java-modules/core-java-11/src/test/java/com/baeldung/OuterUnitTest.java
+++ b/core-java-modules/core-java-11/src/test/java/com/baeldung/OuterUnitTest.java
@@ -1,6 +1,5 @@
 package com.baeldung;
 
-import static org.hamcrest.CoreMatchers.is;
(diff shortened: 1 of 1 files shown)


lor6:
Merged https://github.com/eugenp/tutorials/pull/19226

PR #19246, opened 2026-06-15
[hibernate7] hibernate7 like regexp

sk1418:


persistence-modules/hibernate7/src/main/java/com/baeldung/likewithregex/MessageRepository.java
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.query.Param;
+
+public interface MessageRepository extends JpaRepository<Message, Long> {
+
+    @Query("select m from Message m where m.content like regexp :pattern")
+    List<Message> findByContentMatchingRegex(@Param("pattern") String pattern);
+
+    @Query("select m from Message m where m.content not like regexp :pattern")
+    List<Message> findByContentNotMatchingRegex(@Param("pattern") String pattern);
```
theangrydev, line 15: Was it intentional to not uppercase the keywords SELECT, FROM, ...?

I can't remember if we have any guideline on this for Java, I think we used to for SQL articles at least

No big deal either way on my side just checking if you had considered it

persistence-modules/hibernate7/src/test/java/com/baeldung/likewithregex/LikeWithRegexUnitTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.test.context.ActiveProfiles;
+
+@DataJpaTest
+@ActiveProfiles("test")
+class LikeWithRegexUnitTest {
+
+    @Autowired
+    private MessageRepository repository;
+
+
```
theangrydev:
```suggestion

```

persistence-modules/hibernate7/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+            <dependency>
+                <groupId>org.jboss.logging</groupId>
+                <artifactId>jboss-logging</artifactId>
+                <version>3.6.1.Final</version>
+            </dependency>
+            <dependency>
+                <groupId>jakarta.persistence</groupId>
+                <artifactId>jakarta.persistence-api</artifactId>
+                <version>3.2.0</version>
+            </dependency>
```
theangrydev, line 35: Can we omit these dependencies if they are pulled in transitively anyway? If not, we should have properties for the version numbers

persistence-modules/hibernate7/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+
+    <properties>
+        <hibernate.version>7.2.0.Final</hibernate.version>
```
theangrydev: Latest is now [7.4.2.Final](https://mvnrepository.com/artifact/org.hibernate.orm/hibernate-core/7.4.2.Final)

Update or intentional to use 7.2.0 specifically?

theangrydev (changes requested):

persistence-modules/hibernate7/pom.xml
sk1418: extracted to properties

persistence-modules/hibernate7/pom.xml
sk1418: upgraded to 7.4.2.Final

persistence-modules/hibernate7/src/main/java/com/baeldung/likewithregex/MessageRepository.java
sk1418: In the hiberate doc and examples, they use lowercase keywords. They do mention keywords are case-insensitive, though.
So, if it's okay with you, I would have them in lowercase to align with the Hibernate doc. E.g., "... like regexp operator..." instead of "LIKE REGEXP operator"

PR #19261, opened 2026-07-02
Fix feign-okhttp version to 13.2.1 to support PATCH requests

### LINKED ISSUE #17792: [ISSUE] spring-cloud-modules/spring-cloud-openfeign-2 | feign-okhttp version
I was reading this post (https://www.baeldung.com/openfeign-http-patch-request) because I had exactly the same problem, but it is important to mention the correct version of feign-okhttp, because when I used 13.3 ( that intellij generates for me) it didn't work, and kept returning the Invalid HTTP method: PATCH executing PATCH error, but when changing to version 13.2.1 it ran

simpanda01:
Fixes #17792

## Problem
The `feign-okhttp` dependency in `spring-cloud-openfeign-2` had no explicit
version, so it resolved transitively to `13.3`. This version throws
`Invalid HTTP method: PATCH` when executing PATCH requests through Feign +
OkHttp.

## Fix
Pinned `feign-okhttp` to version `13.2.1` explicitly, which does not have
this issue.

## Verification
- Ran `mvn compile` on the module — build succeeds.
- Confirmed the reported PATCH failure no longer occurs with 13.2.1.

### DIFF
--- a/spring-cloud-modules/spring-cloud-openfeign-2/pom.xml
+++ b/spring-cloud-modules/spring-cloud-openfeign-2/pom.xml
@@ -50,6 +50,7 @@
         <dependency>
             <groupId>io.github.openfeign</groupId>
             <artifactId>feign-okhttp</artifactId>
+            <version>13.2.1</version>
         </dependency>
         <dependency>
             <groupId>io.github.openfeign.form</groupId>


lor6:
Thanks @simpanda01 I would prefer to keep this dependency managed and fix the issue when the Spring Cloud dependency is upgraded.

PR #19263, opened 2026-07-04
Fix weak email validation in Vaadin custom validator example

### LINKED ISSUE #16682: In this File vaadin/src/main/java/com/baeldung/introduction/basics/VaadinFlowBasics.java
When we change contact to   **var contact = new Contact("John Doe", "johndoe.com", "123-456-7890");** having wrong email, then also if condition **binder.validate().isOk()** is true.

simpanda01:
Fixes #16682

The custom email validator in `FormView.java`'s `addCustomValidationExample()` only checked `email.contains("@")`, which incorrectly passes invalid strings like `"@"` or `"abc@"` as valid emails.

This PR replaces it with a proper regex-based check, consistent with the phone number validator already used in the same method.


lor6:
Thanks @simpanda01

Issue #18683, opened 2025-07-15
Unnecessary finisher


viliam-durina:
In this code: https://github.com/eugenp/tutorials/blob/5a74185fb083537a2ad7021ea214e37abadf8791/core-java-modules/core-java-streams-7/src/main/java/com/baeldung/streams/gatherer/SlidingWindowGatherer.java#L34

the size of `state` is never 3 - the `integrator` never leaves such state. The `finisher` can be removed.

Also, using `ArrayList` for this task is insane: `removeFirst` needs to copy the entire array. `ArrayDeque` would be much more efficient.

lor6:
Fixed by https://github.com/eugenp/tutorials/pull/19284

PR #19284, opened 2026-07-26
Fix broken sliding window logic in SlidingWindowGatherer

soyeon1806:
`SlidingWindowGatherer` currently emits one single-element list per input element
instead of overlapping windows of three, which makes its own unit test fail.

### Cause

The `state.size() == 3` guard was dropped from the integrator in aae2714723.
Without the guard, each element is pushed downstream
and then immediately removed from the state,
so the window never grows past one.

For the stream `1..5`:

| | Result |
|---|---|
| `SlidingWindowGathererUnitTest` expects | `[[1, 2, 3], [2, 3, 4], [3, 4, 5]]` |
| Current code produces | `[[1], [2], [3], [4], [5]]` |

The test asserts a window count of 3, so it fails with 5.

### Changes

- Restore the `state.size() == 3` guard in `integrator()`.
- Remove the `finisher()` override.
  It returned an empty `BiConsumer`, which is exactly what `Gatherer`'s default finisher already does.
  This addresses the remaining point of #18683.

### Why this went unnoticed

`core-java-streams-7` requires JDK 24, so it is commented out in `core-java-modules/pom.xml`
and only listed under the `default-jdk24` and `integration-jdk24` profiles.
The module is not part of an ordinary build.

Refs #18683

### DIFF
--- a/core-java-modules/core-java-streams-7/src/main/java/com/baeldung/streams/gatherer/SlidingWindowGatherer.java
+++ b/core-java-modules/core-java-streams-7/src/main/java/com/baeldung/streams/gatherer/SlidingWindowGatherer.java
@@ -1,7 +1,6 @@
 package com.baeldung.streams.gatherer;
 
 import java.util.*;
-import java.util.function.BiConsumer;
 import java.util.function.Supplier;
 import java.util.stream.Gatherer;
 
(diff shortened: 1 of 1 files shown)


lor6:
Thanks @soyeon1806 I've merged the PR.

PR #19295, opened 2026-08-13
BAEL-8820: DB-Scheduler - Persistent cluster-friendly scheduler for Java

sazzer:


libraries-7/src/test/resources/dbscheduler.sql
```diff
@@ -0,0 +1,15 @@
...
+                                 picked BIT,
+                                 picked_by varchar(50),
+                                 last_success TIMESTAMP WITH TIME ZONE,
+                                 last_failure TIMESTAMP WITH TIME ZONE,
+                                 consecutive_failures INT,
+                                 last_heartbeat TIMESTAMP WITH TIME ZONE,
+                                 version BIGINT,
+                                 priority SMALLINT,
+                                 PRIMARY KEY (task_name, task_instance)
+)
```
theangrydev: Let's fix the formatting here

libraries-7/src/test/java/com/baeldung/dbscheduler/DbSchedulerLiveTest.java
```diff
@@ -0,0 +1,149 @@
...
+import com.github.kagkarlsson.scheduler.task.schedule.CronSchedule;
+import com.github.kagkarlsson.scheduler.task.schedule.FixedDelay;
+import org.hsqldb.jdbc.JDBCDataSource;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DbSchedulerLiveTest {
```
theangrydev: JUnit 5 test classes are package-private per the repo conventions, so let's drop the public

libraries-7/src/test/java/com/baeldung/dbscheduler/DbSchedulerLiveTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+        Scheduler scheduler = Scheduler.create(dataSource)
+            .startTasks(task)
+            .registerShutdownHook()
+            .build();
+
+        scheduler.start();
+
+        // Just block forever
+        Thread.currentThread().join();
```
theangrydev: These join() tests block forever here, so they never complete or assert anything, even when run deliberately. Let's use a CountDownLatch (or Awaitility) in the execute handler, await with a timeout, assert the task actually ran, then scheduler.stop(). Same for the one-time and dynamic recurring tests below.

libraries-7/src/test/java/com/baeldung/dbscheduler/DbSchedulerLiveTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+        assertTrue(scheduler.getSchedulerState().isStarted());
+
+
+        scheduler.stop();
+        assertTrue(scheduler.getSchedulerState().isShuttingDown());
+    }
+
+    @Test
+    void whenStartingARecurringTask_theTaskRunsRegularly() throws Exception {
```
theangrydev:
```suggestion
    void whenStartingARecurringTask_thenTheTaskRunsRegularly() throws Exception {
```

whenXxx_thenXxxx

Same below

theangrydev (changes requested):

libraries-7/src/test/resources/dbscheduler.sql
sazzer: This was directly copied from the db-scheduler files, formatting and all. I can update it if you want though?

theangrydev: Yes please, we should follow the Baeldung formatting style for PRs

